PR #3, opened 2019-12-17
Fix broken pdf redirects

hmstepanek:
Previously if a pdf url returned a redirect the pdf would not be proxied correctly. This PR fixes a couple issue that were causing this:
1. The pdf url that was being passed from the python app was incomplete (it was missing query params).
1. The Content-Type header request was not following redirects so if the first redirect did not respond with a pdf Content-Type the request would be routed to legacy via.
1. The nginx redirect url replacement was incorrectly replacing the url with the previous /_id/ endpoint instead of the new /proxy/static/ endpoint.
1. The CORS header was being appended to rather than overwritten inside nginx when proxying the third-party pdf which lead to an invalid value in the CORS header when the third-party pdf server also returned a CORS header.

### DIFF
--- a/py_proxy/views.py
+++ b/py_proxy/views.py
@@ -55,7 +55,11 @@ def content_type(request):
 
     with requests.get(url, stream=True, allow_redirects=True) as rsp:
         if rsp.headers.get("Content-Type") in ("application/x-pdf", "application/pdf",):
-            return exc.HTTPFound(request.route_url("pdf", pdf_url=url))
+            return exc.HTTPFound(
+                request.route_url(
+                    "pdf", pdf_url=request.matchdict["url"], _query=request.params
+                )
+            )
     via_url = request.registry.settings["legacy_via_url"]
     return exc.HTTPFound(f"{via_url}/{url}")
 
@@ -66,12 +70,17 @@ def _drop_from_url_begining(drop_chars, url):
     return url[drop_before:]
 
 
-def includeme(config):
-    """Pyramid config."""
+def add_routes(config):
+    """Add routes to pyramid config."""
     config.add_route("index", "/")
     config.add_route("status", "/_status")
     config.add_route("favicon", "/favicon.ico")
     config.add_route("robots", "/robots.txt")
     config.add_route("pdf", "/pdf/{pdf_url:.*}")
     config.add_route("content_type", "/{url:.*}")
+
+
+def includeme(config):
+    """Pyramid config."""
+    add_routes(config)
     config.scan(__name__)
--- a/py_proxy/views.py
+++ b/py_proxy/views.py
@@ -38,7 +38,8 @@ def favicon(request):
 def pdf(request):
     """HTML page with client and the PDF embedded."""
     nginx_server = request.registry.settings["nginx_server"]
-    pdf_url = request.matchdict["pdf_url"]
+    pdf_url = _drop_from_url_begining("/pdf/", request.path_qs)
+
     return {
         "pdf_url": f"{nginx_server}/proxy/static/{pdf_url}",
         "client_embed_url": request.registry.settings["client_embed_url"],
@@ -50,12 +51,19 @@ def pdf(request):
 @view.view_config(route_name="content_type")
 def content_type(request):
     """Routes the request according to the Content-Type header."""
-    url = request.matchdict["url"]
-    with requests.get(url, stream=True) as rsp:
+    url = _drop_from_url_begining("/", request.path_qs)
+
+    with requests.get(url, stream=True, allow_redirects=True) as rsp:
         if rsp.headers.get("Content-Type") in ("application/x-pdf", "application/pdf",):
             return exc.HTTPFound(request.route_url("pdf", pdf_url=url))
     via_url = request.registry.settings["legacy_via_url"]
     return exc.HTTPFound(f"{via_url}/{url}")
+
+
+def _drop_from_url_begining(drop_chars, url):
+    """Drop drop_chars from begining of url."""
+    drop_before = len(drop_chars)
+    return url[drop_before:]
 
 
 def includeme(config):
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -130,6 +130,26 @@ class TestPdfRoute:
 
         assert result["pdf_url"] == f"{nginx_server}/proxy/static/{pdf_url}"
 
+    @pytest.mark.parametrize(
+        "query_param", ["via.request_config_from_frame", "via.open_sidebar"]
+    )
+    @pytest.mark.xfail(raises=AssertionError)
+    @httpretty.activate
+    def test_does_not_include_via_query_params_in_pdf_url(
+        self, pyramid_config, query_param
+    ):
+        url = (
+            "http://thirdparty.url/foo.pdf?param1=abc&param2=123"
+            "via.request_config_from_frame=lms.hypothes.is&via.open_sidebar=1"
+        )
+
+        request = Request.blank(f"/pdf/{url}")
+        request.registry = pyramid_config.registry
+
+        result = views.pdf(request)
+
+        assert query_param not in result["pdf_url"]
+
     def test_pdf_passes_client_embed_url_to_renderer(self, pyramid_config):
         request = Request.blank("/pdf/https://thirdparty.url/foo.pdf")
         request.registry = pyramid_config.registry
@@ -198,7 +218,6 @@ class TestContentTypeRoute:
             ),
         ],
     )
-    @httpretty.activate
     def test_redirect_location(self, pyramid_config, requested_path, expected_location):
 
         request = self._create_request(
@@ -220,7 +239,6 @@ class TestContentTypeRoute:
             ("text/html", "http://via.hypothes.is/https://thirdparty.url"),
         ],
     )
-    @httpretty.activate
     def test_redirects_based_on_content_type_header(
         self, pyramid_config, content_type, redirect_url
     ):
@@ -233,11 +251,31 @@ class TestContentTypeRoute:
 
         result = views.content_type(request)
 
-        redirect_location = redirect_url
-        assert result.location == redirect_location
+        assert result.location == redirect_url
 
+    @pytest.mark.parametrize(
+        "query_param", ["via.request_config_from_frame", "via.open_sidebar"]
+    )
+    @pytest.mark.xfail(raises=AssertionError)
+    def test_does_not_pass_via_query_params_to_thirdparty_server(
+        self, pyramid_config, query_param
+    ):
+        request = self._create_request(
+            request_url="/https://thirdparty.url?"
+            "via.request_config_from_frame=lms.hypothes.is&via.open_sidebar=1",
+            thirdparty_url="https://thirdparty.url",
+            content_type="application/pdf",
+            pyramid_config=pyramid_config,
+        )
+
+        views.content_type(request)
+
+        # pylint: disable-msg=E1101
+        assert query_param not in httpretty.last_request().path
+
+    # pylint: disable-msg=too-many-arguments
     def _create_request(
-        self, request_url, thirdparty_url, content_type, pyramid_config
+        self, request_url, thirdparty_url, content_type, pyramid_config,
     ):
         httpretty.register_uri(
             httpretty.GET,
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -111,6 +111,16 @@ class TestPdfRoute:
 
         assert result["pdf_url"] == f"{nginx_server}/proxy/static/{pdf_url}"
 
+    def test_pdf_passes_thirdparty_url_with_query_params_to_renderer(
+        self, pyramid_request
+    ):
+        pdf_url = "http://thirdparty.url/foo.pdf?param1=abc&param2=123"
+        pyramid_request.path_qs = f"/pdf/{pdf_url}"
+        nginx_server = pyramid_request.registry.settings.get("nginx_server")
+        result = views.pdf(pyramid_request)
+
+        assert result["pdf_url"] == f"{nginx_server}/proxy/static/{pdf_url}"
+
     def test_pdf_passes_client_embed_url_to_renderer(self, pyramid_request):
         result = views.pdf(pyramid_request)
 
@@ -151,22 +161,22 @@ class TestPdfRoute:
 
     @pytest.fixture
     def pyramid_request(self, pyramid_request, pdf_url):
-        pyramid_request.matchdict = {"pdf_url": pdf_url}
+        pyramid_request.path_qs = f"/pdf/{pdf_url}"
         return pyramid_request
 
 
 class TestContentTypeRoute:
     @pytest.mark.parametrize(
-        "content_type,redirect_location",
+        "content_type,redirect_path",
         [
-            ("application/pdf", "/pdf/http://thirdparty.url"),
-            ("application/x-pdf", "/pdf/http://thirdparty.url"),
-            ("text/html", "http://via.hypothes.is/http://thirdparty.url"),
+            ("application/pdf", "/pdf/"),
+            ("application/x-pdf", "/pdf/"),
+            ("text/html", "http://via.hypothes.is/"),
         ],
     )
     @httpretty.activate
     def test_redirects_based_on_content_type_header(
-        self, pyramid_request, content_type, redirect_location
+        self, pyramid_request, url, content_type, redirect_path
     ):
         httpretty.register_uri(
             httpretty.GET,
@@ -177,11 +187,18 @@ class TestContentTypeRoute:
 
         result = views.content_type(pyramid_request)
 
+        redirect_location = redirect_path + url
         assert result.location == redirect_location
 
+    @pytest.fixture(
+        params=["http://thirdparty.url", "http://thirdparty.url?param1=abc&param2=123",]
+    )
+    def url(self, request):
+        return request.param
+
     @pytest.fixture
-    def pyramid_request(self, pyramid_request):
-        pyramid_request.matchdict = {"url": "http://thirdparty.url"}
+    def pyramid_request(self, pyramid_request, url):
+        pyramid_request.path_qs = f"/{url}"
 
         def route_url(path, pdf_url):
             return f"/{path}/{pdf_url}"
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -4,15 +4,15 @@ import httpretty
 import pytest
 from bs4 import BeautifulSoup
 from jinja2 import Environment, FileSystemLoader
-from pyramid.request import Request
 
 from py_proxy import views
 
 
 class TestIndexRoute:
-    def test_index_returns_empty_parameters_to_pass_to_template(self, pyramid_config):
-        request = Request.blank("/status")
-        request.registry = pyramid_config.registry
+    def test_index_returns_empty_parameters_to_pass_to_template(
+        self, make_pyramid_request
+    ):
+        request = make_pyramid_request("/status")
 
         result = views.index(request)
 
@@ -34,9 +34,8 @@ class TestIndexRoute:
 
 
 class TestStatusRoute:
-    def test_status_returns_200_response(self, pyramid_config):
-        request = Request.blank("/status")
-        request.registry = pyramid_config.registry
+    def test_status_returns_200_response(self, make_pyramid_request):
+        request = make_pyramid_request("/status")
 
         result = views.status(request)
 
@@ -61,9 +60,8 @@ class TestIncludeMe:
 
 
 class TestFaviconRoute:
-    def test_returns_favicon_icon(self, pyramid_config):
-        request = Request.blank("/favicon.ico")
-        request.registry = pyramid_config.registry
+    def test_returns_favicon_icon(self, make_pyramid_request):
+        request = make_pyramid_request("/favicon.ico")
 
         result = views.favicon(request)
 
@@ -72,9 +70,8 @@ class TestFaviconRoute:
 
 
 class TestRobotsTextRoute:
-    def test_returns_robots_test_file(self, pyramid_config):
-        request = Request.blank("/robots.txt")
-        request.registry = pyramid_config.registry
+    def test_returns_robots_test_file(self, make_pyramid_request):
+        request = make_pyramid_request("/robots.txt")
 
         result = views.robots(request)
 
@@ -121,9 +118,8 @@ class TestPdfRoute:
             "http://thirdparty.url/foo.pdf?param1=abc&param2=123",
         ],
     )
-    def test_pdf_passes_thirdparty_url_to_renderer(self, pyramid_config, pdf_url):
-        request = Request.blank(f"/pdf/{pdf_url}")
-        request.registry = pyramid_config.registry
+    def test_pdf_passes_thirdparty_url_to_renderer(self, make_pyramid_request, pdf_url):
+        request = make_pyramid_request(f"/pdf/{pdf_url}")
         nginx_server = request.registry.settings.get("nginx_server")
 
         result = views.pdf(request)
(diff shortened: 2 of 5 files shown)


py_proxy/views.py
```diff
@@ -50,8 +52,9 @@ def pdf(request):
 @view.view_config(route_name="content_type")
 def content_type(request):
     """Routes the request according to the Content-Type header."""
-    url = request.matchdict["url"]
-    with requests.get(url, stream=True) as rsp:
+    # Drop / from url.
+    url = request.path_qs[1:]
```
hmstepanek: I'm not particularly proud of this index based approach but I thought it was simpler than piecing the url back together.

Previously we were getting the path from matchdict. However, in order to be complete we would have to generate a url with that and the query params which would most likely involve using urllib3 and it seems kinda silly and complex to me to break apart the url into pieces and then put it back together into the almost original url.

py_proxy/views.py
```diff
@@ -38,7 +38,9 @@ def favicon(request):
 def pdf(request):
     """HTML page with client and the PDF embedded."""
     nginx_server = request.registry.settings["nginx_server"]
-    pdf_url = request.matchdict["pdf_url"]
+    # Drop /pdf/ from url.
+    pdf_url = request.path_qs[5:]
```
seanh: Can you add a unit test for the query string preserving here? There doesn't seem to be a test that ensures that the query string is preserved on `pdf_url`. There should be tests for both when the request does and does not have a query string.

I think adding a unit test is the main thing but the code could be _slightly_ nicer by removing the hard-coded `5` in one way or another. Here are a couple of suggestions that I think would work:

You can get the PDF URL (without query string) as `request.matchdict['pdf_url']` and then get the query string (without the leading `?`) as `request.query_string`, so you can construct the desired `pdf_url` without indexing and without `urllib3` like this:

```suggestion
    pdf_url = request.matchdict["pdf_url"]
    if request.query_string:
        pdf_url = pdf_url + "?" + request.query_string
```

This is longer but it's more explicit and readable, and the code is more decoupled from the exact URL path string.

Or alternatively use `len("/pdf/")` to make it more explicit that you're truncating the `"/pdf/"` from the beginning of the string:

```python
pdf_url = request.path_qs[len("/pdf/"):]
```

py_proxy/views.py
seanh: Can you also add a unit test for this one, that the query string is preserved on `url`? There should be tests for both when the request does and does not have a query string.

I think this is fine, and not as bad as the similar `/pdf/` truncating above. You could use the same `request.query_string`-based approach as I suggested above though, and it could be moved into a helper method:

```diff
diff --git a/py_proxy/views.py b/py_proxy/views.py
index 845527c..3851c2f 100644
--- a/py_proxy/views.py
+++ b/py_proxy/views.py
@@ -52,8 +52,7 @@ def pdf(request):
 @view.view_config(route_name="content_type")
 def content_type(request):
     """Routes the request according to the Content-Type header."""
-    # Drop / from url.
-    url = request.path_qs[1:]
+    url = append_query_string(request.matchdict["url"], request)
     with requests.get(url, stream=True, allow_redirects=True) as rsp:
         if rsp.headers.get("Content-Type") in ("application/x-pdf", "application/pdf",):
             return exc.HTTPFound(request.route_url("pdf", pdf_url=url))
@@ -61,6 +60,16 @@ def content_type(request):
     return exc.HTTPFound(f"{via_url}/{url}")
 
 
+def append_query_string(url, request):
+    """Return ``url`` with the query string from ``request`` appended.
+
+    If ``request`` has no query string then just return ``url`` unmodified.
+    """
+    if request.query_string:
+        return url + "?" + request.query_string
+    return url
+
+
 def includeme(config):
     """Pyramid config."""
     config.add_route("index", "/")
```

seanh (changes requested):
Looks great. I'd like you to add a couple of unit tests for the query string-preserving in the Python code.

Optionally, I also suggested ways of constructing those URLs more explicitly in case you want to get rid of the string indexing.

py_proxy/views.py
hmstepanek: I'm going with the len("/pdf/") approach and I've made it into a private helper function as it's used in a couple places.

py_proxy/views.py
hmstepanek: I'm currently trying to decide how to actually test this since right now we are mocking everything so any test I wrote would in no way guarantee that the mocked argument return actually contained the query params when not mocked. I can write a test that tests both of them but since we are mocking the value of pdf url if someone unknowingly switched back to matchdict and just changed the mock to return the pdf url, the tests would still pass even though matchdict doesn't include the query params.

hmstepanek: It doesn't feel like a very strong test to me. Wondering if there's a better way to test this?

seanh: The general guideline for our Python unit tests is that:

1. In general when one of our modules uses _another one of our modules_ then you _should_ mock the other module. We test each of our modules in isolation from our own other modules, using mocking when necessary to achieve this isolation.

2. But in general when one of our modules uses a third-party or standard library module you should _not_ mock that third-party module. So for example our model tests run against a real SQLAlchemy and Postgres DB, we don't mock SQLAlchemy. Or in your tests here, you're using httpretty to fake network responses rather than mocking `requests`. Similarly, _tests should not mock the Pyramid request_ or parts of the request. They should either use the `DummyRequest` class that Pyramid provides for testing or use a real Pyramid request

Here's an example of how to test this. I think this is a full-fidelity test that really tests the view code using a real `Request` object and the real `route_url()`. Some notes on this example:

* You wouldn't really write it as a separate `TestContentTypeWithAndWithoutQueryString` class like the example does. You'd just add it into `TestContentTypeRoute`

* You wouldn't really import `Request` and `testing` in the test method (you'd move those imports to the top of the file)

* We don't want real network requests to accidentally be sent if a test forgets to activate httpretty or forgets to register a URI with httpretty. We also don't want httpretty boilerplate (activating it, registering URIs) to be in our tests, we want that tidied away into fixtures. For an example of how to do this see [the `httpretty_` fixture in LMS's `conftest.py`](https://github.com/hypothesis/lms/blob/a860bc44807e6135f5b498c36c74a3b1ff2a4445/tests/unit/conftest.py#L158-L189) and then see [an example test fixture that registers a response](https://github.com/hypothesis/lms/blob/f7649b8da9703488d48c9acbf771ab564c3b1b8f/tests/unit/lms/services/exceptions_test.py#L189-L199). For now, in my example below, the httpretty stuff is just done in the test itself

* Normally tests would use [Pyramid's `DummyRequest`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/testing.html#pyramid.testing.DummyRequest), which is a dummy request object (not a mock) but [`DummyRequest` doesn't support `path_qs`](https://github.com/Pylons/pyramid/blob/497e667e41b83193d7c5a4f74e9419320d755c46/src/pyramid/testing.py#L361). Or rather, `path_qs` is always an empty string with `DummyRequest`'s.

  This means that, if the code reads `request.path_qs`, the tests for that code would need to manually set `pyramid_request.path_qs = "<PATH_WITH_QUERY_STRING>"`. 

  This isn't a big problem but it does couple the tests to internal implementation details of the code.

  For this reason I've used a real Pyramid `Request` object (constructed by calling [`Request.blank()`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.blank)) in my example below. This sets `path_qs` and most other request attributes automatically.

* The code calls `request.route_url()`. Rather than mocking `route_url()` I've used [`with testing.testConfig(request=request) as config:`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/testing.html#pyramid.testing.testConfig) and then set up the route with `config.add_route("pdf", "/pdf/{pdf_url:.*}")`. This way the code can call the real `route_url()` and get the expected result. This works with both real and `DummyRequest` objects. In general you shouldn't mock request methods like `route_url()`.

  You probably want something global (a fixture in `conftest.py`) to call the real code that does the `config.add_route()`'s, rather than duplicating the `add_route()` call in the test as I've done in the example.

```python
class TestContentTypeWithAndWithoutQueryString:
    @pytest.mark.parametrize(
        "requested_path,expected_location",
        [
            # If the requested URL has no query string then it should just
            # redirect to the requested URL, with no query string (but with the
            # leading "/" truncated).
            ("/https://example.com/foo", "http://localhost/pdf/https://example.com/foo"),

            # If the requested URL has a query string then the query string
            # should be preserved in the URL that it redirects to.
            ("/https://example.com/foo?bar=baz", "http://localhost/pdf/https://example.com/foo?bar=baz"),
        ],
    )
    @httpretty.activate
    def test_redirect_location(self, requested_path, expected_location):
        from pyramid.request import Request
        from pyramid import testing

        self.register_uri()
        with testing.testConfig() as config:
            config.add_route("pdf", "/pdf/{pdf_url:.*}")
            request = Request.blank(requested_path)

            redirect = views.content_type(request)

        assert redirect.location == expected_location

    @staticmethod
    def register_uri():
        httpretty.register_uri(
            httpretty.GET,
            "https://example.com/foo",
            body="{}",
            adding_headers={"Content-Type": "application/pdf"},
        )

```

Note that there's a tradeoff between making the tests more realistic by using a real `Request` and setting up things like `path_qs` and `route_url()` in a real way, vs keeping the test code simpler and easier to read and write.

`request.matchdict` is a good example of this. Even with the real `Request` object used in my code above, `request.matchdict` is still `{}`. There might be a way for the tests to trigger the necessary Pyramid code to set `matchdict` in the real way. But it would be a lot simpler for the tests to just create either a real `Request` or a `DummyRequest` and then just do `pyramid_request.matchdict = {"url": "..."}`. This does couple the tests a little bit to an implementation detail of the code under test. But it's also a lot more straightforward

seanh (changes requested):
Testing this in Canvas locally the login integration isn't working (client doesn't automatically log in to user account).

I think the handling of the `via.*` query params is buggy:

1. The `content_type()` view doesn't remove the `via.*` query params from the URL before doing the `GET` request to detect the content type. So it's sending a `GET` to `https://example.com/foo.pdf?via.open_sidebar=1&via.request_config_from_frame=http%3A%2F%2Flocalhost%3A8001`. It should remove the `via.*` params from the URL before sending the `GET`

2. When the `pdf()` view receives a request (because the `content_type()` view has redirected to it) the `via.*` query params don't work because they've been escaped. When the `pdf()` view is called `request.url` is `https://example.com/foo.pdf%3Fvia.open_sidebar=1&via.request_config_from_frame=http%253A%252F%252Flocalhost%253A8001` and because of the escaping `request.params` is `{}`.

   I think this is the fault of the `content_type()` view again. It's calling `request.route_url("pdf", pdf_url=url)` where `url` is a `url` with `?via.open_sidebar=1&via.request_config_from_frame=http%3A%2F%2Flocalhost%3A8001`. But `route_url()` will escape the query params at the end of the URL. It needs to:

   1. Remove the query params from `url`
   2. Pass the query params as a dict using `route_url()`'s `_query` argument: `request.route_url("pdf", pdf_url=url_without_query_params, _query={"via.open_sidebar": ..., "via.request_config_from_frame": ...})`

   See [`route_url()`'s docs](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.route_url)

3. I also noticed that in the `pdf()` view `pdf_url` already has the `via.*` params on it but then the two `via.*` params are also passed to the template separately as `h_open_sidebar` and `h_request_config`. It seems like the `via.*` params should only appear once in the template variables.

hmstepanek:
I believe this coverage failure is just due to the xfail tests and it can be ignored.

seanh:
Part of why coverage is failing is that the second `assert` in the xfail test is never being hit because the first assert fails. Coverage is kind of right that this is a problem--the xfail test is not currently testing that the second assert in fact fails. It can be fixed using `parametrize`:

```diff
diff --git a/tests/unit/py_proxy/views_test.py b/tests/unit/py_proxy/views_test.py
index 3039193..67a5388 100644
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -131,8 +131,13 @@ class TestPdfRoute:
         assert result["pdf_url"] == f"{nginx_server}/proxy/static/{pdf_url}"
 
     @pytest.mark.xfail
+    @pytest.mark.parametrize(
+        "query_param", ["via.request_config_from_frame", "via.open_sidebar"]
+    )
     @httpretty.activate
-    def test_does_not_include_via_query_params_in_pdf_url(self, pyramid_config):
+    def test_does_not_include_via_query_params_in_pdf_url(
+        self, pyramid_config, query_param
+    ):
         url = (
             "http://thirdparty.url/foo.pdf?param1=abc&param2=123"
             "via.request_config_from_frame=lms.hypothes.is&via.open_sidebar=1"
@@ -143,8 +148,7 @@ class TestPdfRoute:
 
         result = views.pdf(request)
 
-        assert "via.request_config_from_frame" not in result["pdf_url"]
-        assert "via.open_sidebar" not in result["pdf_url"]
+        assert query_param not in result["pdf_url"]
 
     def test_pdf_passes_client_embed_url_to_renderer(self, pyramid_config):
         request = Request.blank("/pdf/https://thirdparty.url/foo.pdf")
```

Now all the lines in the xfail test are being hit.

seanh:
The second piece of missing coverage is due to a similar reason: again an xfail test has lines that aren't being reached because an `assert` that comes before those test lines fail. I think again coverage is kind of right here: since the assert isn't the last thing in the test, part of the test isn't being run, so we don't actually know that the test would pass if the code was correct and the xfail was removed.

In this case, as well as adding another `parametrize`, we need to remove the callback and move the assert to the end of the test, which also nicely simplifies the test code:

```diff
diff --git a/tests/unit/py_proxy/views_test.py b/tests/unit/py_proxy/views_test.py
index 3039193..a82326f 100644
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -252,37 +252,33 @@ class TestContentTypeRoute:
         assert result.location == redirect_url
 
     @pytest.mark.xfail
+    @pytest.mark.parametrize(
+        "query_param", ["via.request_config_from_frame", "via.open_sidebar"]
+    )
     @httpretty.activate
-    def test_does_not_pass_via_query_params_to_thirdparty_server(self, pyramid_config):
-        def request_callback(request, uri, response_headers):
-            assert "via.request_config_from_frame" not in uri
-            assert "via.open_sidebar" not in uri
-            return [200, response_headers, "{}"]
-
+    def test_does_not_pass_via_query_params_to_thirdparty_server(
+        self, pyramid_config, query_param
+    ):
         request = self._create_request(
             request_url="/https://thirdparty.url?"
             "via.request_config_from_frame=lms.hypothes.is&via.open_sidebar=1",
             thirdparty_url="https://thirdparty.url",
             content_type="application/pdf",
             pyramid_config=pyramid_config,
-            request_callback=request_callback,
         )
 
         views.content_type(request)
 
+        assert query_param not in httpretty.last_request().path
+
     # pylint: disable-msg=too-many-arguments
     def _create_request(
-        self,
-        request_url,
-        thirdparty_url,
-        content_type,
-        pyramid_config,
-        request_callback=None,
+        self, request_url, thirdparty_url, content_type, pyramid_config,
     ):
         httpretty.register_uri(
             httpretty.GET,
             thirdparty_url,
-            body=request_callback or "{}",
+            body="{}",
             adding_headers={"Content-Type": content_type},
         )
         request = Request.blank(request_url)
```

Also note that putting `assert`s in an httpretty request callback isn't a good idea: since the `assert` fails the `request_callback()` method never completes, `return [200, response_headers, "{}"]` never gets run, no response ever gets send back to `requests`. So it's good to avoid doing that

seanh:
You need to add `xfail_strict=true` so that the tests will actually fail if an xfail test passes:

```diff
diff --git a/tox.ini b/tox.ini
index b547c0b..4d9a924 100644
--- a/tox.ini
+++ b/tox.ini
@@ -3,6 +3,7 @@ filterwarnings =
     error
     ; Suppress warnings about an import of `imp` by Pyramid
     ignore:the imp module is deprecated in favour of importlib
+xfail_strict=true
 
 [tox]
 envlist = tests
```

seanh:
You need to add `raises=AssertionError` to the xfails to make sure that they're failing because the assertion fails and not because something else is going wrong. I think one of your xfail tests is actually failing for a different reason other than the assertion.

```diff
diff --git a/tests/unit/py_proxy/views_test.py b/tests/unit/py_proxy/views_test.py
index 3039193..ed01f6d 100644
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -130,7 +130,7 @@ class TestPdfRoute:
 
         assert result["pdf_url"] == f"{nginx_server}/proxy/static/{pdf_url}"
 
-    @pytest.mark.xfail
+    @pytest.mark.xfail(raises=AssertionError)
     @httpretty.activate
     def test_does_not_include_via_query_params_in_pdf_url(self, pyramid_config):
         url = (
@@ -251,7 +251,7 @@ class TestContentTypeRoute:
 
         assert result.location == redirect_url
 
-    @pytest.mark.xfail
+    @pytest.mark.xfail(raises=AssertionError)
     @httpretty.activate
     def test_does_not_pass_via_query_params_to_thirdparty_server(self, pyramid_config):
         def request_callback(request, uri, response_headers):
```

seanh:
Can you remove all the `@httpretty.activate`'s? They don't do anything, the global `httpretty_` fixture already ensures that it's activated for all tests:

```diff
diff --git a/tests/unit/py_proxy/views_test.py b/tests/unit/py_proxy/views_test.py
index 3039193..ebbe374 100644
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -131,7 +131,6 @@ class TestPdfRoute:
         assert result["pdf_url"] == f"{nginx_server}/proxy/static/{pdf_url}"
 
     @pytest.mark.xfail
-    @httpretty.activate
     def test_does_not_include_via_query_params_in_pdf_url(self, pyramid_config):
         url = (
             "http://thirdparty.url/foo.pdf?param1=abc&param2=123"
@@ -214,7 +213,6 @@ class TestContentTypeRoute:
             ),
         ],
     )
-    @httpretty.activate
     def test_redirect_location(self, pyramid_config, requested_path, expected_location):
 
         request = self._create_request(
@@ -236,7 +234,6 @@ class TestContentTypeRoute:
             ("text/html", "http://via.hypothes.is/https://thirdparty.url"),
         ],
     )
-    @httpretty.activate
     def test_redirects_based_on_content_type_header(
         self, pyramid_config, content_type, redirect_url
     ):
@@ -252,7 +249,6 @@ class TestContentTypeRoute:
         assert result.location == redirect_url
 
     @pytest.mark.xfail
-    @httpretty.activate
     def test_does_not_pass_via_query_params_to_thirdparty_server(self, pyramid_config):
         def request_callback(request, uri, response_headers):
             assert "via.request_config_from_frame" not in uri
```

seanh:
I think there's a problem with the `httpretty_` fixture copied from LMS, which suggests that the LMS one has the same problem as well. The `httpretty_` fixture registers responses, and these sometimes seem to override the responses registered by the individual tests. It can be fixed like this:

```diff
diff --git a/tests/unit/py_proxy/conftest.py b/tests/unit/py_proxy/conftest.py
index ae20b8f..307de26 100644
--- a/tests/unit/py_proxy/conftest.py
+++ b/tests/unit/py_proxy/conftest.py
@@ -50,24 +50,7 @@ def httpretty_():
     all with mock responses. This handles requests sent using the standard
     urllib2 library and the third-party httplib2 and requests libraries.
     """
-    httpretty.enable()
-
-    # Tell httpretty which HTTP requests we want it to mock (all of them).
-    for method in (
-        httpretty.GET,
-        httpretty.PUT,
-        httpretty.POST,
-        httpretty.DELETE,
-        httpretty.HEAD,
-        httpretty.PATCH,
-        httpretty.OPTIONS,
-        httpretty.CONNECT,
-    ):
-        httpretty.register_uri(
-            method=method,
-            uri=re.compile(r"http(s?)://.*"),  # Matches all http:// and https:// URLs.
-            body="",
-        )
+    httpretty.enable(allow_net_connect=False)
 
     yield
 
```

seanh (changes requested):
Apart from the few test issues above, the code seems to work correctly in Canvas :+1:

tests/unit/py_proxy/views_test.py
```diff
@@ -4,19 +4,21 @@
...
 
 from py_proxy import views
 
 
 class TestIndexRoute:
-    def test_index_returns_empty_parameters_to_pass_to_template(self, pyramid_request):
-        result = views.index(pyramid_request)
+    def test_index_returns_empty_parameters_to_pass_to_template(self, pyramid_config):
+        request = Request.blank("/status")
+        request.registry = pyramid_config.registry
```
seanh: These two lines are repeated several times. Optional, but maybe change them into a fixture? It can use pytest's [factories as fixtures](https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures) pattern:

```python
class TestIndexRoute:
    def test_index_returns_empty_parameters_to_pass_to_template(self, make_pyramid_request):
        pyramid_request = make_pyramid_request("/status")
        ...

@pytest.fixture
def make_pyramid_request(pyramid_config):
    def _make_pyramid_request(path):
        pyramid_request = Request.blank(path)
        pyramid_request.registry = pyramid_config.registry
        return pyramid_request
    return _make_pyramid_request
```

We should name it `pyramid_request` as that's the name that we use for test request objects throughout our code. The reason is that `request` has a special meaning in pytest and conflicting with it can be very confusing.

tests/unit/py_proxy/views_test.py
```diff
@@ -105,86 +114,176 @@ def test_pdf_renders_parameters_in_pdf_template(
...
+        )
 
-    @pytest.fixture
-    def pyramid_request(self, pyramid_request):
-        pyramid_request.matchdict = {"url": "http://thirdparty.url"}
+        views.content_type(request)
 
-        def route_url(path, pdf_url):
-            return f"/{path}/{pdf_url}"
+        # pylint: disable-msg=E1101
```
seanh:
```suggestion
        # pylint: disable=no-member
```

You just need `disable=` not `disable-msg=` and we should always use the human-readable message names not the numeric codes.

tests/unit/py_proxy/views_test.py
```diff
@@ -105,86 +114,176 @@ def test_pdf_renders_parameters_in_pdf_template(
...
+        assert query_param not in httpretty.last_request().path
 
-        pyramid_request.route_url = route_url
-        return pyramid_request
+    # pylint: disable-msg=too-many-arguments
```
seanh:
```suggestion
```

The pylint comment doesn't seem to be necessary (`make lint` still passes)

tests/unit/py_proxy/views_test.py
```diff
@@ -105,86 +114,176 @@ def test_pdf_renders_parameters_in_pdf_template(
...
+        httpretty.register_uri(
+            httpretty.GET,
+            thirdparty_url,
+            body="{}",
+            adding_headers={"Content-Type": content_type},
+        )
+        request = Request.blank(request_url)
+        request.registry = pyramid_config.registry
+        request.matchdict = {"url": thirdparty_url}
+        return request
```
seanh: Optional, but you could make this helper function easier to call and remove some test code duplication by using fixtures.

This has a few advantages:

* The test methods no longer need to pass `pyramid_config` to `make_pyramid_request()` (one less argument to pass each time you call it)
* The test methods no longer need to take `pyramid_config` as a test method argument
* Code duplication around creating request objects (setting `pyramid_request.registry = pyramid_config.registry`) is removed

The way you do it is:

1. Replace the `TestContentTypeRoute._create_request()` helper method with a `TestContentTypeRoute.make_pyramid_request` fixture (with `@pytest.fixture`)
 This allows `make_pyramid_request()` to have a shorter call signature than `self._create_request()` (no `self` and no `pyramid_config` argument)

2. Make the `TestContentTypeRoute.make_pyramid_request` depend on and override the module-level `make_pyramid_request` fixture that I suggested in my other comment above. This removes some test code duplication.

It would look something like this:

```python
class TestContentTypeRoute:
    def test_redirects_based_on_content_type_header(self, make_pyramid_request, content_type, redirect_url):
        pyramid_request = make_pyramid_request(
            request_url="/https://thirdparty.url",
            thirdparty_url="https://thirdparty.url",
            content_type=content_type,
        )
        ...

    ...

    @pytest.fixture
    def make_pyramid_request(self, make_pyramid_request):
        def _make_pyramid_request(request_url, thirdparty_url, content_type):
            httpretty.register_uri(
                httpretty.GET,
                thirdparty_url,
                body="{}",
                adding_headers={"Content-Type": content_type},
            )
            pyramid_request = make_pyramid_request(request_url)
            pyramid_request.matchdict = {"url": thirdparty_url}
            return pyramid_request

        return _make_pyramid_request

...

@pytest.fixture
def make_pyramid_request(pyramid_config):
    def _make_pyramid_request(path):
        pyramid_request = Request.blank(path)
        pyramid_request.registry = pyramid_config.registry
        return pyramid_request
    return _make_pyramid_request
```

seanh (changes requested):
I made a couple of optional test fixture refactoring suggestions, an suggested a couple of pylint comment fixes.

In testing in Canvas, PDFs seem to work but the login integration is broken for HTML assignments. I think the `content_type()` view is forgetting to add the `via.*` query params to the Via URL when it redirects to Via to proxy an HTML file. Can you add a failing test for that and then fix the code so that the test passes?

tests/unit/py_proxy/views_test.py
hmstepanek: Oh yeah that's cause we removed that callback arg.

tests/unit/py_proxy/views_test.py
hmstepanek: Ah ok that makes sense. I was trying to remember if there was a way to do this in a fixture that accepted args but now I'm remembering...return a func.

hmstepanek:
> I think the content_type() view is forgetting to add the via.* query params to the Via URL when it redirects to Via to proxy an HTML file. 

I don't see an issue with this so I'm not sure why it's not auto-logging in but I added a test for it anyway.

seanh:
Re-tested and it seems to work now. I think there might have been a problem with my local client

seanh (approved):

seanh:
Let's get this deployed, then I think we may be ready to switch LMS over to it

PR #34, opened 2020-01-15
Strip client query params from thirdparty url

hmstepanek:
Previously the via.* query params used to configure the hypothesis client were not getting removed from requests to the thirdparty server or in the thirdparty url embedded in the page. This is a fix for https://github.com/hypothesis/py-proxy/issues/6.

### DIFF
--- a/py_proxy/views.py
+++ b/py_proxy/views.py
@@ -1,8 +1,14 @@
 """The views for the Pyramid app."""
+import urllib
+
 import pyramid.httpexceptions as exc
 import requests
 from pyramid import response, view
 from pyramid.settings import asbool
+
+# Client configuration query parameters.
+OPEN_SIDEBAR = "via.open_sidebar"
+CONFIG_FROM_FRAME = "via.request_config_from_frame"
 
 
 @view.view_config(renderer="py_proxy:templates/index.html.jinja2", route_name="index")
@@ -39,20 +45,24 @@ def favicon(request):
 def pdf(request):
     """HTML page with client and the PDF embedded."""
     nginx_server = request.registry.settings["nginx_server"]
-    pdf_url = _drop_from_url_begining("/pdf/", request.path_qs)
+    pdf_url = _generate_url_without_client_query_params(
+        request.matchdict["pdf_url"], request.params
+    )
 
     return {
         "pdf_url": f"{nginx_server}/proxy/static/{pdf_url}",
         "client_embed_url": request.registry.settings["client_embed_url"],
-        "h_open_sidebar": asbool(request.params.get("via.open_sidebar", False)),
-        "h_request_config": request.params.get("via.request_config_from_frame", None),
+        "h_open_sidebar": asbool(request.params.get(OPEN_SIDEBAR, False)),
+        "h_request_config": request.params.get(CONFIG_FROM_FRAME, None),
     }
 
 
 @view.view_config(route_name="content_type")
 def content_type(request):
     """Routes the request according to the Content-Type header."""
-    url = _drop_from_url_begining("/", request.path_qs)
+    url = _generate_url_without_client_query_params(
+        request.matchdict["url"], request.params
+    )
 
     with requests.get(url, stream=True, allow_redirects=True) as rsp:
         if rsp.headers.get("Content-Type") in ("application/x-pdf", "application/pdf",):
@@ -62,6 +72,7 @@ def content_type(request):
                 )
             )
     via_url = request.registry.settings["legacy_via_url"]
+    url = _drop_from_url_begining("/", request.path_qs)
     return exc.HTTPFound(f"{via_url}/{url}")
 
 
@@ -69,6 +80,36 @@ def _drop_from_url_begining(drop_chars, url):
     """Drop drop_chars from begining of url."""
     drop_before = len(drop_chars)
     return url[drop_before:]
+
+
+def _generate_url_without_client_query_params(base_url, query_params):
+    """Return ``base_url`` with non-Via params from ``query_params`` appended.
+
+    Return ``base_url`` with all the non-Via query params from ``query_params``
+    appended to it as a query string. Any params in ``query_params`` that are
+    meant for Via (the ``"via.*`` query params) will be ignored and *not*
+    appended to the returned URL.
+
+    :param base_url: the protocol, domain and path, for example: https://thirdparty.url/foo.pdf
+    :type base_url: str
+
+    :param query_params: the query params to be added to base_url
+    :type query_params: dict
+
+    :return: ``base_url`` with the non-Via query params appended
+    :rtype: str
+    """
+    client_params = [OPEN_SIDEBAR, CONFIG_FROM_FRAME]
+    filtered_params = urllib.parse.urlencode(
+        {
+            param: value
+            for param, value in query_params.items()
+            if param not in client_params
+        }
+    )
+    if filtered_params:
+        return f"{base_url}?{filtered_params}"
+    return base_url
 
 
 def add_routes(config):
--- a/tests/unit/py_proxy/views_test.py
+++ b/tests/unit/py_proxy/views_test.py
@@ -119,7 +119,9 @@ class TestPdfRoute:
         ],
     )
     def test_pdf_passes_thirdparty_url_to_renderer(self, make_pyramid_request, pdf_url):
-        request = make_pyramid_request(f"/pdf/{pdf_url}")
+        request = make_pyramid_request(
+            f"/pdf/{pdf_url}", "http://thirdparty.url/foo.pdf"
+        )
         nginx_server = request.registry.settings.get("nginx_server")
 
         result = views.pdf(request)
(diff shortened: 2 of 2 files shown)


py_proxy/views.py
```diff
@@ -38,20 +42,24 @@ def favicon(request):
...
+    pdf_url = _generate_url_without_client_query_params(
+        request.matchdict["pdf_url"], request.params
+    )
 
     return {
         "pdf_url": f"{nginx_server}/proxy/static/{pdf_url}",
         "client_embed_url": request.registry.settings["client_embed_url"],
-        "h_open_sidebar": int(request.params.get("via.open_sidebar", "0")),
-        "h_request_config": request.params.get("via.request_config_from_frame", None),
+        "h_open_sidebar": int(request.params.get(OPEN_SIDEBAR, "0")),
```
seanh: Calling `int()` means it'll crash if a non-int `via.open_sidebar` value is passed. For example this will crash: <http://localhost:9082/pdf/http://pdf995.com/samples/pdf.pdf?via.open_sidebar=foo>

You could use Pyramid's `asbool()` instead: https://docs.pylonsproject.org/projects/pyramid/en/latest/api/settings.html#pyramid.settings.asbool

seanh: Can you add a test for this? If given an invalid `via.open_sidebar` query param it should not crash, and should fall back to the default behaviour. If think `open_sidebar` should probably default to 1 rather than 0, as well.

py_proxy/views.py
```diff
@@ -64,10 +72,29 @@ def content_type(request):
...
+
+    :param base_url: the protocol, domain and path aka: https://thirdparty.url/foo.pdf
+    :type base_url: string
+
+    :param query_params: the query params to be added to the base_url
+    :type query_params: dict
+
+    :return: the url without the client query params
+    :rtype: string
+    """
```
seanh, line 101: The distinction between "query params" and "client query params" is a little subtle in this docstring, and on first read I was a little confused about whether the method was appending the query params to the URL or not. I think just making it a bit more explicit might help to clarify it. Maybe something like this?

```python
    """Return ``base_url`` with non-Via params from ``query_params`` appended.

    Return ``base_url`` with all the non-Via query params from ``query_params``
    appended to it as a query string. Any params in ``query_params`` that are
    meant for Via (the ``"via.*`` query params) will be ignored and *not*
    appended to the returned URL.

    :param base_url: the protocol, domain and path, for example: https://thirdparty.url/foo.pdf
    :type base_url: str

    :param query_params: the query params to be added to base_url
    :type query_params: dict

    :return: ``base_url`` with the non-Via query params appended
    :rtype: str
    """
```

py_proxy/views.py
```diff
@@ -64,10 +72,29 @@ def content_type(request):
...
+    client_params = [OPEN_SIDEBAR, CONFIG_FROM_FRAME]
+    filtered_params = "&".join(
+        [
+            f"{param}={value}"
+            for param, value in query_params.items()
+            if param not in client_params
+        ]
+    )
```
seanh, line 109: Can we use urllib here? It might be a little safer when it comes to encoding and things.

```diff
diff --git a/py_proxy/views.py b/py_proxy/views.py
index a9e2f53..5ec5384 100644
--- a/py_proxy/views.py
+++ b/py_proxy/views.py
@@ -1,4 +1,6 @@
 """The views for the Pyramid app."""
+import urllib
+
 import pyramid.httpexceptions as exc
 import requests
 from pyramid import response, view
@@ -85,12 +87,9 @@ def _generate_url_without_client_query_params(base_url, query_params):
     :rtype: string
     """
     client_params = [OPEN_SIDEBAR, CONFIG_FROM_FRAME]
-    filtered_params = "&".join(
-        [
-            f"{param}={value}"
-            for param, value in query_params.items()
-            if param not in client_params
-        ]
+
+    filtered_params = urllib.parse.urlencode(
+        {key: value for key, value in query_params.items() if key not in client_params}
     )
     if filtered_params:
         return f"{base_url}?{filtered_params}"
```

I think this is what Via does too: <https://github.com/hypothesis/via/blob/986c2f772b0c5b1d8d7a4fdafb5a87362a834707/via/config_extractor.py#L48-L74>. Via actually has to parse the query params first (using `urlparse.parse_qsl(orig_qs, keep_blank_values=True)`) but py-proxy doesn't need to do that because it can just use Pyramid's `request.params`. But then, after popping the `via.*` query params, it looks like Via re-constructs the query string without them by [calling `urllib.urlencode()`](https://github.com/hypothesis/via/blob/986c2f772b0c5b1d8d7a4fdafb5a87362a834707/via/config_extractor.py#L67-L68).

Note that the organization of Python's standard `urllib` and `urlparse` libraries are different for Via (Python 2) and py-proxy (Python 3).

tests/unit/py_proxy/views_test.py
```diff
@@ -185,12 +186,21 @@ def test_pdf_passes_open_sidebar_query_parameter_to_renderer(
...
         result = views.pdf(request)
 
         assert result["h_request_config"] == expected_h_request_config
 
+    @pytest.fixture
+    def make_pyramid_request(self, make_pyramid_request):
+        def _make_pyramid_request(request_url, thirdparty_url):
+            request = make_pyramid_request(request_url)
+            request.matchdict = {"pdf_url": thirdparty_url}
+            return request
```
seanh, line 201: :+1:

seanh (changes requested):
I'm not sure how but this seems to break opening the sidebar on page load for HTML assignments in Canvas.

All three types of PDF assignment do seem to work correctly though.

Has this also been tested in SpeedGrader?

py_proxy/views.py
hmstepanek: Since this isn't directly related to removing the query params let's do this in a separate PR.

hmstepanek:
I haven't tested this particular change in speedgrader or any lms for that matter-it is working with speedgrader on master though and I don't really anticipate anything about this change that would break that. My current dev env for lms is broken-I've been trying to fix it today but so far it's a no go.

py_proxy/views.py
hmstepanek: > If think open_sidebar should probably default to 1 rather than 0, as well.

The legacy via behaves as follows:
```
if "via.open_sidebar" in via_params:
    template_params["h_open_sidebar"] = True
```
Note if it is not passed, it is not passed to the client config and the sidebar does not open by default. I think if we are going to change this we should discuss it with the team first as this is a change to the current behavior.

seanh:
Hi Hannah,

Ok, I didn't realise that you didn't have a local development environment set up and working. That's a big red flag -- py-proxy is being developed for the LMS app and is a core part of the LMS experience. I think we need to get your dev env running and you testing things in Canvas and the other LMS's before sending PRs. Let's get that working before we go any further with this or other PRs.

Let me know if you get stuck with getting it working. It should be very easy to set up a local LMS dev env. See LMS's [README](https://github.com/hypothesis/lms) for instructions. It is literally just five commands:

1. `git clone https://github.com/hypothesis/lms.git`
2. `cd lms`
3. `make services`
4. `make devdata`
5. `make dev`

That's the entire set up. But maybe there's some problem that you're running into, that we haven't spotted before?

hmstepanek:
Ok managed to get it all setup this am and it looks good in speedgrader, html, and pdf assignments locally.  (Thank you for all the help!)

seanh:
Rebased

seanh (changes requested):
I think this needs some tests for escaping of special characters (such as spaces) in the query param keys and values.

Currently when proxying an HTML page, it seems to use `+` characters for escaping when making the `GET` request to look at the content type but then it uses `%` for escaping in the legacy Via URL that it redirects to.

For example if I go to <http://localhost:9082/> and paste in:

    http://example.com/?hi=there 1 2 3&via.open_sidebar=gar& foo = foo &bar=b a r

Then it sends a `GET` request to:

    http://example.com/?hi=there+1+2+3&+foo+=+foo+&bar=b+a+r

But it redirects the browser to:

    http://localhost:9080/http://example.com/?hi=there%201%202%203&via.open_sidebar=gar&%20foo%20=%20foo%20&bar=b%20a%20r

In the case of PDFs the `GET` request URL, the redirect (to `/pdf/<URL>`) and the PDF URL (that it configures PDF.js to download the PDF from) all seem to be the `+`-escaped version.

So I think only the redirect to Via for HTML pages is using `%`

I'm not sure which type of escaping we should prefer, `+` or `%`, but I think it should be consistent because, even though the two are _supposed_ to be equivalent, servers could in theory respond differently to each.

Maybe look at what type of escaping legacy Via uses and follow that.

We should add some tests for all three URLs when there are spaces in query params.

seanh:
> I'm not sure which type of escaping we should prefer, + or %, but I think it should be consistent because, even though the two are supposed to be equivalent, servers could in theory respond differently to each.

It looks to me like browsers use the `%20` approach. For example if you open `http://example.com/?foo= b a ar` in a browser it'll change to `http://example.com/?foo=%20b%20a%20a`. Or if you enter `http://example.com/?foo= b a ar` into http://localhost:9082/ then it comes out as `http://example.com/?foo=%20b%20a%20a` in Pyramid's `request.url` and `request.query_string` etc.

So maybe the solution is to get `urllib.parse.urlencode()` to output `%20` instead of `+`. According to [the docs](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) that can be done like this:

```python
urllib.parse.urlencode(..., quote_via=urllib.parse.quote)
```

seanh:
I've opened a separate issue for the encoding thing: https://github.com/hypothesis/py-proxy/issues/41

seanh (approved):

PR #37, opened 2020-01-17
Convert open_sidebar w/ asbool instead of int

hmstepanek:
asbool is a safer way of converting via.open_sidebar to a True/False value than using an int cast. Although the value does generally come from another Hypothesis owned tool, there's nothing stopping a user from misconfiguring a call and causing failures by providing some invalid string value like via.open_sidebar=foo.

### DIFF
--- a/py_proxy/views.py
+++ b/py_proxy/views.py
@@ -2,6 +2,7 @@
 import pyramid.httpexceptions as exc
 import requests
 from pyramid import response, view
+from pyramid.settings import asbool
 
 
 @view.view_config(renderer="py_proxy:templates/index.html.jinja2", route_name="index")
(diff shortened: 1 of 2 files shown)


py_proxy/views.py
```diff
@@ -43,7 +44,7 @@ def pdf(request):
     return {
         "pdf_url": f"{nginx_server}/proxy/static/{pdf_url}",
         "client_embed_url": request.registry.settings["client_embed_url"],
-        "h_open_sidebar": int(request.params.get("via.open_sidebar", "0")),
+        "h_open_sidebar": asbool(request.params.get("via.open_sidebar", False)),
```
seanh, line 47: Can we also, as part of this PR, change the default value from `False` to `True`? Currently if I go to https://via3.hypothes.is/ and paste in a PDF URL it'll proxy the PDF but not open the sidebar. While this is the same behaviour as legacy Via, it almost certainly isn't what users want (the point of proxying things through Via is supposed to be to annotate them or to view annotations).

Alternatively we could remove `via.open_sidebar` as an option entirely and just always pass `openSidebar: true` to the client. I'm not sure there's ever really a reason to proxy something through Via and _not_ open the sidebar.

Leaving the option in but just defaulting it to `True` is also fine, though

seanh (changes requested):
Looks good but let's just change the default

py_proxy/views.py
seanh: Opened a separate issue for this: https://github.com/hypothesis/py-proxy/issues/39

seanh (approved):

PR #46, opened 2020-01-21
Convert the app to run via pserve (not gunicorn directly)

jon-betts:
This is mostly for my own sanity as it means there is a dev INI config file which makes it very easy for me to debug in PyCharm. It also makes it more like our other products.

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -1,0 +1,12 @@
+[app:main]
+use = call:py_proxy.app:app
+
+nginx_server: http://localhost:9083
+client_embed_url: http://localhost:5000/embed.js
+legacy_via_url: http://localhost:9080
+
+[server:main]
+use: egg:gunicorn#main
+host: 0.0.0.0
+port: 9082
+timeout: 0
--- a/py_proxy/app.py
+++ b/py_proxy/app.py
@@ -16,6 +16,7 @@ def load_settings(settings):
 
     :param settings: Settings dict
     :raise ValueError: If a required parameter is not filled
+    :return: A dict of settings
     """
     for param in REQUIRED_PARAMS:
         value = settings[param] = settings.get(param, os.environ.get(param.upper()))
--- a/py_proxy/app.py
+++ b/py_proxy/app.py
@@ -5,40 +5,37 @@ import pyramid.config
 
 from py_proxy.sentry_filters import SENTRY_FILTERS
 
+REQUIRED_PARAMS = ["client_embed_url", "nginx_server", "legacy_via_url"]
 
-def settings():
-    """Return the app's configuration settings as a dict.
 
-    Settings are read from environment variables and fall back to hardcoded
-    defaults if those variables aren't defined.
+def load_settings(settings):
+    """Load application settings from a dict or environment variables.
 
+    Checks that the required parameters are either filled out in the provided
+    dict, or that the required values can be loaded from the environment.
+
+    :param settings: Settings dict
+    :raise ValueError: If a required parameter is not filled
     """
-    client_embed_url = os.environ.get("CLIENT_EMBED_URL")
-    nginx_server = os.environ.get("NGINX_SERVER")
-    legacy_via_url = os.environ.get("LEGACY_VIA_URL")
+    for param in REQUIRED_PARAMS:
+        value = settings[param] = settings.get(param, os.environ.get(param.upper()))
 
-    result = {
-        "client_embed_url": client_embed_url,
-        "nginx_server": nginx_server,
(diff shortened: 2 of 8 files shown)


jon-betts:
I've made a massive hash of this getting branches confused locally

py_proxy/app.py
```diff
@@ -5,40 +5,38 @@
...
 from py_proxy.sentry_filters import SENTRY_FILTERS
 
+REQUIRED_PARAMS = ["client_embed_url", "nginx_server", "legacy_via_url"]
 
-def settings():
-    """Return the app's configuration settings as a dict.
 
-    Settings are read from environment variables and fall back to hardcoded
-    defaults if those variables aren't defined.
+def load_settings(settings):
```
hmstepanek, line 11: Probably a nit-you can take it or leave it. When I see load_settings it makes me think they are being loaded into pyramid but really they are just being grabbed from the env/added to the dict of settings here. I think `get_settings` or `generate_settings` might make more sense in this scenario but I'll leave it up to you.

setup.py
```diff
@@ -0,0 +1,16 @@
+import os
```
hmstepanek, line 1: Do we really need this? It's not on pypi and I don't think we have any intention of putting it there nor is it needed for releasing to prod or locally running in dev as far as I know.

hmstepanek (commented):
Just some pre-lim thoughts. Thanks for adding running from pserve! I wasn't sure if the setup.py was really necessary?

setup.py
jon-betts: Yep. 

* This is what allows `pip install -e .` to work in `requirements-dev.in`
* ... making `py_proxy` available in the `tox` environment where `pserve` is located
* ... allowing `pserve` to find the class mentioned in the `conf/development.ini` lines:

```
[app:main]
use = call:py_proxy.app:app
```
*phew!*

To see this work (or not) have a go deleting the file and running `make dev`. It should fail with complaints about not being able to find `py_proxy`, unless its cached somehow.

py_proxy/app.py
jon-betts: I think "get_settings" is better, I'll change it. I just wanted to differentiate between *doing* something about settings and *being* the settings.

setup.py
seanh: Seems to be one of the things our apps are inconsistent about. LMS has a `setup.py`. So does Potemkin. h and bouncer somehow manage without one. The cookiecutters that come with Pyramid (https://github.com/Pylons/pyramid-cookiecutter-starter and https://github.com/Pylons/pyramid-cookiecutter-alchemy) also _do_ create a `setup.py`.

seanh: I suspect that h and bouncer not having a `setup.py` is another case of "we know about than Pyramid" or "we know better than the normal way of doing things" which kind of plagued Hypothesis for a while and has left quite some mess behind. I'm sure there _was_ a reason for it but can't remember what

seanh: h's `setup.py` was [removed in 2017](https://github.com/hypothesis/h/pull/4514) but it doesn't say why. Bouncer's was also [removed in 2017](https://github.com/hypothesis/bouncer/pull/36) as part of "converting bouncer to an application" because "it doesn't really make sense to keep bouncer as a Python package anymore in my opinion." One of the popular blog posts out there about Python packages vs Python applications will explain a way of doing apps without `setup.py` and why, I'm sure.

There seems to have been a general attitude that apps like bouncer and h shouldn't be installable Python packages because they aren't intended for other people to reuse them, aren't intended for redistribution or for uploading to PyPI (various brief comments to this effect in ancient Slack history if you search for "setup.py").

My opinion: Hypothesis isn't special enough to deviate from the normal way that Pyramid does things, especially on basics like this that we should not be wasting our time on. `setup.py` is _not_ only used for installing the package and listing its dependencies, as past comments seem to claim. Having no `setup.py` breaks things like Pyramid's builtin `pshell` command and other builtin scripts, breaks the standard way of adding your own custom command line scripts to a Pyamid app, and who knows what else. While I'm sure that can all be reinvented it seems like a waste of time to do so.

conf/development.ini
```diff
@@ -0,0 +1,12 @@
+[app:main]
+use = call:py_proxy.app:app
+
+nginx_server: http://localhost:9083
+client_embed_url: http://localhost:5000/embed.js
+legacy_via_url: http://localhost:9080
```
seanh, line 6: Is there still some code somewhere that supports reading these from envvars?

seanh: Oh yeah, spotted it below

seanh (approved):

PR #53, opened 2020-01-22
Caching views

jon-betts:
For https://github.com/hypothesis/py-proxy/issues/30: Implement appropriate caching for the root endpoint

This adds caching timeouts for the pdf and content_type end-points. The values are different depending on context and the rationalle is explained in:

 * https://github.com/hypothesis/py-proxy/blob/caching-views/docs/caching-strategy.md

### DIFF
--- a/docs/caching-strategy.md
+++ b/docs/caching-strategy.md
@@ -2,7 +2,7 @@
 
 ## Short `max-age`, long `stale-while-revalidate`
 
-One strategy we are using for caching the python calls is to provide both a
+One strategy we are using for caching the Python calls is to provide both a
 `max-age` and `stale-while-revalidate` value.
 
 Setting a `max-age` informs the consumer to not come back at all during the
--- a/docs/caching-strategy.md
+++ b/docs/caching-strategy.md
@@ -1,0 +1,77 @@
+# Caching strategy
+
+## Short `max-age`, long `stale-while-revalidate`
+
+One strategy we are using for caching the python calls is to provide both a
+`max-age` and `stale-while-revalidate` value.
+
+Setting a `max-age` informs the consumer to not come back at all during the
+period as the content is known 'good'. Setting a `stale-while-revalidate` as well
+tells the consumer that they can safely use the content in their cache _while_ they check for updates.
+
+### An example user journey
+
+This should result in the following sort of behavior:
+
+ * User requests a resource for the first time
+    * __Result:__ Cache is populated
+ * User requests the same resource very quickly
+    * __Result:__ User does not even make a second request due to `max-age`
+ * User requests the same resource after the `max-age` has expired but before
+   the `stale-while-revalidate` has expired
+   * __Result:__
+     * User is immediately served the old cached content
+     * _Then asynchronously_ the content is checked for validity
+     * The cache updated as necessary
+ * User requests the same resource very quickly after receiving the previous copy is found 'stale'
+   * __Result:__ User is returned the updated content retrieved asynchronously last time
+
+
+This means the user may see a stale version on the second try if the content has changed,
+but can quickly fix this simply checking again.
+
+### How this works to our advantage
+
+In a very crude overview the proxy works by:
+
+ 1. A Python call to determine the type (PDF or not) - Slowish
+ 2. A call through NGINX to serve the content - Fastish
+
+If this call is made repeatedly the flow is:
+
+ 1. Python call #1
+ 2. NGINX call #1 (user journey 1 complete)
+ 3. Python call #2
+ 4. NGINX call #2 (user journey 2 complete)
+ 5. Python call #3
+ 6. NGINX call #3 (user journey 3 complete)
+
+
+Using the `stale-while-revalidate` the sequence from the users point of view becomes
+
+ 1. Python call #1
+ 2. NGINX call #1 (user journey 1 complete)
+ 3. NGINX call #2 (user journey 2 complete _with contents of Python call #1_)
+ 4. Python call #2 (async)
+ 5. NGINX call #3 (user journey 3 complete _with contents of Python call #2_)
+ 6. Python call #3 (async)
+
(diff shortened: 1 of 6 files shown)


py_proxy/views.py
```diff
@@ -70,22 +74,26 @@ def content_type(request):
...
-    url = _drop_from_url_begining("/", request.path_qs)
-    return exc.HTTPFound(f"{via_url}/{url}")
+    url = request.path_qs.lstrip("/")
 
+    return exc.HTTPFound(f"{via_url}/{url}", headers=_caching_headers(max_age=60))
 
-def _drop_from_url_begining(drop_chars, url):
-    """Drop drop_chars from begining of url."""
-    drop_before = len(drop_chars)
-    return url[drop_before:]
```
jon-betts, line 88: This is not good. If I read this correctly it just removes *any* number of characters from the start matching what you pass in. I've replaced this with `lstrip` which I assume was what this was meant to do.

py_proxy/views.py
```diff
@@ -70,22 +74,26 @@ def content_type(request):
...
+
+def _caching_headers(max_age, stale_while_revalidate=86400):
+    # I tried using webob.CacheControl for this but it's total rubbish
+    header = (
+        f"public, max-age={max_age}, stale-while-revalidate={stale_while_revalidate}"
+    )
```
jon-betts, line 112: I tried using `webob.CacheControl`, but that module is a total disaster. It will cheerfully create an invalid `Cache-Control` header if you send it even mildly the wrong thing. It's not worth it.

docs/caching-strategy.md
```diff
@@ -0,0 +1,77 @@
+# Caching strategy
+
+## Short `max-age`, long `stale-while-revalidate`
+
+One strategy we are using for caching the python calls is to provide both a
```
seanh:
```suggestion
One strategy we are using for caching the Python calls is to provide both a
```

docs/caching-strategy.md
```diff
@@ -0,0 +1,77 @@
...
+## Different timeouts on different calls
+
+The different calls have different time-outs as we can rely on some checks to
+effectively invalidate others.
+
+#### `content_type()` view
+
+* `max-age`: PDF: 5m, HTML 60s
+* PDFs should be relatively static, and take a while to update and change
+* HTML could be served from a Wiki or other volatile source
```
seanh:
```suggestion
* HTML could be served from a wiki or other volatile source
```

py_proxy/views.py
```diff
@@ -70,22 +74,26 @@ def content_type(request):
...
     with requests.get(url, stream=True, allow_redirects=True) as rsp:
-        if rsp.headers.get("Content-Type") in ("application/x-pdf", "application/pdf",):
+        if rsp.headers.get("Content-Type") in ("application/x-pdf", "application/pdf"):
             return exc.HTTPFound(
                 request.route_url(
-                    "pdf", pdf_url=request.matchdict["url"], _query=request.params
-                )
+                    "pdf", pdf_url=request.matchdict["url"], _query=request.params,
+                ),
+                headers=_caching_headers(max_age=300),
```
seanh, line 84: What happens if the response from `requests.get()` is an unsuccessful response? We're asking requests to follow redirects so I think 3xx's are out. But what if we receive a 4xx or 5xx response (or a 1xx, which would be weird)? These would seem to indicate that something went wrong, so we probably wouldn't want to cache them.

Could be dealt with as a separate issue

py_proxy/views.py
```diff
@@ -70,22 +74,26 @@ def content_type(request):
...
             )
 
     via_url = request.registry.settings["legacy_via_url"]
-    url = _drop_from_url_begining("/", request.path_qs)
-    return exc.HTTPFound(f"{via_url}/{url}")
+    url = request.path_qs.lstrip("/")
 
+    return exc.HTTPFound(f"{via_url}/{url}", headers=_caching_headers(max_age=60))
```
seanh: Same comment for the caching headers here about unsuccessful responses

tests/unit/py_proxy/views_test.py
```diff
@@ -3,10 +3,26 @@
 import httpretty
 import pytest
 from bs4 import BeautifulSoup
+from h_matchers import Any
 from jinja2 import Environment, FileSystemLoader
 from pkg_resources import resource_filename
+from webtest import TestApp as WebTestApp  # No pytest, this isn't a test class
```
seanh, line 10: I'd suggest just `import webtest` and then use it as `webtest.TestApp`, instead of renaming it

tests/unit/py_proxy/views_test.py
```diff
@@ -3,10 +3,26 @@
...
+    # pylint: disable=no-value-for-parameter
+    # Pylint doesn't seem to understand h_matchers here for some reason
+    assert dict(headers) == Any.dict.containing({"Cache-Control": Any.string()})
+
+    header = headers["Cache-Control"]
+    parts = sorted(header.split(", "))
+
+    assert "public" in parts
+    assert f"max-age={max_age}" in parts
+    assert f"stale-while-revalidate={stale_while_revalidate}" in parts
```
seanh, line 28: This feels like it could itself be a matcher: `assert response == Any.response.with.headers(max_age=..., stale_while_revalidate=...)`

seanh (approved):
Looks good. We may want to separately look into _not_ caching unsuccessful responses?

py_proxy/views.py
jon-betts: See code below?

tests/unit/py_proxy/views_test.py
jon-betts: They aren't actually headers, they are values in the header. It's like this:

`Cache-Control: public, max-age=60, stale-while-revalidate=4853854`

But yes, I'm sure we could put something together

PR #58, opened 2020-01-23
Adding gzip and Brotli compression to the static files

jon-betts:
This is for: https://github.com/hypothesis/via3/issues/42

To test this you can:

* Run `make dev`
* Make the following request
```
curl -I --request GET --url http://0.0.0.0:9082/favicon.ico  --header 'accept-encoding: br, gzip'
```
* You should see no "Content-Encoding" in the response
* Run `make build`
* Restart the dev server
* Make the same request, you should now see "Content-Encoding: br"

----

The result is most dramatic on the largest files:

```
curl -I --request GET --url http://0.0.0.0:9082/vendor/pdfjs-2/build/pdf.worker.js --header 'accept-encoding: br, gzip'
```
Content-Length: 233825

```
curl -I --request GET --url http://0.0.0.0:9082/vendor/pdfjs-2/build/pdf.worker.js --header 'accept-encoding: gzip'
```
Content-Length: 308992

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 .tox
 *.pyc
 *.egg-info/
+py_proxy/static/**/*.gz
+py_proxy/static/**/*.br
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.egg-info/
 py_proxy/static/**/*.gz
 py_proxy/static/**/*.br
+build.tar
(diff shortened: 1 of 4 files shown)


seanh (approved):

jon-betts:
This has been all round the houses, but the final state is:

 * We enabled gzip compression on our end, but it only matters between us and Cloudflare
 * We enabled Brotli compression in general at the Cloudflare level
 * As it's on the fly, it's not as aggressive as our ahead of time Brotli, and so doesn't get the same benefits (10% rather than say 33% in some cases)
 * We abandoned our own Brotli because Cloudflare totally ignores it and it's slow

PR #65, opened 2020-01-27
Content type error catching

jon-betts:
Before this change any URL given to the 'content_type()` end-point would fail if it:

 * Wasn't a URL
 * Didn't have a scheme or a non HTTP scheme
 * Had a domain that could not be resolved
 * Timed out or had any other issue connecting

We now catch these errors and serve very basic error pages for them.

### DIFF
--- a/py_proxy/views.py
+++ b/py_proxy/views.py
@@ -75,21 +75,6 @@ def pdf(request):
     }
 
 
-def _get_url_details(url):
-    try:
-        with requests.get(url, stream=True, allow_redirects=True) as rsp:
-            return rsp.headers.get("Content-Type"), rsp.status_code
-
-    except REQUESTS_BAD_URL as err:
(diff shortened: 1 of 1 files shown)


jon-betts:
For https://github.com/hypothesis/via3/issues/23

jon-betts:
I reviewed this with @seanh on Friday over video chat, so as soon as I can get the tests etc. to pass I'll merge.

jon-betts:
I'm seeing some weird behavior where tests are failing due to coverage on Jenkins but not locally...

jon-betts:
Yeah I don't get this. The code in Jenkins is different and has two copies of a function. It looks like it's not applied a diff correctly and is caching it somehow.

jon-betts:
Not sure what happened there, I think I might have merged most, but not all of this before.

PR #78, opened 2020-01-29
Rename and split views into separate files

jon-betts:
This changes the view names from nouns (e.g. "pdf") to verbs that describe what they let you do with them (e.g. "view_pdf") and splits the views into separate files.

This follows a pattern we've discussed but not fully implemented in LMS where:

 * views/*_view(s).py - Are views
 * views/*anything else* - Regular code

----

This should be merged after https://github.com/hypothesis/via3/pull/77


seanh:
Can we try [this approach](https://github.com/hypothesis/lms/issues/1245#issuecomment-578640771) instead?

* views/monitoring_views.py -> views/monitoring.py
* views/route_by_content_views.py -> views/route_by_content.py
* views/view_pdf_view.py -> views/view_pdf.py
* views/query_params.py -> views/_query_params.py

jon-betts:
Ok, looks like this has moved on since we discussed this. So I'll jigger this about again.

A nice effect of the above is all views are grouped vs non views (prefixed with _)

tests/unit/conftest.py
```diff
@@ -73,3 +74,11 @@ def httpretty_():
...
     httpretty.disable()
     httpretty.reset()
+
+
+def assert_cache_control(headers, cache_parts):
+    """Assert that all parts of the Cache-Control header are present."""
+    assert dict(headers) == Any.dict.containing({"Cache-Control": Any.string()})
+    assert (
+        headers["Cache-Control"].split(", ") == Any.list.containing(cache_parts).only()
+    )
```
seanh, line 84: This could be made a pytest fixture and then you wouldn't need to import it, plus you could override and customize if if you ever needed to:

```python
@pytest.fixture
def assert_cache_control():
    """Return a function that asserts that all parts of the Cache-Control header are present."""

    def assert_cache_control(headers, cache_parts):
        <same function as before>

    return assert_cache_control
```

seanh (approved):

tests/unit/conftest.py
jon-betts: On that day then maybe. I don't think it's bad to import it and it complicates a simple function.

PR #79, opened 2020-01-29
Fix query string bugs relating to repeated elements

jon-betts:
* Also separate out testing of the component
 * Remove duplicated testing in other places
 * Use the same code more times
   * Rather than inventing a new way at the end of get_content_type()

### DIFF
--- a/tests/unit/via/views/test_route_by_content.py
+++ b/tests/unit/via/views/test_route_by_content.py
@@ -13,57 +13,11 @@ from requests.exceptions import (
 
 from tests.unit.conftest import assert_cache_control
 from via.exceptions import BadURL, UnhandledException, UpstreamServiceError
+from via.views._query_params import QueryParams
 from via.views.route_by_content import route_by_content
 
 
 class TestRouteByContent:
-    @pytest.mark.parametrize(
-        "requested_path,expected_location,content_type",
-        [
-            # If the requested pdf URL has no query string then it should just
-            # redirect to the requested URL, with no query string (but with the
-            (
-                "/https://example.com/foo",
-                "http://localhost/pdf/https://example.com/foo",
-                "application/pdf",
-            ),
-            # If the requested pdf URL has a query string then the query string
-            # should be preserved in the URL that it redirects to.
-            (
-                "/https://example.com/foo?bar=baz",
-                "http://localhost/pdf/https://example.com/foo?bar=baz",
-                "application/pdf",
-            ),
-            # If the requested html URL has a query string then the query string
-            # should be preserved in the URL that it redirects to.
-            (
-                "/https://example.com/foo?bar=baz",
-                "http://via.hypothes.is/https://example.com/foo?bar=baz",
-                "text/html",
-            ),
-            # If the requested html URL has a client query params then the query params
-            # should be preserved in the URL that it redirects to.
-            (
-                "/https://example.com/foo?via.open_sidebar=1",
-                "http://via.hypothes.is/https://example.com/foo?via.open_sidebar=1",
-                "text/html",
-            ),
-        ],
-    )
-    def test_redirect_location(
-        self, make_pyramid_request, requested_path, expected_location, content_type
-    ):
-
-        request = make_pyramid_request(
-            request_url=requested_path,
-            thirdparty_url="https://example.com/foo",
-            content_type=content_type,
-        )
-
-        redirect = route_by_content(request)
-
-        assert redirect.location == expected_location
-
     @pytest.mark.parametrize(
         "content_type,redirect_url",
         [
@@ -85,10 +39,8 @@ class TestRouteByContent:
 
         assert result.location == redirect_url
 
-    @pytest.mark.parametrize(
-        "query_param", ["via.request_config_from_frame", "via.open_sidebar"]
-    )
-    def test_does_not_pass_via_query_params_to_thirdparty_server(
+    @pytest.mark.parametrize("query_param", QueryParams.ALL_PARAMS)
+    def test_does_not_pass_via_query_params_to_third_parties(
         self, make_pyramid_request, query_param
     ):
         request = make_pyramid_request(
--- a/via/views/_query_params.py
+++ b/via/views/_query_params.py
@@ -1,37 +1,35 @@
 """Methods and data relating to query parameters."""
 
-from urllib.parse import urlencode
+from urllib.parse import urlencode, urlparse
 
-# Client configuration query parameters.
-OPEN_SIDEBAR = "via.open_sidebar"
-CONFIG_FROM_FRAME = "via.request_config_from_frame"
+from webob.multidict import MultiDict
 
 
-def strip_client_query_params(base_url, query_params):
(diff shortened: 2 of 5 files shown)


tests/unit/via/views/test_route_by_content.py
```diff
@@ -17,52 +17,23 @@
...
-                "/https://example.com/foo?bar=baz",
-                "http://via.hypothes.is/https://example.com/foo?bar=baz",
-                "text/html",
-            ),
-            # If the requested html URL has a client query params then the query params
-            # should be preserved in the URL that it redirects to.
-            (
-                "/https://example.com/foo?via.open_sidebar=1",
-                "http://via.hypothes.is/https://example.com/foo?via.open_sidebar=1",
-                "text/html",
```
jon-betts, line 49: All of these tests amount to the same thing, and are replicated lower down. We always pass all params onto the final request. It only changes when we call the original URL to check the resource content type.

via/views/route_by_content.py
```diff
@@ -12,15 +12,17 @@
...
     UpstreamServiceError,
 )
-from via.views._query_params import strip_client_query_params
+from via.views._query_params import QueryParams
 
 
 @view.view_config(route_name="route_by_content")
 def route_by_content(request):
     """Routes the request according to the Content-Type header."""
-    url = strip_client_query_params(request.matchdict["url"], request.params)
```
jon-betts, line 21: This `url` parameter is re-used later on and confused the heck out of me about when we do or don't strip via params

via/views/view_pdf.py
```diff
@@ -20,12 +16,14 @@
...
+        request.matchdict["pdf_url"], request.params, strip_via_params=True
+    )
 
     return {
         "pdf_url": f"{nginx_server}/proxy/static/{pdf_url}",
         "client_embed_url": request.registry.settings["client_embed_url"],
-        "h_open_sidebar": asbool(request.params.get(OPEN_SIDEBAR, False)),
-        "h_request_config": request.params.get(CONFIG_FROM_FRAME, None),
+        "h_open_sidebar": asbool(request.params.get(QueryParams.OPEN_SIDEBAR, False)),
+        "h_request_config": request.params.get(QueryParams.CONFIG_FROM_FRAME, None),
```
jon-betts, line 27: This bit of processing could be moved to QueryParams, but then:

 * This code wouldn't have anything to do and would get sad
 * The code wouldn't get any shorter as we'd have to unpack whatever we were sent

via/views/_query_params.py
```diff
@@ -1,37 +1,35 @@
...
+        :param query: Dict of query parameters
+        :param strip_via_params: Enable removal of via params
+        :return: A new URL with the relevant query params added
+        """
+
+        # Coerce any immutable NestedMultiDict's into mutable a MultiDict
+        if strip_via_params:
+            query = MultiDict(query)
+
+            bad_keys = [key for key in query.keys() if key.lower() in cls.ALL_PARAMS]
```
jon-betts: Process the bad ones up front so we don't change the thing we are iterating over

seanh: Could be `via_keys` maybe

via/views/_query_params.py
```diff
@@ -1,37 +1,35 @@
...
-    return base_url
+from urllib.parse import urlencode, urlparse
+
+from webob.multidict import MultiDict
+
+
+class QueryParams:
+    """Client configuration query parameters and related functions."""
+
+    # pylint: disable=too-few-public-methods
```
seanh, line 11: Despite pylint's complaint I like this being in a class, it's nice to use a class as a namespace to group together a function and some constants. I suppose pylint is right that they're already grouped in the module (which only contains this one class) but :man_shrugging:

tests/unit/via/views/test_route_by_content.py
```diff
@@ -17,52 +17,23 @@
...
-            ),
-        ],
-    )
-    def test_redirect_location(
-        self, make_pyramid_request, requested_path, expected_location, content_type
-    ):
-
+    def test_we_do_not_lookup_via_parameters(self, make_pyramid_request, requests):
```
seanh: The test name confused me. This is testing that we don't pass the `via.*` params to the third-party server, right? Maybe `test_we_dont_send_via_params_to_third_parties`?

via/views/route_by_content.py
seanh: Honestly just writing `request.matchdict["url"]` every time you want to use it, rather than assigning a variable name to it, would probably make the code easier to follow

seanh: It actually is already referenced as `request.matchdict["url"]` in one place further down in the function, but referenced as `path_url` in others

via/views/route_by_content.py
```diff
@@ -49,9 +51,13 @@ def route_by_content(request):
         headers = {"Cache-Control": "no-cache"}
 
     via_url = request.registry.settings["legacy_via_url"]
-    url = request.path_qs.lstrip("/")
 
-    return exc.HTTPFound(f"{via_url}/{url}", headers=headers)
+    return exc.HTTPFound(
+        QueryParams.build_url(
+            f"{via_url}/{path_url}", request.params, strip_via_params=False
```
seanh, line 55:
```suggestion
            f"{via_url}/{path_url}", request.params
```

It defaults to `False`

seanh (approved):

via/views/route_by_content.py
jon-betts: Yeah, I think I might remove the default to force you to say. This was to make crystal clear the intent.

via/views/route_by_content.py
jon-betts: I'm a bit confused about what you're saying here, but to clarify:

 * Previously `url` was used twice, for different purposes obtained in different ways
 * We now get the `url` once, and use the same method for constructing the final URL with slight differences

tests/unit/via/views/test_route_by_content.py
jon-betts: Oh god. I've just spotted this whole test is replicated again lower down. So the first test was already duplicating things in here twice. I've just removed mine entirely and left the other there.

via/views/_query_params.py
jon-betts: Yep. That's better. We can't dumbly iterate, as theoretically they could in the list 100 times. With multidict you'd have to pop them 100 times to get rid of them. Edge-casey, but might as well do it right.

via/views/_query_params.py
jon-betts: Me too. I get the idea, but at an emotional level I find classes tidier. A module tends to accumulate things which may, or may not be like each other. Better to over organise first than under.

via/views/route_by_content.py
jon-betts: Ah I get it. There's a _third_ place I hadn't spotted. I've fixed this now.

In the first iteration of this there were about 4 different URLs floating about of which 2 had the same name etc.

jon-betts:
So after some more research it turns out that the query parameters are case sensitive. I've changed the PR to reflect that and the review points above.

PR #82, opened 2020-01-30
redirect nginx access log 

dmfine:
to stdout so that it gets written to the Elastic Beanstalk host's logs (which then forward to paper trail)


LMS007 (approved):

seanh:
This causes NGINX to crash on startup, completely breaking the app. It had taken down QA

Please test your PRs locally before merging them, and after merging a PR please test it on QA and then either deploy it to prod or revert it.

PR #84, opened 2020-01-31
Try more parallel builds to see the effect

jon-betts:


jon-betts:
The combined changes here appear to take the build time (not including release etc.) from ~2.5 minutes to ~1.5. Which is quite significant when developing.

tox.ini
```diff
@@ -18,6 +18,10 @@ tox_pip_extensions_ext_venv_update = true
 tox_pyenv_fallback = false
 
 [testenv]
+envdir =
+    {tests,coverage}: {toxworkdir}/monotest
```
seanh:
```suggestion
    {tests,coverage}: {toxworkdir}/tests
```

But I'm not sure about using `envdir` to make one testenv use the same venv as another, anyway. It's unclear to me from the tox docs what `envdir` is for and whether it's really meant to be used in this way. It's possible that it is meant to be. But regardless it does break at least one thing:

When both the testenv's are run in parallel, it now crashes. For example using tox's `-p` for parallelism, if the `monotest` venv hasn't been created yet, will crash:

```
tox -p all -e tests,coverage
```

Or using make's `-j` to do the same thing:

```
make -j 2 test coverage
```

Or simply running them as two separate `tox` or `make` commands in two separate shells at the same time, etc.

I know that it doesn't make sense to ever run tests and coverage at the same time since coverage depends on the result of tests. But that suggests that tests and coverage should actually be one testenv. And still I don't think we should combine their venvs in this way as it might encourage using this method to combine other venvs as well, and because it might also break other tox things besides parralelism.

tox.ini
```diff
@@ -18,6 +18,10 @@ tox_pip_extensions_ext_venv_update = true
...
+    {.tox,dev,lint,format,checkformatting,docstrings,checkdocstrings,build,pip-compile,docker-compose,update-pdfjs}: {toxworkdir}/{envname}
```
seanh: I think `{toxworkdir}/{envname}` is the default so this line should be unnecessary?

seanh (changes requested):
On my local machine this seems to take about 3 seconds off the time taken to run `tox -e tests,coverage`. It goes from 19 to 16 secs.

But it's even faster if you instead combine the coverage and tests _testenvs_ into one, instead of having two separate testenvs that share the same venv, so just `tox -e tests` on its own (or just `tox` since `tests` is the default testenv) runs both the tests and then coverage. it shaves off about 4.5 secs total compared to master.

And it avoids potential problematicness of making two testenvs share the same venv that I noted above.

It's also a much simpler change to make and results in a simpler `tox.ini` and `Makefile`. On master `coverage` is already being installed into the `tests` testenv, all you need to do is move the `coverage combine` and `coverage report` into `tests` and delete the `coverage` testenv. `make coverage` has to be deleted from `Makefile` since `make test` now does it, which simplifies `Makefile`:

```diff
diff --git a/Makefile b/Makefile
index c0f8d36..18a766b 100644
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,6 @@ help:
 	@echo "make format            Correctly format the code"
 	@echo "make checkformatting   Crash if the code isn't correctly formatted"
 	@echo "make test              Run the unit tests"
-	@echo "make coverage          Print the unit test coverage report"
 	@echo "make docstrings        View all the docstrings locally as HTML"
 	@echo "make checkdocstrings   Crash if building the docstrings fails"
 	@echo "make update-pdfjs      Update our copy of PDF-js"
@@ -52,10 +51,6 @@ test: python
 update-pdfjs: python
 	@tox -qe update-pdfjs
 
-.PHONY: coverage
-coverage: python
-	@tox -qe coverage
-
 .PHONY: docstrings
 docstrings: python
 	@tox -qe docstrings
diff --git a/tox.ini b/tox.ini
index d7ffd88..62db343 100644
--- a/tox.ini
+++ b/tox.ini
@@ -40,7 +40,6 @@ deps =
     lint: pydocstyle
     {format,checkformatting}: black
     {format,checkformatting}: isort
-    coverage: coverage
     docstrings: sphinx-autobuild
     {docstrings,checkdocstrings}: sphinx
     docker-compose: docker-compose
@@ -63,11 +62,11 @@ commands =
     checkformatting: black --check via tests bin
     checkformatting: isort --recursive --quiet --check-only via tests bin
     tests: coverage run -m pytest {posargs:tests/unit/}
+    tests: -coverage combine
+    tests: coverage report
     {docstrings,checkdocstrings}: sphinx-apidoc -ePMF -a -H "Dooccsstrinngs!!" --ext-intersphinx --ext-todo --ext-viewcode -o {envtmpdir}/rst .
     docstrings: sphinx-autobuild -BqT -z via -z tests -b dirhtml {envtmpdir}/rst {envtmpdir}/dirhtml
     checkdocstrings: sphinx-build -qTn -b dirhtml {envtmpdir}/rst {envtmpdir}/dirhtml
-    coverage: -coverage combine
-    coverage: coverage report
     pip-compile: pip-compile {posargs}
     update-pdfjs: sh bin/update-pdfjs
     update-pdfjs: python bin/create_pdf_template.py
```

This does mean that you can't run the tests without also running coverage, but I think that's fine.

You also can't run coverage without running the tests. That's not only fine but possibly even a good thing since previously running `make coverage` without first running `make test` would either fail because coverage needs the data from `make test` or would print out a coverage report based on a previous run of `make test` that might be out of date compared to the other code.

tox.ini
jon-betts: If you don't have this comment then you have set the default to `None` which means `.`, which causes tox to fais complaining it will overwrite your project directory.

I wonder if you can just put `{toxworkdir}/{envname}` by itself without specifying the test env to re-assert the default?

seanh: I think we should just avoid using `envdir` anyway, the alternative is both simpler and faster (see comment above)

seanh:
@jon-betts Ping

jon-betts:
Ok, this has been reworked substantially. We now run coverage and tests together and limit the output from coverage with skip-covered.

Just need to see if the tests still pass etc.

jon-betts:
This still retains all the speed benefits from before, so I think this is a good enough solution without tox convolutions.

seanh (approved):

PR #86, opened 2020-01-31
Add a script for minifying assets

jon-betts:
The effect of this is a little hard to guage, as Cloudflare has Brotli compression on and we only have gzip locally. But most things lost about 25% of the size. The best lost 75% and our main template is down by 30%. 

What this actually relates to when released to the wild is anyone's guess once Brotli is applied.

### Review notes

 * Holy smokes: -100k lines!
 * Don't worry though, all the code is in the first commit, minification in the second
 * The script is run as part of the 'build' make target

----

This is for: https://github.com/hypothesis/via3/issues/45

This should be rebased and merged after https://github.com/hypothesis/via3/pull/85


seanh (approved):
:+1: My only question is whether this works with `tools/update-pdfjs` (https://github.com/hypothesis/via3#updating-the-pdf-viewer)?

PR #97, opened 2020-02-05
add access log forwarding to stdout 

dmfine:
So that it will be recorded on the docker host and forwarded to papertrail.

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -33,7 +33,7 @@ http {
         include /var/lib/hypothesis/nginx_envsubst.conf;
         listen 9083;
         merge_slashes off;
-
+        access_log /dev/stdout;
         location ~ /proxy/static/(?<proxied_uri>.*) {
             set $upstream $proxied_uri$is_args$args;
             proxy_ssl_server_name on;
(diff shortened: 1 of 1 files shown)


dmfine:
The previous PR had the access_log directive at the top level, which it turns out is not allowed. Nginx starts when it's under the server block. QA has some issue at the moment which is preventing me from testing it as thoroughly as I'd like.

seanh:
The commit was based off an old, broken version of the master branch. I've rebased it for you

conf/nginx.conf
```diff
@@ -33,7 +33,7 @@ http {
         include /var/lib/hypothesis/nginx_envsubst.conf;
         listen 9083;
         merge_slashes off;
-
+        access_log /dev/stdout;
```
seanh: What's the difference between putting this in the `server { ... }` block, vs putting it higher up in the `http { ... }` block? It looks like putting it in the `server` block might mean that it applies only to that one virtual server, when the `nginx.conf` might contain multiple `server`'s in the future.

seanh: Yeah, this should be in the `http { ... }`. See https://www.nginx.com/resources/wiki/start/topics/examples/full/ and https://nginx.org/en/docs/http/ngx_http_log_module.html#access_log. I've pushed a commit to fix it

seanh (approved):

conf/nginx.conf
dmfine: future proofing +1

PR #100, opened 2020-02-10
Disable cache for redirects only

### LINKED ISSUE #98: Fix expiring redirect being cached
A bug has been raised in LMS about Canvas files not working correctly:

* https://github.com/hypothesis/lms/issues/1448

We currently believe the cause of this is due to the following:

 * LMS gets a URL passed to it from Canvas (URL A) which is a redirect to the actual PDF (URL B)
 * This is passed to via3 where the python app checks the file type of URL A
 * The python app follows URL A to discover the file is a PDF
 * The python app then sets the page content to load the PDF from URL A
 * This is served via NGINX which returns a modified version of the redirect content of URL A, effectively becoming a redirect to a Via 3 wrapped version of URL B
 * NGINX adds headers so that this redirect is cached for 1 day
 * URL B expires
 * Users and Cloudflare continue to serve our version of URL B

----

# Fixing this

There are two parts to the fix:

 * [ ] First of all we are going to disable caching because we can do this quickly
 * [ ] Then we are going to try and get NGINX to disable caching for redirects only

jon-betts:
Details in https://github.com/hypothesis/via3/issues/98, but basically our "PDF" URL can in fact be a redirect to a PDF etc. This could refer to a resource that expires, so it's not safe to signal it can be cached.

### DIFF
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -34,33 +34,29 @@ http {
         include /var/lib/hypothesis/nginx_envsubst.conf;
         listen 9083;
         merge_slashes off;
+
         location ~ /proxy/static/(?<proxied_uri>.*) {
-            set $upstream $proxied_uri$is_args$args;
-            proxy_ssl_server_name on;
-            proxy_pass $upstream;
-            proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/$1;
+            proxy_intercept_errors on;
+            error_page 301 302 307 = @handle_redirect;
 
-            # Strip hypothesis cookies and authorization header.
-            set $stripped_cookie $http_cookie;
+            include via/direct_proxy.conf;
 
-            if ($stripped_cookie ~ "(.*)\s*auth=[^;]+;?(.*)") {
-                set $stripped_cookie $1$2;
-            }
-            if ($stripped_cookie ~ "(.*)\s*session=[^;]+;?(.*)") {
-                set $stripped_cookie $1$2;
-            }
-            proxy_set_header Cookie $stripped_cookie;
-            proxy_set_header Authorization "";
+            # Cache for a 1 day, but allow serving from cache while revalidating for a week
+            proxy_hide_header "Cache-Control";
+            add_header "Cache-Control" "public, max-age=86400, stale-while-revalidate=604800";
+
+            add_header "X-Via" "static-proxy, direct";
+        }
+
+        location @handle_redirect {
+            include via/direct_proxy.conf;
 
             # Disable caching to prevent Cloudflare from kicking in where this
             # might be a temporary redirect to an expiring PDF link
             proxy_hide_header "Cache-Control";
             add_header "Cache-Control" "no-cache, no-store, must-revalidate";
 
-            # Do not allow the third party server to set cookies.
-            add_header "Set-Cookie" "";
-            proxy_hide_header "Access-Control-Allow-Origin";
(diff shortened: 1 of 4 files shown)


conf/nginx/nginx.conf
```diff
@@ -34,33 +34,29 @@ http {
...
         listen 9083;
         merge_slashes off;
+
         location ~ /proxy/static/(?<proxied_uri>.*) {
-            set $upstream $proxied_uri$is_args$args;
-            proxy_ssl_server_name on;
-            proxy_pass $upstream;
-            proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/$1;
+            proxy_intercept_errors on;
+            error_page 301 302 307 = @handle_redirect;
```
seanh, line 40: Not 303 or 308?

conf/nginx/nginx.conf
```diff
@@ -34,33 +34,29 @@ http {
...
+
+            include via/direct_proxy.conf;
```
seanh, line 42: Ah, that's handy

conf/nginx/via/direct_proxy.conf
```diff
@@ -0,0 +1,21 @@
...
+if ($stripped_cookie ~ "(.*)\s*session=[^;]+;?(.*)") {
+    set $stripped_cookie $1$2;
+}
+proxy_set_header Cookie $stripped_cookie;
+proxy_set_header Authorization "";
+
+# Do not allow the third party server to set cookies.
+add_header "Set-Cookie" "";
+proxy_hide_header "Access-Control-Allow-Origin";
+add_header "Access-Control-Allow-Origin" $access_control_allow_origin;
```
seanh, line 21: So this is just all the proxying code that we already had before, minus adding the caching headers, just extracted into a separate file so it can be reused via `include`

conf/nginx/via/direct_proxy.conf
```diff
@@ -0,0 +1,21 @@
+set $upstream $proxied_uri$is_args$args;
+proxy_ssl_server_name on;
+proxy_pass $upstream;
+proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/$1;
```
seanh, line 4: This line may not be needed in the `location ~ /proxy/static/(?<proxied_uri>.*)` now (only in the `location @handle_redirect`) since, if it is a redirect, it's gonna hit the `error_page 301 302 307 = @handle_redirect` line?

conf/nginx/nginx.conf
```diff
@@ -34,33 +34,29 @@ http {
...
+
+            # Cache for a 1 day, but allow serving from cache while revalidating for a week
+            proxy_hide_header "Cache-Control";
+            add_header "Cache-Control" "public, max-age=86400, stale-while-revalidate=604800";
 
-            # Strip hypothesis cookies and authorization header.
-            set $stripped_cookie $http_cookie;
+            add_header "X-Via" "static-proxy, direct";
```
seanh, line 48: :+1:

seanh (approved):
Tested it locally in Canvas and it seems to work :+1: The redirect from Canvas's file server isn't cached but the ultimate PDF response is

conf/nginx/via/direct_proxy.conf
jon-betts: Yep

conf/nginx/via/direct_proxy.conf
jon-betts: Maybe? I'd need to test it, but seems likely

PR #102, opened 2020-02-10
Encode the path so we reliably pass on encoded values

### LINKED ISSUE #104: View PDF end-point should preserve encoded chars (including commas)
TODO: Fill this out

### LINKED ISSUE #105: Route by content end-point should preserve encoded chars (including commas)
Things to test:

* Does it request the URL correctly when checking the type?
* If it's PDF does it redirect correctly?
* If it's not a PDF does it redirect correctly?

As commas and spaces behave differently, we should check that we preserve escaping for both if present.

We believe it is ok to add encoding where it was not before, but not ok to remove it.

jon-betts:
Things to test:

Route by content:

* Does it request the URL correctly when checking the type?
* If it's PDF does it redirect correctly?
* If it's not a PDF does it redirect correctly?

View PDF:

 * TBA

As commas and spaces behave differently, we should check that we preserve escaping for both if present.

We believe it is ok to add encoding where it was not before, but not ok to remove it.

### DIFF
--- a/via/views/_query_params.py
+++ b/via/views/_query_params.py
@@ -1,6 +1,6 @@
 """Methods and data relating to query parameters."""
 
-from urllib.parse import urlencode, urlparse
+from urllib.parse import quote, unquote, urlencode, urlparse
 
 from webob.multidict import MultiDict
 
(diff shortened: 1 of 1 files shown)


jon-betts:
We are not going this way at all. So this is dead code.

PR #107, opened 2020-02-11
Stop NGINX from decoding URL encoding in the path

### LINKED ISSUE #103: NGINX should not decode encoded commas in the /proxy/static/ endpoint
### Bug

Canvas files with commas in their filenames don't work with Via 3's NGINX "static" proxy endpoint.

### Steps to reproduce

1. Create a PDF file locally with a comma in its filename

2. Upload the file to a Canvas course.

3. Get a public download URL for the file.

   One way to get this is using the Canvas API. You first need to [get an API access token](https://github.com/hypothesis/lms/wiki/How-to-Test-the-Canvas-API) for a Canvas user who has access to the file. You can then list all of a course's files like this, substituting your `<COURSE_ID>` and `<ACCESS_TOKEN>` (the course ID for _Test Course 1_ is **83**)

   ```terminal
   http 'https://hypothesis.instructure.com/api/v1/courses/<COURSE_ID>/files' Authorization:'Bearer <ACCESS_TOKEN>'
   ```

   You can then get a public download for a particular file like this, substituting your `<FILE_ID>` and `<ACCESS_TOKEN>` (**342** is the file ID of one file that has commas in its filename):

   ```terminal
   http 'https://hypothesis.instructure.com/api/v1/files/<FILE_ID>/public_url' Authorization:'Bearer <ACCESS_TOKEN>'
   ```

4. Notice that the URL from the Canvas API contains the PDF's filename in it and that **commas have been encoded as `%2C` in the URL**.

5. Also be aware that these Canvas file download URLs expire, so you'll need to get a fresh one after a while.

5. Attempt to proxy this public download URL with commas-encoded-as-`%2C` in it through Via 3's `/proxy/static/` endpoint, which is handled by Via 3's NGINX directly (no Python). For example:

   ```
   http://localhost:9083/proxy/static/https://inst-fs-iad-prod.inscloudgate.net/files/.../....pdf?token=eyJ...tUw
   ```

   Or:
   
   ```
   https://qa-vi3.hypothes.is/proxy/static/https://inst-fs-iad-prod.inscloudgate.net/files/.../....pdf?token=eyJ...tUw
   ```

6. You'll see an HTML page with the error message "JWT rejected: resource claim does not match URL path"

   This HTML page and error message are from Canvas's file server. They're the response that Canvas's file server gave to Via 3's NGINX when NGINX tried to proxy the file. NGINX then proxies the error page from Canvas's file server back to the client.

### Testing

Because NGINX is difficult to debug I've created [a test server](https://gist.github.com/seanh/dd6ea774a8d02e6d8f07f1ea2677cf20) that you can easily run locally and have Via 3's `/proxy/static/` endpoint proxy to, and it will print out exactly what request it received from NGINX (e.g. exactly what path and query string).

### Diagnosis

What's happening is that NGINX is decoding any `%2C`'s in the path part of the URL to `,`'s before it sends its request to the third-party server. Since the URL has been modified Canvas's `token` query param no longer matches it, and authentication to Canvas's file server fails.

### Solution

We need to somehow stop NGINX from decoding things for us...

jon-betts:


seanh (approved):
Let's wait until tomorrow morning to merge and deploy this, but I've tested it and it seems to work

seanh:
Gonna close this as I think we may not be merging it (?), it can always be re-opened if we decide we do want it

seanh (approved):

PR #120, opened 2020-03-17
Move nginx config into nginx directory

jon-betts:
A small change to move the NGINX config together


jon-betts:
We appear to be failing on some setup tools based nonsense here.

```
File "/tmp/pip-install-xrl4nqfz/zope.interface/setup.py", line 26, in <module>
    from setuptools import setup, Extension, Feature
    ImportError: cannot import name 'Feature'
```

seanh (approved):

PR #129, opened 2020-04-07
POC - Use google API to determine mime type

jon-betts:
Google file access currently takes 93% of our running time in route by
content. Getting it via API has a few benefits:

 * It takes about 250ms rather than 1000ms for me locally (4x better)
 * Is polite to Google, and therefore less likely to get banned

To use this POC you will need to set GOOGLE_DRIVE_API_KEY environment
variable which I can pass to you.

### DIFF
--- a/via/views/route_by_content.py
+++ b/via/views/route_by_content.py
@@ -1,5 +1,9 @@
 """View for redirecting based on content type."""
 
+import os
+import re
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


seanh:
I think we can close this?

jon-betts:
This has served it's purpose now. There is a real PR for this in: https://github.com/hypothesis/via3/pull/130

PR #130, opened 2020-04-07
Add support for Google Drive API looking up mime-types

jon-betts:
We're probably not doing this now, but the first half is a good start point for modification as it splits out the get URL function and tests it separately.

This should make it easier to extend and test in isolation from the view that calls it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.egg-info/
 via/static/**/*.gz
 build.tar
+.devdata.env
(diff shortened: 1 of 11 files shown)


seanh:
I think this can be closed for archival? (The PR contents don't go away if you close it)

jon-betts:
We're not using this for now, so I'm sure we can do without. Wasn't terribly much work as I recall even if we have to do it again.

PR #131, opened 2020-04-07
Assume PDF for Google Drive documents

jon-betts:
I'm not sure if this is a feature or a bug, but we are improving the experience for most, at the cost of some:

 * We now assume that all google drive documents are PDFs
 * If we're right this is obviously much faster
 * If we're wrong it will result in an error message in PDF.js

The downside should only effect users who construct URLs for themselves, or use Via 3's root page.

### DIFF
--- a/tests/unit/via/get_url_details_test.py
+++ b/tests/unit/via/get_url_details_test.py
@@ -18,13 +18,27 @@ class TestGetURLDetails:
         "content_type,status_code", (("text/html", 501), ("application/pdf", 200))
     )
     def test_it_calls_get_for_normal_urls(self, requests, content_type, status_code):
-        requests.get.return_value = self._make_response(content_type, status_code)
+        response = Response()
+        response.raw = BytesIO(b"")
+        response.headers = {"Content-Type": content_type}
+        response.status_code = status_code
+        requests.get.return_value = response
+
         url = "http://example.com"
 
         result = get_url_details(url)
 
         assert result == (content_type, status_code)
         requests.get.assert_called_once_with(url, allow_redirects=True, stream=True)
+
+    def test_it_assumes_pdf_with_a_google_drive_url(self, requests):
+        result = get_url_details(
+            "https://drive.google.com/uc?id=--FILEID--&export=download"
+        )
+
+        assert result == ("application/pdf", 200)
+
+        requests.get.assert_not_called()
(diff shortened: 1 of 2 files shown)


tests/unit/via/get_url_details_test.py
```diff
@@ -14,18 +14,34 @@
...
     @pytest.mark.parametrize(
         "content_type,status_code", (("text/html", 501), ("application/pdf", 200))
     )
     def test_it_calls_get_for_normal_urls(self, requests, content_type, status_code):
-        requests.get.return_value = self._make_response(content_type, status_code)
+        response = Response()
+        response.raw = BytesIO(b"")
+        response.headers = {"Content-Type": content_type}
+        response.status_code = status_code
+        requests.get.return_value = response
```
jon-betts, line 25: This was based on a branch which re-used this code. We only use it once now so I've in-lined it.

tests/unit/via/get_url_details_test.py
```diff
@@ -14,18 +14,34 @@
 
 
 class TestGetURLDetails:
+    GOOGLE_API_KEY = "google_api_key"
+
```
seanh:
```suggestion
```

Not used?

seanh (approved):
:+1: Seems pretty simple and should hopefully be a big improvement

robertknight:
Do we have the necessary logging in place that we could measure the number of 403 responses when attempting to proxy PDFs from Google Drive before and after this change and other upcoming related changes?

seanh:
No, we don't have any monitoring of NGINX for any of our apps, and NGINX is what actually proxies the PDFs for Via 3. We have a card to add NGINX monitoring going into the coming sprint: https://github.com/hypothesis/via3/issues/21

tests/unit/via/get_url_details_test.py
jon-betts: Oops. Left overs.

jon-betts:
We do have the monitoring in place to check the changes made here. Currently NewRelic is showing this end-point spending 93% of it's time waiting about for Google Drive, so it should be easy to see the change hopefully.

seanh:
I think you might be able to filter for 403's from Via 3's NGINX's access logs in Papertrail and graph them in Papertrail

PR #132, opened 2020-04-07
Split get_url_details out from the view so it can be tested separately

jon-betts:
We are going to be making changes to this function now, and it's going to gain a bit more complexity. This will keep view and attendant view tests size down.

### DIFF
--- a/tests/unit/via/get_url_details_test.py
+++ b/tests/unit/via/get_url_details_test.py
@@ -22,8 +22,8 @@ class TestGetURLDetails:
         url = "http://example.com"
 
         result = get_url_details(url)
+ 
         assert result == (content_type, status_code)
-
         requests.get.assert_called_once_with(url, allow_redirects=True, stream=True)
 
     @pytest.mark.parametrize("bad_url", ("no-schema", "glub://example.com", "http://"))
--- a/tests/unit/via/get_url_details_test.py
+++ b/tests/unit/via/get_url_details_test.py
@@ -22,7 +22,7 @@ class TestGetURLDetails:
         url = "http://example.com"
 
         result = get_url_details(url)
- 
+
         assert result == (content_type, status_code)
         requests.get.assert_called_once_with(url, allow_redirects=True, stream=True)
 
--- a/tests/unit/via/get_url_details_test.py
+++ b/tests/unit/via/get_url_details_test.py
@@ -1,0 +1,63 @@
+from io import BytesIO
+
+import pytest
+from requests import Response
+from requests.exceptions import (
+    MissingSchema,
+    ProxyError,
+    SSLError,
+    UnrewindableBodyError,
+)
+
+from via.exceptions import BadURL, UnhandledException, UpstreamServiceError
+from via.get_url_details import get_url_details
+
+
+class TestGetURLDetails:
+    @pytest.mark.parametrize(
+        "content_type,status_code", (("text/html", 501), ("application/pdf", 200))
+    )
+    def test_it_calls_get_for_normal_urls(self, requests, content_type, status_code):
+        requests.get.return_value = self._make_response(content_type, status_code)
+        url = "http://example.com"
+
+        result = get_url_details(url)
+        assert result == (content_type, status_code)
+
+        requests.get.assert_called_once_with(url, allow_redirects=True, stream=True)
+
+    @pytest.mark.parametrize("bad_url", ("no-schema", "glub://example.com", "http://"))
+    def test_it_raises_BadURL_for_invalid_urls(self, bad_url):
+        with pytest.raises(BadURL):
+            get_url_details(bad_url)
+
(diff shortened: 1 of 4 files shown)


tests/unit/via/get_url_details_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @staticmethod
+    def _make_response(content_type="application/pdf", status_code=200, body=None):
+        response = Response()
+
+        response.raw = BytesIO(body.encode("utf-8") if body else b"")
+        response.headers = {"Content-Type": content_type}
+        response.status_code = status_code
+
+        return response
```
jon-betts, line 59: This function was introduced as this was snipped out of the full google API PR which made use of it. We actually only use this once now. This is inlined in https://github.com/hypothesis/via3/pull/132

tests/unit/via/views/route_by_content_test.py
```diff
@@ -94,56 +80,31 @@ def test_sets_correct_cache_control(
...
     def call_route_by_content(self, make_request):
-        def call_route_by_content(
-            content_type="text/html", target_url="http://example.com", params=None
-        ):
-            httpretty.register_uri(
-                method=httpretty.GET,
-                uri=urlparse(target_url)._replace(query=None, fragment=None).geturl(),
-                body="DUMMY",
-                adding_headers={"Content-Type": content_type},
-            )
```
jon-betts, line 145: We don't need to mock HTTP calls any more as we now mock `get_url_details` as an external dependency.

tests/unit/via/views/route_by_content_test.py
```diff
@@ -28,41 +17,37 @@ class TestRouteByContent:
         ),
     )
     def test_it_routes_by_content_type(
-        self, content_type, location, call_route_by_content
+        self, content_type, location, call_route_by_content, get_url_details
     ):
-        result = call_route_by_content(content_type)
+        get_url_details.return_value = (content_type, 200)
+        result = call_route_by_content()
```
jon-betts, line 24: Previously we had a more complicated `call_route_by_content` which setup an HTTP pretty call etc. Now we can just mock `get_url_details` as an external dependency.

tests/unit/via/get_url_details_test.py
```diff
@@ -0,0 +1,63 @@
...
+        "content_type,status_code", (("text/html", 501), ("application/pdf", 200))
+    )
+    def test_it_calls_get_for_normal_urls(self, requests, content_type, status_code):
+        requests.get.return_value = self._make_response(content_type, status_code)
+        url = "http://example.com"
+
+        result = get_url_details(url)
+        assert result == (content_type, status_code)
+
+        requests.get.assert_called_once_with(url, allow_redirects=True, stream=True)
```
seanh:
```suggestion
        result = get_url_details(url)
 
        assert result == (content_type, status_code)
        requests.get.assert_called_once_with(url, allow_redirects=True, stream=True)
```

tests/unit/via/views/route_by_content_test.py
seanh:
```suggestion
        get_url_details.return_value = (content_type, 200)

        result = call_route_by_content()
```

tests/unit/via/views/route_by_content_test.py
```diff
@@ -74,9 +59,10 @@ def test_requests_to_legacy_via_are_formatted_correctly(
         "content_type,max_age", [("application/pdf", 300), ("text/html", 60)],
     )
     def test_sets_correct_cache_control(
-        self, content_type, max_age, call_route_by_content
+        self, content_type, max_age, call_route_by_content, get_url_details
     ):
-        result = call_route_by_content(content_type)
+        get_url_details.return_value = (content_type, 200)
+        result = call_route_by_content()
```
seanh:
```suggestion
        get_url_details.return_value = (content_type, 200)
 
        result = call_route_by_content()
```

tests/unit/via/views/route_by_content_test.py
```diff
@@ -94,56 +80,31 @@ def test_sets_correct_cache_control(
...
     ):
-        response = Response()
-        response.status_code = status_code
-        response.raw = mock.Mock()
-
-        requests.get.return_value = response
-
-        result = call_route_by_content("text/html")
+        get_url_details.return_value = (sentinel.content_type, status_code)
+        result = call_route_by_content()
```
seanh:
```suggestion
        get_url_details.return_value = (sentinel.content_type, status_code)

        result = call_route_by_content()
```

tests/unit/via/views/route_by_content_test.py
seanh: (Separating test set up from the test action with a blank line.)

seanh (approved):
:+1:

PR #142, opened 2020-04-22
3A - Remove backwards compatibility for old style LMS params

jon-betts:
This cannot be merged until LMS is using the new style params in: https://github.com/hypothesis/lms/pull/1679


via/configuration.py
```diff
@@ -107,12 +107,3 @@ def _move_legacy_params(via_params, client_params):
 
         # This is like to be around for a while
```
LMS007, line 108: "This is like to be around for a while" - old comment, but should this read "likely" ?

LMS007 (approved):
👍

PR #143, opened 2020-04-23
Add a resource for accessing the URL which raises if missing or bad

### LINKED ISSUE #135: Via 3 fails if "url" parameter is missing on route_by_content view
We currently fail with an un-handled exception if we omit the URL param when routing by content.

We should catch this and raise a sensible error.

jon-betts:
This isn't very fancy error handling, we now raise a Pyramid `HTTPBadRequest` and let the default handling kick in. But it's better than falling on our face and raising errors in monitoring.

We don't actually attempt to make sure the URL is valid, just that it's not empty.
To test run `make dev` then goto:

View PDF:

 * http://localhost:9082/pdf?url=http://pdf995.com/samples/pdf.pdf (200)
 * http://localhost:9082/pdf (400)
 * http://localhost:9082/pdf?ulk=foo (400)
 * http://localhost:9082/pdf?url= (400)

Route by content:

 * http://localhost:9082/route?url=http://pdf995.com/samples/pdf.pdf (302)
 * http://localhost:9082/route (400)
 * http://localhost:9082/route?ulk=foo (400)
 * http://localhost:9082/route?url= (400)

### DIFF
--- a/via/resources.py
+++ b/via/resources.py
@@ -1,0 +1,27 @@
+"""Context objects for views."""
+from pyramid.httpexceptions import HTTPBadRequest
+
+# pylint: disable=too-few-public-methods
+
+
+class URLRoute:
+    """Methods for routes which accept a 'url'."""
+
+    def __init__(self, request):
+        self._request = request
+
+    @property
+    def url(self):
+        """Get the 'url' parameter.
+
+        :raise HTTPBadRequest: If the url is missing or empty
+        """
+        try:
+            pdf_url = self._request.params["url"]
+        except KeyError as err:
+            raise HTTPBadRequest("Required parameter 'url' missing") from err
+
+        if not pdf_url:
+            raise HTTPBadRequest("Required parameter 'url' is blank")
+
+        return pdf_url
--- a/via/resources.py
+++ b/via/resources.py
@@ -4,24 +4,24 @@ from pyramid.httpexceptions import HTTPBadRequest
 # pylint: disable=too-few-public-methods
 
 
-class URLRoute:
+class URLResource:
     """Methods for routes which accept a 'url'."""
 
     def __init__(self, request):
         self._request = request
 
-    @property
     def url(self):
         """Get the 'url' parameter.
 
+        :return: The URL as a string
         :raise HTTPBadRequest: If the url is missing or empty
         """
         try:
-            pdf_url = self._request.params["url"]
+            url = self._request.params["url"]
         except KeyError as err:
             raise HTTPBadRequest("Required parameter 'url' missing") from err
 
-        if not pdf_url:
+        if not url:
             raise HTTPBadRequest("Required parameter 'url' is blank")
 
-        return pdf_url
+        return url
--- a/tests/unit/via/resources_test.py
+++ b/tests/unit/via/resources_test.py
@@ -1,0 +1,20 @@
+import pytest
+from pyramid.httpexceptions import HTTPBadRequest
+
+from via.resources import URLRoute
+
+
+class TestURLRoute:
+    def test_it_returns_url(self, make_request):
+        url = "http://example.com"
+        context = URLRoute(make_request(params={"url": url}))
+
+        assert context.url == url
+
+    @pytest.mark.parametrize("params", ({}, {"urk": "foo"}, {"url": ""}))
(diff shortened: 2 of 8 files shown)


via/resources.py
```diff
@@ -0,0 +1,27 @@
...
+        """
+        try:
+            pdf_url = self._request.params["url"]
+        except KeyError as err:
+            raise HTTPBadRequest("Required parameter 'url' missing") from err
+
+        if not pdf_url:
+            raise HTTPBadRequest("Required parameter 'url' is blank")
+
+        return pdf_url
```
seanh: Is it necessarily a PDF URL? Rename to just `url`?

```suggestion
            url = self._request.params["url"]
        except KeyError as err:
            raise HTTPBadRequest("Required parameter 'url' missing") from err

        if not url:
            raise HTTPBadRequest("Required parameter 'url' is blank")

        return url
```

via/resources.py
```diff
@@ -0,0 +1,27 @@
+"""Context objects for views."""
+from pyramid.httpexceptions import HTTPBadRequest
+
+# pylint: disable=too-few-public-methods
+
+
+class URLRoute:
```
seanh:
```suggestion
class URLResource:
```

These are [called "resources" in Pyramid](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/resources.html) so I think `FooResource` is a better naming style for them. "Routes" are something else entirely

via/resources.py
```diff
@@ -0,0 +1,27 @@
...
+    """Methods for routes which accept a 'url'."""
+
+    def __init__(self, request):
+        self._request = request
+
+    @property
```
seanh: I think it's a little weird for `@property`'s to raise exceptions. See PEP 8:

> Avoid using properties for computationally expensive operations; the attribute notation makes the caller believe that access is (relatively) cheap.

Of course that's not talking about raising exceptions but I think the same reasoning applies: the attribute notation makes it look, from the caller's point of view, like a simple attribute access and so it's surprising that it would raise.

It's a PITA to make a mock property access raise an exception with the `mock` library, if you wanted to add tests for the views like "it does X if context.url raises HTTPBadRequest" (in this PR you didn't write those tests for the views, and I don't think they're necessary in this case either).

When writing the unit tests for a property that raises you'll end up falsely triggering PyLint's `pointless-statement` warning, as you did in your test above, which wouldn't happen if it were a method call.

I think the fact that both mock and PyLint don't expect property accesses to raise and make it awkward to test ones that do, is telling us something.

via/resources.py
```diff
@@ -0,0 +1,27 @@
+"""Context objects for views."""
+from pyramid.httpexceptions import HTTPBadRequest
+
+# pylint: disable=too-few-public-methods
```
seanh, line 4: I wonder whether we should globally disable this one

seanh (changes requested):
This looks good to me.

I suggested a few optional changes none of which seem crucial. Of the suggestions, I think the most important is actually probably to rename `URLRoute` to `URLResource` 'cos it's a resource not a route.

I haven't tested it but I can do so after lunch and after you've made any changes.

### General commentary

A context resource is being used here for what view-guard schemas are used for in LMS.

We need some way of getting the `url` and raising the right kind of error response if it's missing or blank, and since multiple views need to do that the code needs to be in some shared place. In LMS that'd be a shared schema that each view uses in its view_config.

I think the view-guard approach is overall nicer 'cos it means the view never gets called at all with invalid params. But for Via 3 and the very limited amount of validation it actually needs to do, introducing all that would probably be way too heavy.

So this approach seems good to me, and the context resource seems like a fine place to put shared code that multiple views can call.

via/resources.py
jon-betts: Nah. It was where it was copied from (view PDF) but it's not in route_by_content. Fixed.

via/resources.py
jon-betts: Yep.

via/resources.py
jon-betts: I've changed this over to a method.

via/resources.py
jon-betts: I generally think so, but I sort of enjoy being made to think about it. But 100% of the time so far I've decided to ignore it.

seanh (approved):

PR #158, opened 2020-05-04
Show non 500 default NGINX error pages for various upstream errors

### LINKED ISSUE #29: Implement Nginx response status handling
When Nginx proxies a PDF or static file for us (the `/proxy/static/<URL>` endpoint), what should the HTTP statuses of Nginx's responses be?

I'm guessing the current situation is that Nginx proxies the status code from the third-party. So if the URL being proxied returns a 500, Nginx will return a 500, and so on.

This probably isn't what we want because these HTTP statuses will pollute our monitoring. We probably want to 5xx's and other unsuccessful status code for actual problems with our service (e.g. Nginx crashed).

This card is to decide what status codes we want Nginx to respond with, depending on the different status codes that the proxied URL might respond with. And then to implement that.

jon-betts:
Currently this uses `try_files` but we don't actually have any of them. This means we get the NGINX fallback page instead with the right code.

In future we can add nicer pages.

### Testing notes

The mapping is pretty clear from the `nginx.conf` file in the diff. Pick any code from the left and you should end-up with the code listed in the named section.

You can test this by:

 * Running `make services`
 * Open: `http://localhost:9083/proxy/static/http://httpbin.org/status/<STATUS CODE HERE>`
 * **Previously**: You would see the `httpbin.org` page with the relevant status code, or your browser interpretation of the error
 * **Now**: You should see the default NGINX page with the relevant code

### DIFF
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -40,6 +40,13 @@ http {
             proxy_intercept_errors on;
             error_page 301 302 307 = @handle_redirect;
 
+            # Handle most errors upstream listed at:
+            # https://www.restapitutorial.com/httpstatuscodes.html
+            error_page 404 410 = @proxy_not_found;
+            error_page 420 429 = @proxy_too_many_requests;
+            error_page 400 401 402 403 405 406 407 408 409 411 412 413 414 415 416 417 418 422 423 424 425 426 428 431 444 449 450 451 = @proxy_client_error;
+            error_page 500 501 502 503 504 505 506 507 508 509 510 511 598 599 = @proxy_upstream_error;
+
             include via/direct_proxy.conf;
 
             # Cache for a 1 day, but allow serving from cache while revalidating for a week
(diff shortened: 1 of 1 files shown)


conf/nginx/nginx.conf
```diff
@@ -40,6 +40,13 @@ http {
             proxy_intercept_errors on;
             error_page 301 302 307 = @handle_redirect;
 
+            # Handle most errors upstream listed at:
+            # https://www.restapitutorial.com/httpstatuscodes.html
+            error_page 404 410 = @proxy_not_found;
+            error_page 420 429 = @proxy_too_many_requests;
+            error_page 400 401 402 403 405 406 407 408 409 411 412 413 414 415 416 417 418 422 423 424 425 426 428 431 444 449 450 451 = @proxy_client_error;
+            error_page 500 501 502 503 504 505 506 507 508 509 510 511 598 599 = @proxy_upstream_error;
```
jon-betts, line 48: Some attempt has been made to split out different cases we might find interesting. In particular "too many requests" type errors could be very interesting to us, and would be otherwise mixed in with other 400s. 404's also seem like a useful thing to reply to the user.

As far as I can tell, there's no way of using pattern matching here to say `5**`. So we have to list everything we can think of.

conf/nginx/nginx.conf
```diff
@@ -60,6 +67,26 @@ http {
             add_header "X-Via" "static-proxy, redirect";
         }
 
+        location @proxy_not_found {
+            # Not found / gone => 404 not found
+            try_files /proxy/not_found.html =404;
```
jon-betts, line 72: These files don't have to exist yet. NGINX will try them, not find them and fallback to internal defaults.

seanh: Do we want to proxy 404s? I guess why not--these won't be considered as "errors" by monitoring services like Elastic Beanstalk's, they shouldn't cause any false alarms in monitoring or alerting

seanh (approved):

conf/nginx/nginx.conf
jon-betts: I think 404 is one of the few HTTP codes which are understood by lay-people, so it's probably a good idea.

PR #160, opened 2020-05-05
Add a list of URLs and a way for scanning them for validity

### LINKED ISSUE #147: Get a list of URLs for the most commonly used domains in LMS HTML assignments
We need some URLs to use as a test bed for HTML rewriting.

These URLs should:

 * A single link for each domain
 * Ordered by the number of assignments using that **domain** in LMS
 * Be HTML, not documents and PDFs etc.
 * Be publicly accessible (any with authentication should be removed)
 * Ideally number about 50 if possible

The results should be stored in the Via 3 repo to form the basis for a test harness.

It would also be good to have:

 * The prevalence (how many assignments use this domain)
 * The total prevalence (how many assignments are there)

This would allow us to make statements like:

 * "Our solution works for 25/50 of the examples, which covers 13.5% of the assignments in the wild"

jon-betts:
This allows us to:

 * Know which URLs are publicly accessible
 * Return HTML
 * And to cache the contents etc.

The list contains 1024 more popular URLs taken from: https://metabase.hypothes.is/question/861 in a CSV (which is why the diff is so big).

### To test locally

 * `pip install diskcache`
 * Run "bin/scan.py"

This should scan each of the URLs listed in the CSV and build a local disk cache of the results. You can stop at any time and it will resume from where it left off. 

The final output is an updated CSV file which has excluded any URLs which can't be reached (this is the result in the repo right now).


html_test_harness/README.md
```diff
@@ -0,0 +1,8 @@
+# HTML Test Harness
+
+This is a temporary set of utilities to allow testing HTML rewriting.
+
+## Interesting things
+
+ * `data/example_urls.csv` - A spreadsheet of URLs and data
+ * `bin/scan.py` - Will request and update a local cache of data about the URLs
```
jon-betts, line 8: Yeah, this isn't likely to stick around for long, so I don't want to put too much effort in. Also this isn't actually the ticket to produce the harness.

html_test_harness/domain_store.py
```diff
@@ -0,0 +1,85 @@
+import csv
+from pkg_resources import resource_filename
+
+from diskcache import Index
```
jon-betts, line 4: This package is pretty amazing. It transparently pickles things to the disk for you so we can store the actual Response objects coming back from requests and inspect them.

html_test_harness/domain_store.py
```diff
@@ -0,0 +1,85 @@
...
+
+        self._prepopulate()
+
+    def unrequest(self, domain_name):
+        if domain_name not in self.cache:
+            return
+
+        domain = self.cache[domain_name]
+        domain.unrequest()
+        self.cache[domain] = domain
```
jon-betts, line 25: This was needed in testing to scrap our results for a URL if we failed.

seanh (approved):

PR #164, opened 2020-05-15
Enable the stats stub page

jon-betts:
This turns on the stats stub page in NGINX.

To test locally:

 * `docker stop via3_nginx-proxy_1; docker rm via3_nginx-proxy_1; make services;`
 * Go to: http://localhost:9083/_stats
 * You should see something like:

```
Active connections: 1 
server accepts handled requests
 2 2 7 
Reading: 0 Writing: 1 Waiting: 0
```


jon-betts:
Ok, after a chat with @robertknight there are more concerns about the maintainability of the password than there are about people seeing our NGINX stats. So we're just going to scrap the protection entirely. This also gets me closer to finding out if this works. We can come back and harden it once we have a known good version.

Also as we have `/_status` end-points on other products it's probably good to not get confused by giving this a similar name, so we're going to move this to '/_stats'

robertknight (approved):
LGTM

jon-betts:
Ok, this "works" but is quite funny in QA, as the results are cached by Cloudflare, making them totally useless from outside of EB. As we only need it to work inside.

PR #180, opened 2020-06-15
Add `make sure`

### LINKED ISSUE #165: Add `make sure`

seanh:
Fixes https://github.com/hypothesis/via3/issues/165

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ help:
 	@echo "make test              Run the unit tests and produce a coverage report"
 	@echo "make docstrings        View all the docstrings locally as HTML"
 	@echo "make checkdocstrings   Crash if building the docstrings fails"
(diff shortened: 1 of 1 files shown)


jon-betts (changes requested):
This works as intended I think, but currently our build will fail if `make checkdocstrings` fails. So if we intend this to "make sure" that the build will pass it's not currently doing it's job.

The obvious thing to do is to just add "checkdocstrings" to the list of make dependencies, but the right thing to do might be to remove that check from the build if we genuinely don't care.

I'm not 100% what this is intending to do, I think it's just been cargo-culted in from other projects.

jon-betts:
I guess this answers my question: https://github.com/hypothesis/via3/pull/179/files

jon-betts (approved):
Just saw the PR to remove the doc string stuff. Together these make sense.

PR #190, opened 2020-06-26
Run the app through Docker in dev

jon-betts:
This _shouldn't_ have any effect in live at all, but makes local testing much more representative of how we run in live. This has been seen working in Ubuntu and MacOS during the spike.

This means we don't need any of the CORS stuff any more (as everything goes through the same domain).

# Testing

 * Rebuild NGINX
 * `docker stop via3_nginx-proxy_1; docker rm via3_nginx-proxy_1; make services`
 * `make dev`
 * Go to: http://localhost:9083
 * All normal Via 3 stuff should work

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,4 +28,4 @@ COPY . .
 
 USER hypothesis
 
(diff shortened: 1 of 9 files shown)


robertknight (approved):
LGTM when tested on macOS. I haven't fully grokked the `dev_host_bridge.sh` changes but I'm going to approve on the basis that it works for me as a user.

PR #195, opened 2020-06-29
Run NGINX as part of `make dev` not `make services`

### LINKED ISSUE #134: NGINX should be a process not a service, in dev
Currently `make services` runs NGINX whereas `make dev` runs the Python app. While `make services` makes sense as a way to run PostgreSQL, Elasticsearch, and RabbitMQ in h and lms, it doesn't really make sense as a way to run NGINX in via3. In via3, NGINX is really part of the app itself, rather than a separate service that the app depends on. It should therefore be run by `make dev`.

Done when:

- [ ] Via 3 no longer has a `make services` command (or `make services` is a no-op)

- [ ] Running `make dev` in Via 3 runs both NGINX and the Pyramid app in a single shell using a process multiplexer.

  `make services` currently uses Docker Compose to run NGINX. Docker Compose may or may not make sense as the way for `make dev` to run NGINX. It may make more sense for `make dev` to call `docker` directly.

  All of our other apps use Honcho to run multiple processes at once in dev. Honcho is _very_ easy to set up so copying that should be quick and easy. But we're not very happy with how Honcho is performing for us, so we may want to take this as an opportunity to try using Supervisor within a project repo instead.

- [ ] The new way of running things has been updated in `superdev`

- [ ] Bonus points: any changes to the NGINX config file (or any files that NGINX or the NGINX config depends on) triggers an automatic restart of the NGINX process that `make dev` is running.

  This one could be moved to a separate card, allowing this card to be closed without it.

seanh:
Fixes https://github.com/hypothesis/via3/issues/134

NGINX is part of the app itself, not a service. Change `make dev` to run both the Gunicorn and NGINX processes just like `make dev` runs all of the app's processes for our other apps (Gunicorn and assets rebuilding for LMS; Gunicorn, assets and Celery beat and workers for h).

1. Change `make services` to a no-op.

   Via 3 no longer has any services. I did leave the `make services` command in there, it just doesn't do anything. This will make scripting easier if every app has a `make services` (we should add a no-op `make services` to our other apps that don't have one too).
   
   I did remove `make services` from the README though, as it's no longer necessary to run it.

2. Change `make dev` to run Honcho.

   Just the same process multiplexer that `make dev` uses in our other apps. Honcho is extremely simple, does exactly what we want and nothing more.
   
   Honcho is added to the dev requirements and a `Procfile` is added telling Honcho which two processes it should run: Gunicorn and NGINX.

Note that you can still run NGINX and Gunicorn separately if you want to. You would just look in `Procfile` to see what commands it runs, and then run them one by one yourself:

```terminal
$ cat Procfile 
web:   make web
nginx: make nginx
$ make web
[2020-06-29 10:41:37 +0100] [236362] [INFO] Starting gunicorn 20.0.4
[2020-06-29 10:41:37 +0100] [236362] [INFO] Listening at: http://0.0.0.0:9082 (236362)
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,20 +19,28 @@ help:
 
 .PHONY: dev
 dev: python
-	@tox -qe dev
+	@tox -qe dev -- honcho start
 
 .PHONY: devdata
 devdata: python
 	@tox -qe dev -- python bin/devdata.py
+
+.PHONY: web
+web: python
+	@tox -qe dev
+
+.PHONY: nginx
+nginx: args?=up
+nginx: python
+	@tox -qe docker-compose -- $(args)
 
 .PHONY: build
 build: python
 	@tox -qe build
 
 .PHONY: services
-services: args?=up -d
-services: python
-	@tox -qe docker-compose -- $(args)
+services:
+	@true
 
 .PHONY: lint
 lint: python
--- a/README.md
+++ b/README.md
@@ -50,12 +50,9 @@ Create the environment variable settings needed to get Via 3 working nicely with
 The first time you run `make dev` it might take a while to start because it'll
 need to install the application dependencies and build the assets.
 
-This will start the server on port 9082 (http://localhost:9082), reload the
+This will start the server on port 9083 (http://localhost:9083), reload the
 application whenever changes are made to the source code, and restart it should
 it crash for some reason.
-
-To simulate how the app works in production, you should access it through NGINX
-on port 9093 (http://localhost:9083)
 
 **That's it!** You’ve finished setting up your Via 3 development environment. Run
 `make help` to see all the commands that're available for running the tests,
--- a/README.md
+++ b/README.md
@@ -50,9 +50,14 @@ Create the environment variable settings needed to get Via 3 working nicely with
 The first time you run `make dev` it might take a while to start because it'll
 need to install the application dependencies and build the assets.
 
-This will start the server on port 9083 (http://localhost:9083), reload the
-application whenever changes are made to the source code, and restart it should
-it crash for some reason.
+This will start NGINX running on http://localhost:9083 and reverse proxying to
+Gunicorn on http://localhost:9082, reload the application whenever changes are
+made to the source code, and restart it should it crash for some reason.
+
+**You should use NGINX (http://localhost:9083) as your main entry point** to
+Via 3 in development. This is how it's used in production, and if you visit
+Gunicorn directly you'll get CORS (Cross Origin Resource Sharing) errors from
+your browser.
 
 **That's it!** You’ve finished setting up your Via 3 development environment. Run
 `make help` to see all the commands that're available for running the tests,
--- a/README.md
+++ b/README.md
@@ -43,13 +43,6 @@ Create the environment variable settings needed to get Via 3 working nicely with
 
     make devdata
 
-### Run the services with Docker Compose
-
-Start the services that Via 3 requires (currently just NGINX) using Docker
-Compose:
-
-    make services
-
 ### Start the development server
 
     make dev
--- a/README.md
+++ b/README.md
@@ -171,32 +171,8 @@ in development does use the same `nginx.conf` file as the NGINX running in
 Docker in production.
 
 The Python WSGI server (Gunicorn) runs on the host (no Docker) and is exposed
-at http://localhost:9082/.
-
-NGINX is _not_ "in front of" Gunicorn in development. Rather, NGINX and
-Gunicorn are "alongside" each other:
-
-1. The front page of the app in development is http://localhost:9082/, which is
-   served by Gunicorn directly without involving NGINX.
-
-   Other URLs that're handled by the Pyramid app are also at
-   `http://localhost:9082/*` and served by Gunicorn directly.
-
-2. For requests that should be handled by NGINX directly (for example the
-   `/proxy/static/*` URLs) the Pyramid app redirects the browser to
-   `http://localhost:9083/*` URLs that're handled by the NGINX instance running
-   in Docker Compose without further involving Python.
-
-### But that means development is different from production!
-
-That's true but the difference is small. In production requests that should be
-handled by Gunicorn first go to NGINX which then proxies to Gunicorn. Whereas
-in dev these requests go directly from the browser to Gunicorn. A small part of
-the `nginx.conf` file that does the proxying to Gunicorn is used in production
-but not in dev.
-
-In practice this isn't really a problem and putting NGINX in front of Gunicorn
-in development would likely create more issues than it would solve.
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -43,27 +43,17 @@ Create the environment variable settings needed to get Via 3 working nicely with
...
 need to install the application dependencies and build the assets.
 
-This will start the server on port 9082 (http://localhost:9082), reload the
+This will start the server on port 9083 (http://localhost:9083), reload the
 application whenever changes are made to the source code, and restart it should
 it crash for some reason.
 
-To simulate how the app works in production, you should access it through NGINX
-on port 9093 (http://localhost:9083)
-
```
jon-betts, line 66: I think it's right to front-and-center getting to this by NGINX, but we should still describe how the app works. It's not an internal detail and a new developer would need to understand this.

So I think I'd prefer this to say something along the lines of:

```
This will start the web server (http://localhost:9083), which is how
you should access the app underneath (http://localhost:9092) to replicate live. 

The application will reload whenever changes are made to the source code, 
and restart it should it crash for some reason.
```

Or you know... not terrible. 

Also, does this really restart if the app crashes? I thought the
benefit of `honcho` over `supervisor` is it didn't do this? I tried it by just raising in the 
app init, and it doesn't recover.

Makefile
```diff
@@ -19,20 +19,27 @@ help:
...
 devdata: python
 	@tox -qe dev -- python bin/devdata.py
 
+.PHONY: web
+web: python
+	@tox -qe dev
+
+.PHONY: nginx
+nginx: python
+	@tox -qe docker-compose up
```
jon-betts: It would be nice to support some of the same semantics as usual `make service` here. I think not detaching is actually fine, but it would be nice to default the args as we usually do, so they can be overridden for things like `make nginx args=down`

```suggestion
.PHONY: nginx
nginx: args?=up
nginx: python
	@tox -qe docker-compose -- $(args)
```

jon-betts (changes requested):
The actual work here is good, but I have a concern over the README. We appear to be hiding the fact we run two things at all. This seems like detail we want users not to think about, but detail devs need to know.

We want devs to go through NGINX primarily, but not out of ignorance.

Less important is it would be nice if `make nginx` worked the same as `make services` mostly. I.e. you could do `make nginx args=down`. This is less important as our NGINX container for dev isn't the same one we use in production (o_o), and it's more sensitive to change, but it's a small addition which allows much more control for little effort.

README.md
seanh: There are separate sections further down in this README titled _How Via 3 works in production_ and _How Via 3 works in development_, so I think we should let those explain it, and let this _Installing..._ section stick to just giving quick and easy install instructions.

I'll add a commit to update the _How Via 3 works in development_ section which is now out of date.

It doesn't restart the app on crash, no. That's weird, I'm not sure why that's in all the READMEs. We could correct that but on the other hand if we're going to move all the apps to supervisor then maybe it will be restarting them on crash soon enough

seanh: Each one of our apps (h, lms, bouncer, via and via3) has a README that follows the same format. The section headings in the READMEs are:

1. Installing the app in a development environment

   This section is just quick-as-possible docs for telling the user what the commands to get the app going are. No diversions into explaining things about the app or how it works. Very consistent wording and formatting is used across all the READMEs. The subsections are always:

   1. You will need
   2. Clone the git repo (`git clone`)
   3. Run the services (`make services`)
   4. Create the development data and settings (`make devdata`)
   5. Start the development server (`make dev`)
   6. Run the tests (`make test`)

2. Further, app-specific discussion.

   Here, at the bottom of the file, is where each app's README can divert into app-specific discussion and link to further docs.

   For example LMS uses this section to explain SSL issues with LMS's and localhost development. Via has stuff about running with SSL in dev and about updating PDF.js. Bouncer has some random stuff here too. And this is where Via 3 has info about updating PDF.js, and about Via 3's architecture / how it works, including NGINX and Gunicorn, since Via 3 is somewhat unique in this respect.

The frontend apps client and browser-extension differ aggressively from this and from each other, but the five backend apps at least have very consistent READMEs.

Makefile
seanh: Done this

README.md
jon-betts: We can include the link later on in two pages text, and if someone misses it we can accurately say it's in the README. But I think in a larger way we've probably failed.

The following seems likely: A person follows the README step by step. Gets to the `make dev` part. The only URL you see in the output when starting up is `http://localhost:9082`. There's only one URL in the docs which is very similar. "That's the guy" they think and click on it, it mostly works, and then they get a CORS issue. They assume they've done something wrong and go _back_ over the steps they've already done (not reading ahead). End up confused.

A section I'd like to see at the top of our READMEs would look like this:

```
# Quick start

    git clone https://github.com/hypothesis/via3.git
    cd via3
    make devdata
    make dev

You should now have the following running:

 * http://localhost:9083 - Main app entry-point (NGINX docker image `via3_nginx-proxy_1`)
 * http://localhost:9082 - Pyramid app underneath
```

This summarises all of the important info you need to know in one very small section you can copy paste, including hints at the architecture of the app (at least in dev). It also gives you the info you need to check if it's all working. 

This would be particularly helpful in more complicated apps like `h`.

We could then keep the existing sections which detail what `make devdata` is etc. and the far more detailed app explanation. The idea being each section adds detail, rather than much totally new info. If you stop reading sooner (and most do) you know a low-fi version of what you need to know. Think 'advertising' rather than DRY.

seanh: Ah yes, I didn't think about it printing out 9082 to the terminal. Ok, I've added info about NGINX vs Gunicorn directly after the `make dev` command: https://github.com/hypothesis/via3/pull/195/commits/b43141d10681a43d2d43d7d87e9e6af34e529b7c

seanh: Regarding the `# Quick start` idea: I think we _could_ add a quickstart section like that to the top of each app's README. It would have to mention that the prerequisites need to be installed first (otherwise the quickstart commands won't work). It would mean that the sequence of install commands is now duplicated in each app's README. Also, to be honest, although it's not as ultra-concise the **Installing `<APP>` in a development environment** section that each README currently has is intended to be a concise quick-start with a minimum of extraneous explanation and it already fits in roughly one screenful or less, so I think the benefit of adding an even-quicker-start section to the top would be minimal. I think these READMEs are pretty quick and easy to follow as-is.

All the apps have this, [including h](https://h.readthedocs.io/en/latest/developing/install/). We don't have more and less complicated install and setup among our apps -- they're all exactly the same.

jon-betts (approved):
Yep looks good! I think we should think about a quick-start like section at the top. Our documentation is well written for a person coming to if for the first time, but the honest audience is probably us as repeat customers. It would be nice to summarise.

I think for us devs the what it runs and where is probably the part we'd want front and center.

PR #198, opened 2020-07-03
Adding a view to help debugging headers

jon-betts:
This is so we can capture example headers to know where to insert `Referer`.

To test do all the usual stuff then go to:

 * http://0.0.0.0:9082/debug/headers

You should see a page like this:

image


LMS007 (approved):
Deploy to QA!

PR #202, opened 2020-07-07
Add HTML error pages which capture all errors

jon-betts:
* This also includes some specialised details for certain errors
 * This page attempts to explain to the user where the error happened

This doesn't try and help with the NGINX errors yet.

## What it looks like

image

A comparison from LMS (not quite the same, but kind-of similar):
image

## Testing notes

There are a number of details in the error view page which displays all the messages associated with particular errors. These are matched by:

 * Looking for the exact class
 * Any thing that we are a sub-class if the exact can't be found
 * `UnhandledException` for anything else

There's no need to trigger all of the errors to see the messages. However to trigger each:

 * BadURL - http://0.0.0.0:9083/route?url=bad
 * UpstreamServiceError - http://0.0.0.0:9083/route?url=http%3A%2F%2Fvalid-url-but-not-a-website.com
 * HTTPNotFound - http://0.0.0.0:9083/not-a-route
 * HTTPClientError - http://0.0.0.0:9083/route
 * UnhandledException - This is a bit tricker as we handle everything we know about. Easiest way is to raise a random exception in a view


seanh:
I think the "voice" (or the content tone or whatever) of this part with "Oops" and "doesn't look right" etc is quite different from what we normally use at Hypothesis. But I don't think we have a writing style guide so :man_shrugging: And it can be easily changed in future.

> ## Oops, that URL doesn't look right
> 
> Hypothesis is a service that gets a web page and lets you annotate it.
> 
> This URL doesn't look right, so we couldn't try to get the web page.

**I would like to suggest a couple of small changes though**:

1. I don't think it's necessary to say _Hypothesis is a service that gets a web page and lets you annotate it_. I think we can go straight to saying that we couldn't get the page. Which also avoids confusion about what "Hypothesis" is vs what "Via" is.

2. It changes from saying "Hypothesis" in the first sentence to saying "we" in the second. I think it should be "Hypothesis" as it's clearer especially when Hypothesis is running inside other apps like Canvas and it might not always be completely clear who a dialog is coming from, and also because I think that's what we've tended to do in text elsewhere (e.g. in the LMS app).

3. Also say "Hypothesis" instead of "Our service" for the same reasons

4. Front-load the most important part of the message which is that we couldn't get the page

Suggestion:

> ## Oops, that URL doesn't look right
> 
> Hypothesis couldn't get the page because the URL is invalid:
>
>     google.com
>
> x Your request > Hypothesis > The web page

tests/unit/via/views/error_test.py
```diff
@@ -0,0 +1,70 @@
+import pytest
+from _pytest.mark import param
```
seanh: `_pytest`?

tests/unit/via/views/error_test.py
```diff
@@ -0,0 +1,70 @@
...
+from h_matchers import Any
+from mock import sentinel
```
seanh: Oof, [Via 3 depends on the third-party `mock`?](https://github.com/hypothesis/via3/blob/838a70d67a93d81db749ae715c26f38cafd284c7/requirements-test.in#L4) It should be using the standard library's version of mock: `from unittest import mock`, and it _does_ actually do that in other files. That dependency should be removed

via/templates/error.jinja2
```diff
@@ -0,0 +1,211 @@
+{% set LINK = {
```
seanh: The filename should be `error.html.jinja2` (that's the naming convention we use everywhere)

via/views/error.py
```diff
@@ -0,0 +1,96 @@
+"""Error views to handle when things go wrong in the app."""
```
seanh: Should be `exceptions.py` (even though that's not consistently done in other apps yet): https://stackoverflow.com/c/hypothesis/questions/295
Pyramid doesn't have "error views" it has [exception views](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/views.html?highlight=%22exception%20views%22#custom-exception-views) (they can be used for any kind of exception, not just ones that represent errors)

seanh (changes requested):
Some minor queries and suggestions.

Beyond the UI / text changes I suggested above I think we may want to run the wording of all these error messages in general through the #design channel on Slack: the use of things like "Oops", "doesn't look right", "we", "our" and "you", "service", exclamation marks, etc. But I think we should deploy this first, then post a bunch of screenshots in #design and maybe change the wording later if we decide to

tests/unit/via/views/error_test.py
jon-betts: Ah pycharm automatically generated mess.

tests/unit/via/views/error_test.py
jon-betts: I wonder when that happened. I've just been using what's around

via/views/error.py
jon-betts: I'm confused on our usage of the words error and exception here. I think you have a definition in mind and it would be good to see it written down.

When we've discussed errors vs. exceptions before, we've consistently described everything as an error in the class name: "ExampleError", but kept them all in a file called "exceptions.py". From our previous descriptions I couldn't really find anything you'd describe as an exception.

For my experience in Java, which has a fairly strong interpretation there are two very distinct things:

 * Exceptions - Happen in normal flow, and you should catch them
 * Errors - Extremely serious, no point in trying to catch it (CPU has gone away, all memory corrupted etc.)

I can't find any similarly concrete definition for Python. It seems from the basic library that the difference is mostly that "Exception" and "BaseException" are mostly base classes, and all concrete classes are "<Something>Error". So all errors are exceptions, and all custom classes should be errors. 

So as far as I can tell, we shouldn't be declaring "exceptions" as such, rather we should be declaring "errors". So "exceptions.py" is an accurate name to give the file we declare them in, but in the same way that "objects.py" is as well.

I have no objection to renaming the file by the way, but it would be good to understand why, and check that there is a sensible rule at play here.

jon-betts:
With regards to the various suggestions about wording etc:

1. In previous comments you asked who the service was, who we are etc. "Hypothesis is a service that gets a web page and lets you annotate it." is an attempt to set the scene so you know the answer to all these questions. (Although more on the fact we make little effort to identify ourselves later)

We can scoop it out, but does the intended user in LMS have any idea what Hypothesis is or that we exist? If you've been set some homework and it comes back with "Glarbon can't get your website, please contact Glarbon" does it make sense without any more info?

Could we replace this with a "What is Hypothesis?" expand at the bottom or something to give students a fighting chance?

2. The swap to "we" is on purpose in light of the above as a result of having established who "we" are as "Hypothesis". Although if we are going to assume people know what Hypothesis is we can just say it without the additional description. If we remove the above wording the same will still likely apply if we establish "Hypothesis" as the we in the first sentence for subsequent sentences etc. I feel like this is natural reading comprehension ("The dog sat on the ball. It left a bit later.", technically ambiguous, but not really)

I was originally going to say Via (as it's more accurate) but this leads to the problem of Hypothesis/Via relationship.

Another attempt to establish who "we" are as opposed to the LMS was the version with an inclusion of the h logo. You suggested I remove it to be consistent with LMS. This also makes us inconsistent with "h" errors which do include the icon so... not sure you can win that one.

Generally branding is the way you solve this problem e.g. http://google.com/kljnsdfkljnsdfgdf. The first thing this page does is tell you it's Google then refers to "this server". "Which server is that?" is obvious by context. I think the problem here is that LMS doesn't establish who it is, and we are copying that style.

Knowing that we will be embedded I would think we'd make an extra effort to visually distinguish ourselves.

3. :heavy_check_mark:  Similar to the above by saying "Hypothesis is a service" we establish what "the service" is, in this case I think you're right it isn't any shorter, so it might as well swap out there. This could cause problems for people who _do_ know what Hypothesis is and how it works, because this is sort of a lie. Via is a service that gets web pages for you and adds Hypothesis to them.

4. :heavy_check_mark:  I can change the wording on that string to something similar. See below about moving the URL presentation though.

Also, have you looked at the others? All of your feedback here seems "BadURL" specific.

----

With regards to changing the order to have the URL below the first line, it works for "BadURL" and "UpstreamServiceError", but it doesn't make much sense in the context of other errors. At present any URL with a "url" parameter will show this part of the error:

* http://0.0.0.0:9082/no-a-thing?url=a-url-here

image

So unless we want to add another flag which says "URL important" or something it's not clear where to put this to make it work in all cases. (All cases being those listed in `via.error.ERROR_MAP` at time of writing). As above, in the case of the 404, it would be actively misleading to have it after the first sentence.

What do you think? Move it somewhere else, add an extra parameter, something else?

via/views/error.py
seanh: I think it's similar in Python to in Java, but Python does not have the concept that errors should not be caught:

* An exception is just an interruption-to-the-normal-control-flow thing, and not necessarily an error. E.g. `GeneratorExit`, `KeyboardInterrupt`, `StopIteration`, `StopAsyncIteration`, `SystemExit`.

* Errors are a subset of exceptions and represent a "something went wrong", and their name ends in `Error`. E.g. `ImportError`, `AttributeError`, there are many.

Not all concrete exception classes in the stdlib are `Error`.

In Pyramid it's very clear that there's no such thing as an "error view". They're called "exception views", and they don't necessarily represent errors. By default you can raise an HTTP _exception_ (from `pyramid.httpexceptions`) and there are default exception views that will catch these. Some HTTP exceptions are error-ish (unauthorized, not found, ...) (the 4xx's and 5xx's are all subclasses of `HTTPError`, though not all the subclass names themselves end in `Error` but that's just for conciseness and consistency with HTTP naming) but some HTTP exceptions clearly aren't errors (redirects, you can even raise `HTTPCreated` or `HTTPAccepted` etc). HTTP exceptions and exception views are a control-flow thing in Pyramid not just for errors.

Similarly when you write your own custom exception views, the custom exception classes you're catching can be error-like exceptions or not.

This terminology is consistent in Pyramid so we should just follow it.

seanh (approved):
I'll just make a few concrete suggestions, and you can take or leave each one:

1. I think it's weird for an error message to divert into explaining what the app is, before talking about the error. I would just remove _Hypothesis is a service that gets a web page and lets you annotate it_.

   I don't think it's necessary to have a _What is Hypothesis_ explainer or anything on the error page. Other error pages in the LMS app don't do this. As long as it says the name "Hypothesis" somewhere on the error page then the user knows the error message is from "Hypothesis", and we don't worry about whether they understand what "Hypothesis" is or not.

2. I would replace "Our service" with "Hypothesis"

3. I think you can also avoid "this service" in the case of a 404 by changing "The URL you have entered isn't part of this service" -> "The requested URL wasn't found"

So all-in-all the 404 page my way would be:

> # Page not found
> 
> The requested URL wasn't found.
>
> x Your request > :heavy_check_mark: Hypothesis > :heavy_check_mark: The web page

Regarding saying "Hypothesis" once and thereafter switching to "we" and "our": I think users just scan dialogs and read parts of them, don't read the whole thing from top to bottom and comprehend it. So as short and clear and scannable as possible. But :man_shrugging: 

I agree we can avoid saying "Via".

For the URL-below-first-line change: if it's complicated let's not bother, seems not worth it.

Maybe we should consider adding the `[h]` logo to the error and maybe also non-error dialogs in LMS, not sure. It's easier to avoid that can-of-worms for now I think. But for Via 3 consistency with LMS (in which Via 3 runs) is more important than consistency with h's web pages.

Generally speaking what I'm thinking about with the design of these Via 3 error dialogs is the LMS experience. There are lots of error and non-error dialogs in the LMS app that all have the same consistent design. The boundary between the LMS app and Via 3 is unclear and uninteresting to the user, so when a Via 3 error dialog comes up the user is just going to see that as part of the same "Hypothesis" app. It's going to be a jarring and inconsistent experience for the user if some of the dialogs sometimes shown in the LMS experience are very different from all the others. Hence my push for consistency with LMS's design here as the most important thing.

Whether or not LMS's dialog design is ideal is a secondary question. For example maybe the dialogs should have `[h]` logos on them. But in that case we'd have to add those logos to both the LMS and Via 3 dialogs. Because an `[h]` logo that appears on just some dialogs in the LMS experience but disappears from others, is gonna be weird.

jon-betts:
I think you might have confused my explaining why I did it with me saying it's a good idea, so I think we're agreeing furiously on some of this but...

The 404 is tricky. Which URL wasn't found? Unusually there are two in play potentially. Do you think the three panel diagram thing is enough to make the difference there?

via/views/error.py
jon-betts: I'm just trying to work out the definition, I'm happy with the terminology.

So "exceptions" that are not "errors" are really just those that are a part of normal execution flow in completely non error situations. I.e. coming to the end of an iterator is an expected part of how things work when their going right.

So what changes an "exception" into being an "error" is that (annoyingly) something exceptional happened which isn't expected.

To map to the Java concepts:

 * Python exception = for loops, control flow etc. (Java doesn't use exceptions like this I don't think)
 * Python error = Java exception (it's bad, but you should catch it)
 * Python SyntaxError maybe? = Java errors (looks like this concept might be missing from Python mostly)

jon-betts: For the sake of brevity I'm just going to go with this naming scheme, but... If "exceptions" are exclusively normal things to happen, then we'll never show an error page as a result of one right? 100% of what we show is under this definition an "error"? You mentioned "HTTPFound" as examples of non-errors. If our page was broad enough in scope to catch those, that would be a mistake right?

My strong feeling is the definition isn't consistent or useful. "errors" == "exceptions" and we just prefer the term "exception" (style). You can't meaningfully say all errors do this and all exceptions do that, because all errors are exceptions. 

The group of control flow items you mention appear to be missing a concept which groups them and distinguishes them from any other child of Exception.

seanh: The intention of `views/exceptions.py` is that it's where we'd put all that directory's custom exception views, including both error-like and non-error-like exception views should it have any. We've discussed this previously, e.g. https://github.com/hypothesis/lms/issues/1245 and https://stackoverflow.com/c/hypothesis/questions/295.

It's honestly just a misnomer to refer to what the Pyramid docs and code always call "exception views" (`@exception_view_config()`, `add_exception_view()`, ...) as "error views" and means if we ever want to add a non-error-like exception view we'd have to put it elsewhere.

Compare [mentions of "error view" in the Pyramid docs](https://www.google.com/search?q=%22error+view%22+site%3Ahttps%3A%2F%2Fdocs.pylonsproject.org%2Fprojects%2Fpyramid%2F&oq=%22error+view%22+site%3Ahttps%3A%2F%2Fdocs.pylonsproject.org%2Fprojects%2Fpyramid%2F&aqs=chrome..69i57.8490j0j4&sourceid=chrome&ie=UTF-8) to [mentions of "exception view" in the Pyramid docs](https://www.google.com/search?ei=fSgHX8qTLtKd1fAPqae08AE&q=%22exception+view%22+site%3Ahttps%3A%2F%2Fdocs.pylonsproject.org%2Fprojects%2Fpyramid%2F&oq=%22exception+view%22+site%3Ahttps%3A%2F%2Fdocs.pylonsproject.org%2Fprojects%2Fpyramid%2F&gs_lcp=CgZwc3ktYWIQA1DFGVi-ImC7JWgBcAB4AIABS4gB9AOSAQE4mAEAoAEBqgEHZ3dzLXdpeg&sclient=psy-ab&ved=0ahUKEwiKxuPMr8DqAhXSThUIHakTDR4Q4dUDCAw&uact=5)

jon-betts: From what I can see from the hierarchy there is a distinction in a way:

https://docs.python.org/3/library/exceptions.html

* `SystemExit`, `KeyboardInterrupt` and `GeneratorExit` are direct children of `BaseException`
* Everything else is a child of `Exception`, including `StopIteration`
* All warnings are children of `Warning` in addition

Not sure what to make of that. I'm sure there's a reason that `StopIteration` is an odd sausage here, but I can't think of it. It would have been nice if the root was "Raisable" (or something non shit) and then "Exception" was only one branch.

I think the distinction of an "error" vs an "exception" might be one we've invented. The stdlib docs don't mention any distrinction and use the terms interchangeably:

> The built-in **exceptions** listed below can be generated by the interpreter or built-in functions. Except where mentioned, they have an “associated value” indicating the detailed cause of the **error**.

seanh: I think the broader question of whether there are separate concepts of "exception" and "error" in Python generally is a distraction here. To me there's a fairly clear semantic distinction between error and non-error exceptions that can be seen in both the stdlib's exceptions and in Pyramid's HTTP exceptions (where you have an `HTTPException` base class that the 2xx and 3xx exceptions subclass directly, with an `HTTPError` subclass that the 4xx and 5xx exceptions subclass). But putting that aside, they're just called exception views in Pyramid not error views. This seems pretty incontrovertible if you look at the Pyramid API and docs. I don't think there's a good reason for us to decide to call them error views instead? This just seems to be going out of our way to differ from Pyramid's terminology

jon-betts: Anywho, as long as we have our own definition I suppose that's enough.

Maybe something like:

**Style:** We define **errors** (as a subset of **exceptions**) as being those classes raised when something is wrong (e.g. `ValueError`, `requests.exception.InvalidURL`). While all errors are exceptions, some **pure exceptions** are not errors, like `StopIteration` and are a part of normal part of control flow.

**Style:** When defining custom errors and exceptions they are stored in a file called **`exceptions.py`** even if the file happens to contain only errors.

**Style:** Pyramid exception views are stored in a file called **`views/**/exceptions.py`**. Similarly this is true even when Pyramid exception views are used exclusively to catch errors (rather than pure exceptions). This matches the terminology in Pyramid and allows the possibility of expanding the role later to catch and react to pure exceptions (like `HTTPFound`), for example to modify them in some way.

jon-betts: I've added a longer version of this as a S/O question: https://stackoverflow.com/c/hypothesis/questions/295

Also renamed everything I could find to be more exception related.

via/templates/error.jinja2
jon-betts: This is now "exception.html.jinja2"

tests/unit/via/views/error_test.py
jon-betts: This has been fixed up now along with the weird import above.

tests/unit/via/views/error_test.py
jon-betts: Removed now

jon-betts:
I've remove reworded some of the errors and I'll go through some of the decisions:

 * Replaced "Our service" with "Hypothesis"
 * Use the active voice
 * We use "we" and "this service" in the sub-descriptions

----

## About the active vs. passive voice

You suggested a change (a change was suggested) from:

> "The URL you have entered isn't part of this service"

To:

> "The requested URL wasn't found"

This is in an attempt to avoid prompting the question of what "this service" is. This isn't really a fix for a number of reasons:

 * The question remains, who are you?
 * This doesn't provide an answer
 * Requested by whom? Who failed to find it?

A user might not know what "this service" is or who "we" are, but they know who "you" refers to. They gain a lot of knowledge by knowing it wasn't them.

The Civil Service Style guide and many other technical writing guides recommend the use of the the active voice for the following reasons:

 * It makes it clear who and what is acting on what
 * It often leads to shorter clearer sentences: 
> "The committee is considering the proposal" 
> "The proposal is being considered by the committee"

I think we should adopt this consciously. 

We might want to revisit some existing language such as "Something went wrong", which is passive and doesn't indicate who did the mucking up. For example "We had an unexpected error" puts the blame on "us", whoever we might be, rather than the user or other services etc.

jon-betts:
I've added grey circles for parts of the flow that haven't been reached yet as suggested. This is good because it doesn't imply it would work if you get that far, which the previous one did.

tests/unit/via/views/error_test.py
jon-betts: Looks like this was added by me, but I don't know why... anyway gone now.

seanh:
Is this icon supposed to look like this? It's a red circle with a white cross in your screenshot above and I _think_ that's how it was for me locally previously as well, but now I'm seeing red on green, which is a really bad combo for colorblind users:

Screenshot from 2020-07-13 13-30-10

In Firefox the X is also misaligned:

ff

I wonder if we should just use SVG icons here? And the entire circle-with-a-cross-through-it can just be one SVG, rather than trying to draw an SVG on top of a CSS circle background. And similarly circle-with-a-check-through-it could be a single, second SVG image, and grey-circle could be a third. It's probably going to be easier to get consistent colors and alignment that way.

The actual SVG code can even be inlined into the HTML if you don't want to have separate files.

seanh:
There's a minor lack-of-vertical-spacing issue below the Error Details:

Screenshot from 2020-07-13 13-36-53

Compare to LMS:

Screenshot from 2020-07-13 13-37-36

seanh:
There's a minor wording difference:

LMS always says "**If the problem persists** send us an email or open a support ticket".

Whereas these new dialogs say "**To get help from Hypothesis** send us an email or open a support ticket".

It's a minor thing for sure but it is slightly odd for the same sentence to be worded differently in half of the LMS-experience error messages

seanh:
The width of the dialogs is different: these ones are narrower than LMS's ones:

Via 3:

Screenshot from 2020-07-13 13-41-00

LMS:

Screenshot from 2020-07-13 13-40-40

I think these kinds of inconsistencies do add up: if the error dialogs that appear in the same location in the UX, are sometimes one width and sometimes much narrower. Probably worth fixing

seanh:
_It might be missing or have returned an error to us_ reads better as _It might be missing or might have returned an error to us_ IMO

seanh:
_Retrying might help if the web page is having temporary problems_ <- This is true of many, almost all, error dialogs that can happen in either Via 3 or LMS, but we don't say this in any of the other dialogs. We do put a <kbd>Try again</kbd> button on the dialog whenever retrying might work, and I think this is enough. If it were me, I would simply remove this sentence. But if you want to keep it in, can we reword it to _**Trying again**_ _might help if the web page is having temporary problems_ to match with the call-to-action on the <kbd>Try again</kbd> button?

seanh:
I would change _Cannot understand request_ to _Bad request_

seanh:
Ok, I've had another look over these and posted observations as individual comments above.

* I think the cross icon issue is definitely worth fixing
* I think the overall dialog width is definitely worth fixing
* The vertical spacing beneath Error Details is a minor vertical rhythm and consistency issue, but I think worth fixing
* The _To get help from Hypothesis_ -> _If the problem persists_ change is another small thing that I think is worth fixing 'cos it's a consistency issue
* I'll leave the other three up to you, I think they're minor wording improvements

seanh:
Regarding the wider question of error design and passive vs active voice etc: I think there's a wider discussion to be had here. Hypothesis could benefit from a writing style guide for UI text and/or an error messages style guide.

seanh:
Regarding the wider question of error design and passive vs active voice etc: I think there's a wider discussion to be had here. Hypothesis could benefit from a writing style guide for UI text and/or an error messages style guide.

There's a degree of consistency in the existing LMS error and non-error dialogs:

1. They all use the same visual design
2. The title is a "friendly" error message. This can be entirely generic in some cases ("Something went wrong") but where possible a more specific error message is used ("Developer key scopes missing", "Authorize Hypothesis", "Authorization failed", ...) ("Something went wrong" is a fallback for when the error is unrecognised / we don't have a special-case for it yet)
3. Further user-friendly explanation of the error message and help links follow in the body of the dialog (with always the exact same wording for the help links)
4. Where relevant further technical details under Error Details
5. Where relevant a <kbd>Close</kbd> or <kbd>Cancel</kbd> button
6. Where relevant a <kbd>Try again</kbd> button
7. I think the voice / writing style is roughly consistent across them all
8. They're all pretty concise / share the same level of conciseness

I think these new Via 3 dialogs also match those 8 points (other than the tweaks I've just suggested) and I think keeping the experience consistent was the right move for this PR.

But we could have a wider _Can we improve our approach to error messages?_ mini project.

seanh:
One more, I would just delete this:

_Retrying might help, but if you get this error repeatedly please report it._

We already say _If the problem persists send us an email or open a support ticket_ (if we accept my _To get help from Hypothesis_ -> _If the problem persists_ suggestion above) and we show a <kbd>Try again</kbd> button so I think we've got this covered, and the extra sentence is unnecessary.

seanh:
One more thing: it should be vertically-centered as the LMS ones are

jon-betts:
Ok. Attempt one meelion!

 * Changes as we went over in the call
 * Added an SVG sprite for the icons (this does give us one external dependency for the error page... seems ok)
 * Made it vertically center, although this looks pretty goofy when not in a tiny iframe

seanh (approved):
Looks good. I think these error dialogs are pretty great

PR #211, opened 2020-08-03
Add GitHub Actions

seanh:


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,27 @@
...
+      run: python -m pip install --upgrade pip
+    - name: Install tox
+      run: python -m pip install tox
+    - name: Checkout git repo
+      uses: actions/checkout@v2
+    - name: Cache the .tox dir
+      uses: actions/cache@v2
+      with:
+        path: .tox
+        key: ${{ runner.os }}-tox-${{ hashFiles('tox.ini', 'requirements*', 'setup.py', 'setup.cfg') }}
```
robertknight, line 23: I don't know how much it matters, but there is currently no `setup.cfg` file in this repo. It looks like only the h, bouncer and lms projects have that file.

robertknight (approved):
LGTM

.github/workflows/continuous-integration.yml
seanh: Yeah, it doesn't seem to break `hashFiles()` and it just allows the key to be the same across all our repos

PR #212, opened 2020-08-03
Fix relative redirect handling

seanh:
The download URLs for PDFs uploaded to Blackboard return a *relative* redirect (with the leading `https://blackboard.hypothes.is` missing). We apparently haven't encountered relative redirects before because they don't work: Via 3's `/proxy/static` endpoint (the endpoint that is served by NGINX directly) fails to re-add the missing
`https://blackboard.hypothes.is` prefix before redirecting and ends up redirecting the browser to
`https://via3.hypothes.is/proxy/static//bbcswebdav/courses/developer-test-course-1/uploaded-directly.pdf?VxJ...%3D` (note the `//` in `/static//bbcswebdav` where the missing `https://blackboard.hypothes.is` should be).

Fix the static endpoint to insert the missing `https://blackboard.hypothes.is` prefix.

Testing:

To test this with Blackboard Files:

1. Check out this WIP branch of the LMS app: https://github.com/hypothesis/lms/pull/1965

2. Set the `BLACKBOARD_CLIENT_SECRET` envvar: `export BLACKBOARD_CLIENT_SECRET="***"`. See Slack for the secret

2. Log in to https://blackboard.hypothes.is/ using the `administrator` account (it's in 1Password)

3. Go to Developer Test Course 1 and launch Blackboard Files Assignment. Choose Select PDF from Blackboard then click Authorize. **It doesn't actually work**, but the authorization popup window should have caused a temporary Blackboard Files download URL to be printed to the terminal. You can paste the URL into http://localhost:9083/

To test that non-Blackboard files still work:

* An HTTP PDF file: http://localhost:9083/route?url=http%3A%2F%2Fwww.orimi.com%2Fpdf-test.pdf

* An HTTPS PDF file: http://localhost:9083/route?url=https%3A%2F%2Fwww.w3.org%2FWAI%2FER%2Ftests%2Fxhtml%2Ftestfiles%2Fresources%2Fpdf%2Fdummy.pdf

* A PDF file behind a redirect: http://localhost:9083/route?url=https%3A%2F%2Fh.readthedocs.io%2F_%2Fdownloads%2Fen%2Flatest%2Fpdf%2F

Test assignments: https://hypothesis.instructure.com/courses/125/assignments

Known issues:

* This only works for hosts that support https. `https` is hard-coded when inserting the `https://$proxy_host` prefix in `direct_proxy.conf`.  This is because I couldn't find an NGINX variable for the scheme (http or https) of the URL being proxied.

  Having this work for https hosts only is better than it not working at all.

### DIFF
--- a/conf/nginx/via/direct_proxy.conf
+++ b/conf/nginx/via/direct_proxy.conf
@@ -21,7 +21,8 @@ rewrite ^/proxy/static/(.*)$ $1? break;
 # Add back on the args which we stripped above so we don't get them twice
 proxy_pass $uri$is_args$args;
 
-proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/$1;
+proxy_redirect ~^(https?:\/\/)(.*)$ $original_scheme://$http_host/proxy/static/$1$2;
+proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/https://$proxy_host$1;
 
 # Strip hypothesis cookies and authorization header.
 set $stripped_cookie $http_cookie;


conf/nginx/via/direct_proxy.conf
```diff
@@ -21,7 +21,8 @@ rewrite ^/proxy/static/(.*)$ $1? break;
 # Add back on the args which we stripped above so we don't get them twice
 proxy_pass $uri$is_args$args;
 
-proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/$1;
+proxy_redirect ~^(https?:\/\/)(.*)$ $original_scheme://$http_host/proxy/static/$1$2;
+proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/https://$proxy_host$1;
```
seanh, line 25: The first `proxy_redirect` regex matches absolute URLs (`http://` or `https://`) and does the same redirect target as the previous version did.

Only if the first `proxy_redirect` does not match the URL will NGINX then look at the second `proxy_redirect` regex which matches anything (and therefore in practice only matches relative URLs). It does the same redirect target as the first line except it inserts the `https://$proxy_host` bit, which is the `https://blackboard.hypothes.is` that's missing from Blackboard's relative URL.

Note the hard-coded `https://` in the second line: this only works with hosts that support https

jon-betts:
I tried testing this with the stated method, but I couldn't get LMS talking to Blackboard with the provided POC... so begins a fun journey with an annoying conclusion.

I started trying to use http://httpbin.org, which claims to provide relative redirects, but all the redirect end-points are 404'ing at the moment. I then wrote a small Flask app to do this, but Flask _won't let you_ make this mistake and send relative redirects.

So eventually I ended up running http://httpbin.org locally and using that, which does work. To do that:

 * `docker pull kennethreitz/httpbin`
 * `docker run -p 12002:80 kennethreitz/httpbin`
 * Then you can visit: http://localhost:12002/relative-redirect/1
 * Which you can then give to Via with something like this: http://localhost:9083/proxy/static/http://localhost:12002/relative-redirect/1
 * That won't work, but you'll be able to see the incoming IP in the logs and fix it to something like:
 * http://localhost:9083/proxy/static/http://172.29.0.1:12002/relative-redirect/1

This does the redirect + demonstrates the HTTP/S bug with HTTP URLs. So I wondered if there was a way to grab the document scheme from the original URI we have laying around. It turns out this is possible with something like this:

    map $request_uri $doc_scheme {
        ~*^/proxy/static/https:   'https';
        ~*^/proxy/static/http:    'http';
        default     'https';
    }

This has to live in the HTTP block for "reasons" and does seem to set the correct value in the variable`'$doc_scheme`. But for more NGINX "reasons" if you try and interpolate this into the URL, it just doesn't work. The path from the URL seems to disappear. I have no idea why...

So no doubt if you bash your head against this for evermore, it will eventually work, or not... but I can confirm this is working as stated right now, and I don't have any smart ideas for how to fix the HTTP/S issue.

jon-betts (approved):
I couldn't test this as stated, but it does work. I suspect there's a way to fix the HTTP/S issue, but I can't work it out, so I think that's a "cross that bridge" aspect of this.

PR #213, opened 2020-08-03
Don't timeout gunicorn workers in dev

seanh:
In development you often want to drop into a debug shell at some point during the handling of the request. You don't want Gunicorn to timeout the worker after 20s and break your debugging session. Let Gunicorn workers run forever in dev, like we do for the rest of our apps.


jon-betts (approved):
Yep that would be annoying. I think there was a default of 30 applied before, so we're inching towards usable.

If we need to test the behavior of live in dev, we can try and remember to change this locally as necessary.

PR #219, opened 2020-08-10
Remove formatting and linting from Jenkins

seanh:
These are now being run on GitHub Actions so there's no point in running them on Jenkins as well.


jon-betts (changes requested):
Does the parallel need to be there any more? If I'm reading this right, the only part left is `Tests`?

Not quite sure of the syntax

PR #222, opened 2020-08-10
Add a flag to swap out the rewriter URL

jon-betts:
* This falls back to the original Via URL if a new one is missing
 * No test coverage yet

## Testing

 * The behavior can be seen by visiting: http://localhost:9083/route?url=http%3A%2F%2Fwww.google.com&via.rewrite=1
 * This should fail (no rewriter yet), but it should fail by going to: http://localhost:9083/html/v/http://www.google.com&via.rewrite=1
 * Regular routing without the param should go to old Via

## Behavior

 * Any truthy value should trigger the behavior
 * The value is read from "internal_rewriter_url" in `conf/development.ini`
 * It should gracefully fall back to Via if it's missing

## Review notes

 * The logic for making these URLs was getting a bit fat, so to separate it out it was moved to a "service"
 * This is written as a Pyramid service, but isn't hooked up like one
 * The testing of the base functionality was also increased as well as the new

## Other random stuff

 * **There is definitely something up with our coverage reports** - Our report always says 100% coverage, but we definitely don't have it. It's easy to prove by adding some random non called code. I don't know what this is about
 * We should have a sit down and think of a name for all these things as it's becoming a bit of a nightmare

### DIFF
--- a/conf/development.ini
+++ b/conf/development.ini
@@ -3,4 +3,5 @@ use = call:via.app:create_app
 
 nginx_server: http://localhost:9083
 client_embed_url: http://localhost:5000/embed.js
+internal_rewriter_url: http://localhost:9083/html/v
 legacy_via_url: http://localhost:9080
--- a/tests/unit/via/services/html_rewriter_test.py
+++ b/tests/unit/via/services/html_rewriter_test.py
@@ -1,0 +1,77 @@
+from unittest.mock import sentinel
+
+import pytest
+from h_matchers import Any
+
+from via.services.html_rewriter import HTMLRewriter, html_rewriter_factory
+
+
+class TestHTMLRewriterFactory:
+    def test_it(self, pyramid_request):
+        settings = {
+            "legacy_via_url": sentinel.legacy_via_url,
+            "internal_rewriter_url": sentinel.internal_rewriter_url,
+        }
+        pyramid_request.registry.settings.update(settings)
+
+        rewriter = html_rewriter_factory(pyramid_request)
+
+        assert rewriter == Any.instance_of(HTMLRewriter).with_attrs(settings)
+
+
+class TestHTMLRewriter:
+    def test_it_defaults_to_original_via(self):
+        rewriter = HTMLRewriter(legacy_via_url=sentinel.legacy_via_url)
+        assert rewriter.internal_rewriter_url == sentinel.legacy_via_url
+
+    def test_it_extracts_the_url_and_prepends_it(self, rewriter):
+        doc_url = "http://example.com/doc"
+
+        url = rewriter.url_for({"url": doc_url})
+
+        assert url.startswith(f"{rewriter.legacy_via_url}/{doc_url}")
+
+    def test_it_merges_params(self, rewriter):
+        doc_url = "http://example.com/doc?a=1&b=2"
+
+        url = rewriter.url_for({"url": doc_url, "via.openSidebar": "1", "other": "any"})
+
+        assert url == Any.url.with_query(
+            {
+                "a": "1",
+                "b": "2",
+                "via.openSidebar": "1",
+                # Currently we merge any old stuff in, perhaps we shouldn't?
+                "other": "any",
+            }
+        )
+
+    @pytest.mark.parametrize(
+        "enable_rewrite,expected_url",
+        (
+            ("False", "legacy_via_url"),
+            ("", "legacy_via_url"),
+            ("0", "legacy_via_url"),
+            ("True", "internal_rewriter_url"),
+            ("1", "internal_rewriter_url"),
+            ("yes", "internal_rewriter_url"),
+        ),
+    )
+    def test_it_switches_rewriter_based_on_rewrite_option(
+        self, rewriter, enable_rewrite, expected_url
+    ):
+        url = rewriter.url_for({"via.rewrite": enable_rewrite, "url": "anything"})
+
+        assert url.startswith(getattr(rewriter, expected_url))
+
+    @pytest.fixture
+    def rewriter(self):
+        return HTMLRewriter(
+            legacy_via_url="http://example.com/legacy_via_url",
+            internal_rewriter_url="http://example.com/internal_rewriter_url",
+        )
+
+
+@pytest.fixture
+def pyramid_request(make_request):
+    return make_request()
--- a/tests/unit/via/views/route_by_content/_html_rewriter_test.py
+++ b/tests/unit/via/views/route_by_content/_html_rewriter_test.py
@@ -3,23 +3,21 @@ from unittest.mock import sentinel
 import pytest
 from h_matchers import Any
 
-from via.services.html_rewriter import HTMLRewriter, html_rewriter_factory
+from via.views.route_by_content._html_rewriter import HTMLRewriter
 
 
-class TestHTMLRewriterFactory:
-    def test_it(self, pyramid_request):
+class TestHTMLRewriter:
+    def test_from_request(self, pyramid_request):
         settings = {
             "legacy_via_url": sentinel.legacy_via_url,
             "internal_rewriter_url": sentinel.internal_rewriter_url,
         }
         pyramid_request.registry.settings.update(settings)
 
-        rewriter = html_rewriter_factory(pyramid_request)
+        rewriter = HTMLRewriter.from_request(pyramid_request)
 
         assert rewriter == Any.instance_of(HTMLRewriter).with_attrs(settings)
 
-
-class TestHTMLRewriter:
     def test_it_defaults_to_original_via(self):
         rewriter = HTMLRewriter(legacy_via_url=sentinel.legacy_via_url)
         assert rewriter.internal_rewriter_url == sentinel.legacy_via_url
--- a/tests/unit/via/views/route_by_content/_html_rewriter_test.py
+++ b/tests/unit/via/views/route_by_content/_html_rewriter_test.py
@@ -2,6 +2,7 @@ from unittest.mock import sentinel
 
 import pytest
 from h_matchers import Any
+from pyramid.testing import DummyRequest
 
 from via.views.route_by_content._html_rewriter import HTMLRewriter
 
@@ -72,4 +73,4 @@ class TestHTMLRewriter:
 
 @pytest.fixture
 def pyramid_request(make_request):
-    return make_request()
+    return DummyRequest()
--- a/tests/unit/via/views/route_by_content_test.py
+++ b/tests/unit/via/views/route_by_content_test.py
@@ -2,6 +2,7 @@ from unittest.mock import sentinel
 
 import pytest
 from h_matchers import Any
+from pyramid.request import Request
 from webob.headers import EnvironHeaders
 
 from tests.unit.conftest import assert_cache_control
@@ -46,18 +47,15 @@ class TestRouteByContent:
         assert results.location == Any.url.with_query({"other": "value", "url": url})
 
     @pytest.mark.usefixtures("html_response")
-    def test_requests_to_legacy_via_are_formatted_correctly(
-        self, call_route_by_content
+    def test_html_payloads_are_handled_by_the_html_rewriter_service(
+        self, call_route_by_content, html_rewriter, html_rewriter_factory
     ):
-        path = "http://example.com/path%2C"
-        url = path + "?a=b"
+        url = "http://example.com/path%2C?a=b"
         results = call_route_by_content(url, params={"other": "value"})
 
-        # This is horrible, but the params just get blended together. It's
-        # via's job to separate them on the other side
-        assert results.location == Any.url.with_path(path).with_query(
-            {"a": "b", "other": "value"}
-        )
+        html_rewriter_factory.assert_called_once_with(Any.instance_of(Request))
+        html_rewriter.url_for.assert_called_once_with({"other": "value", "url": url})
+        assert results.location == html_rewriter.url_for.return_value
 
     @pytest.mark.parametrize(
         "content_type,max_age", [("application/pdf", 300), ("text/html", 60)],
@@ -117,3 +115,13 @@ class TestRouteByContent:
             return route_by_content(context, request)
 
         return call_route_by_content
+
+    @pytest.fixture
+    def html_rewriter(self, patch):
+        return patch("via.services.html_rewriter.HTMLRewriter")
+
+    @pytest.fixture
+    def html_rewriter_factory(self, patch, html_rewriter):
+        factory = patch("via.views.route_by_content.html_rewriter_factory")
+        factory.return_value = html_rewriter
+        return factory
--- a/via/views/route_by_content/_html_rewriter.py
+++ b/via/views/route_by_content/_html_rewriter.py
@@ -7,20 +7,20 @@ from webob.multidict import MultiDict
 
 from via.configuration import Configuration
 
-# pylint: disable=too-few-public-methods
-
-
-def html_rewriter_factory(request):
-    """Get an instance of HTMLRewriter from a pyramid Request object."""
-
-    return HTMLRewriter(
-        legacy_via_url=request.registry.settings["legacy_via_url"],
-        internal_rewriter_url=request.registry.settings.get("internal_rewriter_url"),
-    )
-
 
 class HTMLRewriter:
     """A service for creating URLs for HTML rewriting engines."""
+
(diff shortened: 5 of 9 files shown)


conf/development.ini
```diff
@@ -3,4 +3,5 @@ use = call:via.app:create_app
 
 nginx_server: http://localhost:9083
 client_embed_url: http://localhost:5000/embed.js
+internal_rewriter_url: http://localhost:9083/html/v
```
seanh, line 6:
```suggestion
html_rewriter_url: http://localhost:9083/html/v
```

I think we should avoid "internal". Via 3 doesn't care whether the rewriter is part of Via 3 or not, it's just a URL that it redirects to. Also, I think the plan is for the new HTML rewriter to be external

tests/unit/via/services/html_rewriter_test.py
```diff
@@ -0,0 +1,77 @@
...
+    def rewriter(self):
+        return HTMLRewriter(
+            legacy_via_url="http://example.com/legacy_via_url",
+            internal_rewriter_url="http://example.com/internal_rewriter_url",
+        )
+
+
+@pytest.fixture
+def pyramid_request(make_request):
+    return make_request()
```
seanh: Does Via 3 not have a global `pyramid_request` fixture? It should. It should be a `DummyRequest` too, `make_request()` returns a real `Request`

tests/unit/via/views/route_by_content_test.py
```diff
@@ -117,3 +115,13 @@ def call_route_by_content(target_url="http://example.com", params=None):
...
+
+    @pytest.fixture
+    def html_rewriter(self, patch):
+        return patch("via.services.html_rewriter.HTMLRewriter")
+
+    @pytest.fixture
+    def html_rewriter_factory(self, patch, html_rewriter):
+        factory = patch("via.views.route_by_content.html_rewriter_factory")
+        factory.return_value = html_rewriter
+        return factory
```
seanh: It's unusual for `route_by_content_test.py` to be patching `services/html_rewriter.py.HTMLRewriter`: these are the unit tests for `views/route_by_content.py`, they shouldn't be calling `services/html_rewriter.py` at all so why are they patching one of its names?

In fact you don't need to patch it, you just need to set `views.route_by_content.html_rewriter_factory()`'s return value to a mock `HTMLRewriter`. We'd normally do that using `create_autospec()` rather than patch:

```diff
diff --git a/tests/unit/via/views/route_by_content_test.py b/tests/unit/via/views/route_by_content_test.py
index 780bd61..6bc2e1e 100644
--- a/tests/unit/via/views/route_by_content_test.py
+++ b/tests/unit/via/views/route_by_content_test.py
@@ -1,4 +1,4 @@
-from unittest.mock import sentinel
+from unittest.mock import create_autospec, sentinel
 
 import pytest
 from h_matchers import Any
@@ -7,6 +7,7 @@ from webob.headers import EnvironHeaders
 
 from tests.unit.conftest import assert_cache_control
 from via.resources import URLResource
+from via.services import HTMLRewriter
 from via.views.route_by_content import route_by_content
 
 
@@ -117,11 +118,12 @@ class TestRouteByContent:
         return call_route_by_content
 
     @pytest.fixture
-    def html_rewriter(self, patch):
-        return patch("via.services.html_rewriter.HTMLRewriter")
+    def html_rewriter(self, html_rewriter_factory):
+        return html_rewriter_factory.return_value
 
     @pytest.fixture
-    def html_rewriter_factory(self, patch, html_rewriter):
-        factory = patch("via.views.route_by_content.html_rewriter_factory")
-        factory.return_value = html_rewriter
-        return factory
+    def html_rewriter_factory(self, patch):
+        return patch(
+            "via.views.route_by_content.html_rewriter_factory",
+            return_value=create_autospec(HTMLRewriter),
+        )
diff --git a/via/services/__init__.py b/via/services/__init__.py
index e69de29..2dd4a49 100644
--- a/via/services/__init__.py
+++ b/via/services/__init__.py
@@ -0,0 +1 @@
+from via.services.html_rewriter import HTMLRewriter
```

seanh: Alternatively, you could put in place what Via 3 should really have which is centralised mocks of the services like LMS has:

https://github.com/hypothesis/lms/blob/5827db48c6cb9f5e40f2a2ebc6c4339257cce2fa/tests/unit/conftest.py#L156-L245

(Note that they're all registered via `pyramid_config.register_service()`.)

You then just put a `pytestmark` at the bottom of your test file to declare which fixtures you need:

```
pytestmark = pytest.mark.usefixtures("ai_getter", "h_api", "group_info_service")
```

seanh: This would require the `HTMLRewriter` service to be a `pyramid_services` service. Currently it appears to imitate `pyramid_services`, separate factory function and all, but doesn't actually use `pyramid_services`? The ability to register alternate versions of services (e.g. mock ones in tests) is part of the intent of `pyramid_services`'s design. The authors of Pyramid and `pyramid_services` (same people) very likely new what they were doing better than we do, so I think we should probably just use it. It will mean the same solutions to problems (such as centralised mocking of services in tests) that we've used in LMS can also be used in Via 3, we won't have to re-solve uninteresting problems differently.

seanh (changes requested):

conf/development.ini
jon-betts: It is specifically both:

 * Not the default
 * Not Via (which is also an HTML rewriter)

So I don't know what the right name is here, but I don't think this is it yet: `experimental_rewriter_url`?

tests/unit/via/services/html_rewriter_test.py
jon-betts: I've changed this to just use `DummyRequest` directly. We don't make use of it anywhere else, so I don't think making it global is useful.

I've not changed the base `make_request` function into a `DummyRequest`.

tests/unit/via/views/route_by_content_test.py
jon-betts: These have been moved into a close collaborator with the view now. For context:

* I used patch because it's around and convenient. The inner object could have been an autospec (the distinction is gone now).
* You can patch target or source location, they both work fine, I just happened to pick source here
* I didn't create a centralised mock as it's only used in one place
* Patching the source allows you to create a central mocks used by multiple files (as the patch happens pre-import)

You don't need something to be a `pyramid_services` style service to have a central re-usable mock of it. I think this is confusing the problem for the solution:

 * _When_ you have `pyramid_services` you have to use `register_services` to mock them correctly
 * We don't use `pyramid_services` in order to be able to mock using `register_services`

Which is to say, using `pyramid_services` introduces a mocking problem that we have a standard solution to, but without it, the even more standard solution of patching works fine.

jon-betts:
This has been moved about to have a view with close collaborator now.

tests/unit/via/views/route_by_content_test.py
seanh: I think you've misunderstood this. Patching at source doesn't work. The patch doesn't happen pre-import. See [Where to patch](https://docs.python.org/3/library/unittest.mock.html#where-to-patch) in the patch docs ("you must ensure that you patch the name used by the system under test ... the basic principle is that you patch where an object is looked up, which is not necessarily the same place as where it is defined").

### Why your previous source patch appeared to work

Previously you had:

```python
# route_by_content_test.py

    @pytest.fixture
    def html_rewriter(self, patch):
        # This was a patch at source:
        return patch("via.services.html_rewriter.HTMLRewriter")

    @pytest.fixture
    def html_rewriter_factory(self, patch, html_rewriter):
        # This was a patch at target:
        factory = patch("via.views.route_by_content.html_rewriter_factory")
        factory.return_value = html_rewriter
        return factory
```

The `html_rewriter()` fixture actually **doesn't** get applied pre-import. If `route_by_content.py` (as it was called previously) had a `from via.services.html_rewriter import HTMLRewriter` the `html_rewriter()` fixture **wouldn't** have replaced `via.views.route_by_content.py::HTMLRewriter` with a mock.

The reason why this appeared to work was that `route_by_content.py` **didn't** import `HTMLRewriter`. It only imported `html_rewriter_factory()` (which your `html_rewriter_factory()` fixture above was correctly patching at _target_). And the mock `HTMLRewriter` was set as the mock `html_rewriter_factory`'s `return_value`.

So:

1. The source patch `patch("via.services.html_rewriter.HTMLRewriter")` was being applied too late / it wasn't working because it was a source patch
2. The target patch `patch("via.views.route_by_content.html_rewriter_factory")` _was_ working correctly
3. The mock `HTMLRewriter` from the source patch was set as the `html_rewriter_factory()` target patch's return value and, since the `html_rewriter_factory()` patch _was_ correct, the code-under-test was getting the mock `HTMLRewriter` via the mock `html_rewriter_factory`'s `return_value`. It was _not_ getting it via the `HTMLRewriter` source patch, which was having no effect

So your `html_rewriter` fixture wasn't actually being used as a patch. The patch was failing, applying too late. But the fixture _was_ being used to create the `return_value` for the `html_rewriter_factory` fixture. That's why I suggested that `patch()` wasn't appropriate here--it was misleading because the test wasn't actually making use of the patch. It just needed an object. Hence `create_autospec()`, which is what we always use for this sort of thing (return values), was more appropriate, and it's actually simpler as well:

```python
    @pytest.fixture
    def html_rewriter(self):
        return create_autospec(HTMLRewriter, instance=True, spec_set=True)

    @pytest.fixture
    def html_rewriter_factory(self, html_rewriter, patch):
        return patch(
            "via.views.route_by_content.html_rewriter_factory",
            return_value=html_rewriter,
        )
```

### Another issue with the source-patch fixture

Just FYI for the future `patch("via.services.html_rewriter.HTMLRewriter")` returns a mock **of the `HTMLRewriter` class** not a mock `HTMLRewriter` object-instance. But an object-instance is what you actually want. There are all sorts of differences between mocks-of-classes and mocks-of-instances, for example mocks-of-classes are callable (and return a mock instance of the class!). Your tests were getting away with this because they happened not to trip on any of the differences but it was incorrect and could have led to problems (either test failures, or tests passing that should fail). The `instance=True` argument to `create_autospec(HTMLRewriter)` makes it return a mock instance not a class.

This is also the reason for the confusion in the fixture naming: the fixture is named `html_rewriter` but it patches `HTMLRewriter`. Because it's using a `patch()` to do a `create_autospec()`'s job.

### Why source patching doesn't work

For example this change would break the `view_test.py` in this PR because `HTMLRewriter` would no longer be successfully mocked in the tests:

```diff
diff --git a/tests/unit/via/views/route_by_content/view_test.py b/tests/unit/via/views/route_by_content/view_test.py
index 5cbbfb3..60b657c 100644
--- a/tests/unit/via/views/route_by_content/view_test.py
+++ b/tests/unit/via/views/route_by_content/view_test.py
@@ -119,4 +119,4 @@ class TestRouteByContent:
 
     @pytest.fixture
     def HTMLRewriter(self, patch):
-        return patch("via.views.route_by_content.view.HTMLRewriter")
+        return patch("via.views.route_by_content._html_rewriter.HTMLRewriter")
```

The order of operations is:

1.`view_test.py` does `from via.views.route_by_content.view import route_by_content`
2. `view.py` does `from via.views.route_by_content._html_rewriter import HTMLRewriter`
3. The `_html_rewriter.py` module is imported for the first time. Names in its namespace, like `HTMLRewriter` are defined as real objects
4. The name `view.HTMLRewriter` in the `view.py` module is created, and points to the real `HTMLRewriter` class
5. The patch fixture gets executed. The name `_html_rewriter.HTMLRewriter` in the `_html_rewriter.py` module's namespace gets replaced with a mock. But `HTMLRewriter` already has a second name in another namespace: `view.HTMLRewriter` has already been imported. It does not get replaced with a mock

You have to patch in the module under test (`"via.views.route_by_content.view.HTMLRewriter"`) not in the module that it imports from (`"via.views.route_by_content._html_rewriter.HTMLRewriter"`).

As a result, you will need a _different_ `HTMLRewriter` patch for every module that imports `HTMLRewriter` because you're patching a different name in a different namespace each time.

This is what `pyramid_services`'s registry allows us to avoid, enabling us to create centralized fixtures for all the services.

## Why Sean begs you not to patch things at source

Please no patch things at source because it can create a wide range of confusing and time consuming problems. I have seen unspeakable things, I tell you.

To give you a flavor: success depends on the timing relationship between import and patch. You can get situations where the patch will apply in time or not depending on what order the tests are run in, or on what subset of tests are run. It can depend on what styles of import statements are used (possibly involving the import statement styles across multiple test and code modules). With tests that are not fully isolated to a single module (more than one module is being run by the tests) you could end up with a `Foo` class patched for the `bar` module but _not_ patched for the `gar` module that's also run by the same test.

A test that was once working fine, can end up running with an un-patched object due to a seemingly unrelated change elsewhere (and this can either cause the test to fail, or to pass when it should now be failing, depending). For example adding an import to a `conftest.py` file or to another module in the production code, can change the import-patch ordering relationships.

I've spent a lot of time debugging these sorts of problems. That period of my life is over now that we patch at target!

Always patching targets not sources, is easy to understand and simple to apply and it avoids the entire class of problems.

conf/development.ini
seanh: I think the problem may be that we simply don't have a name for the new HTML rewriter. I guess we just need to come up with one. I feel like even just a random word would do. Or some hilarious play like viable or viaduct or vial (!) etc. The German word for via is uber. In Esperanto it's vojo. Etc.

via/views/route_by_content/_html_rewriter.py
```diff
@@ -0,0 +1,75 @@
+"""Logic for interacting with HTML rewriters."""
+
+from urllib.parse import parse_qsl, urlencode, urlparse
+
+from pyramid.settings import asbool
+from webob.multidict import MultiDict
+
+from via.configuration import Configuration
```
seanh, line 8: I think something's up with `configuration.py`, which `_html_rewriter.py`'s unit tests don't patch. It's also used by `view_pdf.py` and `view_pdf.py`'s unit tests _do_ patch _except_ for one of its unit tests which depends on the real version. What is `Configuration`? Is it a close collaborator of the pdf and route_by_content views?

Its job seems to be to extract things from `via.*` query params so I think it either belongs in views as a collaborator or it should actually be a validation schema and separately unit-tested (it looks like the kind of thing that'd be a Marshmallow schema in LMS)

tests/unit/via/views/route_by_content/_html_rewriter_test.py
```diff
@@ -0,0 +1,76 @@
...
+    def rewriter(self):
+        return HTMLRewriter(
+            legacy_via_url="http://example.com/legacy_via_url",
+            internal_rewriter_url="http://example.com/internal_rewriter_url",
+        )
+
+
+@pytest.fixture
+def pyramid_request(make_request):
+    return DummyRequest()
```
seanh, line 76: At some point we should make this a global fixture. All of our apps have one. If we had a Pyramid app cookiecutter, it would be part of that. It's just convenient to always be able to assume that the `pyramid_request` fixture is at your fingertips

seanh (approved):

via/views/route_by_content/_html_rewriter.py
jon-betts: I wouldn't sweat it right now, as that's candidate #1 to be moved out to a separate library. Given that and the side-effectless behavior, it's mostly a patch when convenient kind of thing.

tests/unit/via/views/route_by_content/_html_rewriter_test.py
jon-betts: It would take a bit of gymnastics to actually make it be the one in play when you call `make_request()` for example, rather than just a stand-alone shortcut for `DummyRequest()`.

seanh: I think that's actually pretty simple. Something like:

```python
# conftest.py

def _make_request():
    return .... # (DummyRequest() or whatever)

@pytest.fixture
def make_request():
    return _make_request

@pytest.fixture
def pyramid_request():
    return _make_request()
```

So now you have a `pyramid_request` fixture if you just want the one global request object, and a `make_request()` fixture that's a function that'll return you a new object every time, for the rare cases when tests need more than one request.

The real-life versions of the above fixtures would be more because they'd do some customisation of the request object. But the gymnastics would be the same

PR #255, opened 2020-09-01
Using h-vialib for config parsing, instead of duplicating it

jon-betts:
Also fixed a random boo boo in the setup.py while I was here.


setup.py
```diff
@@ -11,7 +11,7 @@
     version='3.0',
     description='Hypothesis Proxy Application',
     long_description=README,
-    url='https://github.com/hypothesis/lms',
+    url='https://github.com/hypothesis/via3',
```
jon-betts, line 14: Copy paste monster strikes again...

seanh (approved):

PR #260, opened 2020-09-07
Pin all Python dependencies

seanh:
Move all Python dependencies for the dev, test, lint, format, etc
environments out of the `tox.ini` file and into separate
`requirements/*.in` files that're compiled into `requirements/*.txt`
files by `pip-compile`.

This is so that all of our tox environments will always contain the
exact same versions of the exact same packages on every developer's
machine and on CI. When a new version of some package gets released
(e.g. PyLint) this will not cause our dev and CI envs to change and
potentially break. Instead, Dependabot will send PRs to upgrade the
necessary `requirements/*.txt` files and, if there's any breakage, we
will see that on the Dependabot PR and the breakage won't get onto
master.

This fixes an issue that new versions of PyLint, pytest, isort, etc etc
are often released and cause CI or dev environment commands to break.

This commit also removes the separate `pip-compile` virtualenv. The
pip-tools docs say:

> `pip-compile` should be run from the same virtual environment as your project so conditional
> dependencies that require a specific Python version, or other environment markers, resolve
> relative to your project's environment.

So get tox to add `pip-tools` to every env after installing the env's
requirements file.

This means that the commands to manage Python dependencies are now:

To add a new dependency or to remove a dependency add the package name
to or remove the package name from the appropriate
`requirements/<ENV>.in` then run:

    tox -e <ENV> --run-command 'pip-compile requirements/<ENV>.in'

This will modify the corresponding `requirements/<ENV>.txt` file. Commit
these changes to git and send a pull request.

To upgrade a single package to the latest version run:

    tox -e <ENV> --run-command 'pip-compile --upgrade-package <PACKAGE> requirements/<ENV>.in'

To upgrade or downgrade a single package to a specific version run:

    tox -e <ENV> --run-command 'pip-compile --upgrade-package "<PACKAGE>==<VERSION>" requirements/<ENV>.in'


tox.ini
```diff
@@ -36,25 +36,17 @@ passenv =
...
+    lint: -r requirements/lint.txt
+    {format,checkformatting}: -r requirements/format.txt
+    updatepdfjs: -r requirements/updatepdfjs.txt
+    build: -r requirements/build.txt
+    pip-tools<5.0.0
 whitelist_externals =
     dev: gunicorn
     dev: newrelic-admin
-    update-pdfjs: sh
+    updatepdfjs: sh
```
seanh, line 49: Removed the `-` from this because `-`'s are meaningful in tox env names

requirements/build.txt
```diff
@@ -0,0 +1,11 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile requirements/build.in
+#
+-e file:///home/seanh/Projects/via3  # via -r requirements/build.in
```
jon-betts: Nope!

jon-betts (changes requested):
This has picked up some Sean H specific local changes.

requirements/build.txt
seanh: Fixed using the workaround here: https://github.com/jazzband/pip-tools/issues/204#issuecomment-550051424

jon-betts (approved):
Yep this is all working fine locally again. I've also tried regenerating the dev requirements and they turn out the same.

PR #262, opened 2020-09-08
Remove `-e file:.#egg=via` from requirements

seanh:
Putting `-e .` in requirements files doesn't work, see:

https://github.com/hypothesis/lms/pull/2111
https://github.com/jazzband/pip-tools/issues/204#issuecomment-550051424

The workaround of putting `-e file:.#egg=via` in the requirements files
instead doesn't work with Dependabot:

https://github.com/hypothesis/via3/issues/261
https://github.com/hypothesis/via3/pull/257#issuecomment-688825972

The `-e .` / `-e file:.#egg=via` appears to be unnecessary anyway: most
of the virtualenvs across most of our apps don't use, only _some_ of Via
3's and some of LMS's venvs use it, for unknown reasons.

Fix the issue by removing the `-e .` / `-e file:.#egg=via`. As far as I
can tell it's unnecessary, everything still works fine. Use of
`pkg_resources.resource_filename()` in `create_pdf_template.py` needed
to be fixed, but that appears to be all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Updating the PDF viewer
 
 Via 3 serves PDFs using [PDF.js](https://mozilla.github.io/pdf.js/). PDF.js is
 vendored into the source tree and the viewer HTML is patched to load the Hypothesis
-client. To update the PDF viewer, run `tools/update-pdfjs`.
+client. To update the PDF viewer, run `make update-pdfjs`.
 
 How Via 3 works
 ---------------
--- a/requirements/build.in
+++ b/requirements/build.in
@@ -1,4 +1,3 @@
--e file:.#egg=via
 whitenoise
 rcssmin
 rjsmin
--- a/requirements/build.txt
+++ b/requirements/build.txt
@@ -4,7 +4,6 @@
 #
 #    pip-compile requirements/build.in
 #
--e file:.#egg=via         # via -r requirements/build.in
 htmlmin==0.1.12           # via -r requirements/build.in
 rcssmin==1.0.6            # via -r requirements/build.in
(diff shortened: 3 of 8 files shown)


bin/create_pdf_template.py
```diff
@@ -1,9 +1,9 @@
 """Create the PDFjs template."""
 
-from pkg_resources import resource_filename
+import os
 
-SOURCE_FILE = resource_filename("via", "static/vendor/pdfjs-2/web/viewer.html")
-TARGET_FILE = resource_filename("via", "templates/pdf_wrapper.html.jinja2")
+SOURCE_FILE = os.path.abspath("via/static/vendor/pdfjs-2/web/viewer.html")
+TARGET_FILE = os.path.abspath("via/templates/pdf_wrapper.html.jinja2")
```
jon-betts: This is kind of a bad sign. Using `pkg_resources` is the right way to get at things inside a package in a `pwd` agnostic way. Not being able to call any code from the package in question in the product suggests our paths just aren't setup correctly.

A different way to fix this is to add `PYTHONPATH = .` to the `setenv` section of `tox.ini`

jon-betts (changes requested):
The main change seems fine, but removing `pgk_resources` in favour of fixed paths suggests fixing the wrong problem to me.

I think the real issue is our PYTHONPATH just isn't configured correctly. This can be fixed in `tox.ini` and the script works as expected. This means we can call our code in scripts, which would be kind of disabling if we couldn't.

bin/create_pdf_template.py
seanh: Done

jon-betts (approved):
Lets go!

PR #297, opened 2020-11-03
Don't emit caching headers on HTML redirects if we switching to ViaHTML

jon-betts:
If the ratio is anything other than 0 or 1, then don't allow caching.

For: https://github.com/hypothesis/via3/issues/236

This isn't very nice, and kind of messes up the design, but we should remove all of this once the cross over is complete.

## Testing notes

* `VIA_HTML_RATIO=0.5 make dev`
* `curl -I http://localhost:9083/route?url=http%3A%2F%2Fexample.com`
* Run the curl command a few times
* You should see `Cache-Control: no-cache`
* The URL should flip between the two

Run either:
* `VIA_HTML_RATIO=0.0 make dev` or...
* `VIA_HTML_RATIO=1.0 make dev`
* `curl -I http://localhost:9083/route?url=http%3A%2F%2Fexample.com`
* You should see `Cache-Control: public, max-age=60, stale-while-revalidate=86400`
* The URL should be one or the other consistently

### DIFF
--- a/via/views/route_by_content/view.py
+++ b/via/views/route_by_content/view.py
@@ -21,8 +21,12 @@ def route_by_content(context, request):
 
         return exc.HTTPFound(redirect_url, headers=_caching_headers(max_age=300))
 
-    url = HTMLRewriter.from_request(request).url_for(request.params)
-    headers = _cache_headers_for_http(status_code)
+    url, cacheable = HTMLRewriter.from_request(request).url_for(request.params)
+    headers = (
+        _cache_headers_for_http(status_code)
+        if cacheable
+        else {"Cache-Control": "no-cache"}
+    )
 
     return exc.HTTPFound(url, headers=headers)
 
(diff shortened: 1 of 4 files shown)


via/views/route_by_content/view.py
```diff
@@ -21,8 +21,12 @@ def route_by_content(context, request):
 
         return exc.HTTPFound(redirect_url, headers=_caching_headers(max_age=300))
 
-    url = HTMLRewriter.from_request(request).url_for(request.params)
-    headers = _cache_headers_for_http(status_code)
+    url, cacheable = HTMLRewriter.from_request(request).url_for(request.params)
+    headers = (
+        _cache_headers_for_http(status_code)
+        if cacheable
+        else {"Cache-Control": "no-cache"}
```
jon-betts, line 28: There isn't actually any test coverage of this behavior. I want to remove it soon, so I thought I'd be lazy

seanh (approved):

PR #313, opened 2021-01-20
Remove the front page

seanh:
We're not using the front page currently (the LMS app doesn't use it). When we do want the front page back (before Via 3 can replace legacy Via) we can just revert this PR. I'll add an issue to the replace-legacy-Via milestone noting this.


via/views/exceptions.py
```diff
@@ -69,7 +70,13 @@ def _get_meta(exception):
 
 
 @exception_view_config(Exception, renderer="via:templates/exception.html.jinja2")
-@exception_view_config(HTTPError, renderer="via:templates/exception.html.jinja2")
+@exception_view_config(
+    HTTPError,
+    renderer="via:templates/exception.html.jinja2",
+    # Match any page *except* the front page ("/").
+    # We don't want our fancy error page to be used on the front page.
+    path_info=not_(r"^\/$"),
```
seanh, line 78: By disabling the exception view for the front page I've turned it into the standard hard-coded 404 response (which comes from either Pyramid or Gunicorn, can't remember) rather than Via 3's very friendly and helpful error page. This is because the error page kind of gives away more obviously that there's an app running and the error dialog even explains that this app is a proxy :)

jon-betts (approved):
This is sort of a wee shame for us as it's a bit of a pain to test it. I'm sure we can make a private test harness to help or something.

This works as advertised gets rid of the landing page and ensures we don't waste any bandwidth replying to the curious by returning a very bland page.

I wonder if there's an even more aggressive version of this where we just drop the connection instead of returning anything at all?

There's also `410 Gone`:

> The requested resource is no longer available at the server and no forwarding address is known. This condition is expected to be considered permanent. Clients with link editing capabilities SHOULD delete references to the Request-URI after user approval.

PR #317, opened 2021-01-25
Tell Google not to index or follow Via 3's pages

seanh:
Add `X-Robots-Tag: noindex, nofollow` headers to all Via 3's responses. This tells Google not to index `/proxy/static/` pages and not to follow links on those pages. See:
    
* https://github.com/hypothesis/checkmate/issues/167
* https://developers.google.com/search/docs/advanced/crawling/block-indexing#http-response-header
* https://developers.google.com/search/reference/robots_meta_tag#xrobotstag
    
Testing
-------
    
You should see an `X-Robots-Tag: noindex, nofollow` header in any Via 3 response. For example:

Launch an assignment and look at all the responses in dev tools: https://hypothesis.instructure.com/courses/125/assignments/874
    
Responses from the NGINX `/proxy/static/*` endpoint:
    
    curl -sSL -D - 'http://localhost:9083/proxy/static/https://h.readthedocs.io/_/downloads/en/latest/pdf/' -o /dev/null
    
Responses from the Pyramid app:
    
    curl -sSL -D - 'http://localhost:9083/route?url=https%3A%2F%2Fh.readthedocs.io%2F_%2Fdownloads%2Fen%2Flatest%2Fpdf' -o /dev/null
    curl -sSL -D - 'http://localhost:9083/pdf?url=https%3A%2F%2Fh.readthedocs.io%2F_%2Fdownloads%2Fen%2Flatest%2Fpdf' -o /dev/null
    
Responses from static files served by Whitenoise:
    
    curl -sSL -D - 'http://localhost:9083/static/ced8d00c07a038ba7e7b7077407ca60d/vendor/pdfjs-2/build/pdf.js' -o /dev/null

### DIFF
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -52,6 +52,7 @@ http {
             add_header "Cache-Control" "public, max-age=86400, stale-while-revalidate=604800";
 
             add_header "X-Via" "static-proxy, direct";
+            add_header "X-Robots-Tag" "noindex, nofollow";
         }
 
         location @handle_redirect {
@@ -104,6 +105,7 @@ http {
             proxy_set_header X-Request-Start "t=${msec}";
 
             add_header "X-Via" "compute";
+            add_header "X-Robots-Tag" "noindex, nofollow";
         }
     }
 }


jon-betts (changes requested):
I mean mostly yeah, but this should probably be in the `/` end-point which covers the app too. This would prevent indexing of any of the app routing and UI etc.

seanh:
For some reason this doesn't seem to work for the front page:

```
$ curl -sSL -D - 'http://localhost:9083/' -o /dev/null
HTTP/1.1 404 Not Found
Server: nginx/1.17.6
Date: Mon, 25 Jan 2021 18:03:08 GMT
Content-Type: text/html; charset=UTF-8
Content-Length: 153
Connection: keep-alive
```

I don't think it matters 'cos the front page is a 404, I just don't get it.

seanh:
@jon-betts I've tweaked this to apply to all Via 3's responses, not just the NGINX proxy ones

jon-betts:
I'll cross post some stuff from slack:

> I don't think you can mix and match add header directives at different levels as the later ones take total precedence and cause the previous ones to be ignored. (Maybe? Not sure) So it has to be in the same location blocks.
>
> That said, how you have it seems fine I think. No actual content can be crawled like this. You can go from one thing you shouldn't crawl to another, but presumably it shouldn't.
>
> I feel like they can cache our error pages if they desparately want to. We can also add it to them all?
>
> Again, how would they get to them without hitting a page that tells them not to index us? I suppose if they already have it might be too late. Maybe just suck it and see? It should be easy to check on

jon-betts (approved):
I think this should be fine. This should block all the ways into our app and any real content that could be indexed.

Probably best to get it out there and see if it does what we think. If anything (like the error pages) still shows up in Google we can have a think and take action.

PR #321, opened 2021-01-27
Added requirements target to run pip-tools in all environments

marcospri:
Updated requirements/*.txt with the result (only header changed)

Will update the cookie-cutter template with this soon but there `requirements.txt` it's just a a file (not a directory) like most projects so that might complicate the update from the template.


Makefile
```diff
@@ -88,4 +89,8 @@ rewriter:
 python:
 	@./bin/install-python
 
+.PHONY: requirements
+requirements:
+	@sh requirements/compile.sh
```
seanh: I think if `compile.sh` is executable and has a `#!` we should be able to just do `@requirements/compile.sh` here but :man_shrugging:

seanh (approved):

Makefile
marcospri: True, but required a `chmod u+x`, done :+1:

PR #322, opened 2021-01-27
Test with checkmate's blocklist in view_pdf

marcospri:
Added the checkmatelib dependency and did the checkmate test in direclty in the view.

Quick local testing, with `main` in checkmate's repo:

Blocked url `http://0.0.0.0:9082/pdf?url=42.228.75.1%2Fbin.sh`

vs

`http://0.0.0.0:9082/pdf?url=http%3A%2F%2Fwww.africau.edu%2Fimages%2Fdefault%2Fsample.pdf`

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -43,6 +43,7 @@ def pyramid_settings():
         "client_embed_url": "http://hypothes.is/embed.js",
         "nginx_server": "http://via3.hypothes.is",
         "via_html_url": "https://viahtml3.hypothes.is/proxy",
+        "checkmate_url": "http://localhost:9099",
     }
 
 
--- a/tests/unit/static_content_test.py
+++ b/tests/unit/static_content_test.py
@@ -2,6 +2,7 @@
 
 import re
 
+import httpretty
 import pytest
 from h_matchers import Any
 
@@ -35,6 +36,7 @@ class TestStaticContent:
         assert_cache_control(response.headers, ["public", "max-age=60"])
 
     def test_immutable_contents(self, test_app):
+
         salt = self.get_salt(test_app)
 
         response = test_app.get(f"/static/{salt}/favicon.ico")
@@ -50,6 +52,13 @@ class TestStaticContent:
         actually make a call with immutable assets and then scrape the HTML
         for the salt value.
         """
+        httpretty.register_uri(
+            httpretty.GET,
+            "http://localhost:9099/api/check",
+            status=204,
+            body={"data": []},
+        )
+
         response = test_app.get("/pdf?url=http://example.com")
         static_match = re.search("/static/([^/]+)/", response.text)
         assert static_match
--- a/tests/unit/via/views/view_pdf_test.py
+++ b/tests/unit/via/views/view_pdf_test.py
@@ -1,6 +1,7 @@
 import json
 from unittest.mock import sentinel
 
+import httpretty
 import pytest
 from h_matchers import Any
 from markupsafe import Markup
@@ -53,11 +54,30 @@ class TestViewPDF:
         )
 
     def test_caching_is_disabled(self, test_app):
+        httpretty.register_uri(
+            httpretty.GET,
+            "http://localhost:9099/api/check",
+            status=204,
+        )
+
         response = test_app.get("/pdf?url=http://example.com/foo.pdf")
 
         assert_cache_control(
             response.headers, ["max-age=0", "must-revalidate", "no-cache", "no-store"]
         )
+
+    def test_blocked_by_checkmate(self, test_app):
+        httpretty.register_uri(
+            httpretty.GET,
+            "http://localhost:9099/api/check",
+            status=200,
+            body='{"data": [{"type": "reason", "id": "not-explicitly-allowed",'
+            '"attributes": {"severity": "advisory"}}], "meta": {"maxSeverity": "advisory"},'
+            '"links": {"html": "http://localhost:9099/error"}}',
+        )
+        response = test_app.get("/pdf?url=http://example.com/foo.pdf")
+
+        assert response.status_code == 307
 
     def test_it_extracts_config(self, call_view_pdf, Configuration):
         response = call_view_pdf()
--- a/via/views/blocker.py
+++ b/via/views/blocker.py
@@ -1,0 +1,34 @@
+"""View decorators to integrate with checkmate's API."""
+import logging
+
+from checkmatelib import CheckmateClient, CheckmateException
+from pyramid.httpexceptions import HTTPTemporaryRedirect
+
+logger = logging.getLogger(__name__)
+
+
+def checkmate_block(view, url_param="url", allow_all=True):
+    """Intended to be used as a decorator for pyramid views.
+
+    The view must accept a url a query param.
+
+    :param url_param name of the query param that contains the URL to check
+    :allow_all Check against checkmate's allow list (True) or not.
+    """
+
+    def view_wrapper(context, request):
+        checkmate = CheckmateClient(request.registry.settings["checkmate_url"])
+
+        url = request.params[url_param]
+        blocked = None
+        try:
+            blocked = checkmate.check_url(url, allow_all=allow_all)
+        except CheckmateException:
+            logging.exception("Failed to check url against checkmate")
+
+        if blocked:
+            return HTTPTemporaryRedirect(location=blocked.presentation_url)
+
+        return view(context, request)
+
+    return view_wrapper
--- a/via/views/view_pdf.py
+++ b/via/views/view_pdf.py
@@ -5,6 +5,8 @@ from h_vialib import Configuration
 from markupsafe import Markup
 from pyramid import view
 
+from via.views.blocker import checkmate_block
+
 
 @view.view_config(
     renderer="via:templates/pdf_viewer.html.jinja2",
@@ -12,6 +14,7 @@ from pyramid import view
     # We have to keep the leash short here for caching so we can pick up new
     # immutable assets when they are deployed
     http_cache=0,
+    decorator=checkmate_block,
 )
 def view_pdf(context, request):
     """HTML page with client and the PDF embedded."""
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -4,188 +4,70 @@
 #
 #    pip-compile requirements/dev.in
 #
-attrs==20.2.0
-    # via jsonschema
-backcall==0.2.0
-    # via ipython
-bcrypt==3.2.0
-    # via paramiko
-cached-property==1.5.1
-    # via docker-compose
-certifi==2020.6.20
-    # via
-    #   -r requirements/requirements.txt
-    #   requests
-    #   sentry-sdk
-cffi==1.14.2
-    # via
-    #   bcrypt
-    #   cryptography
-    #   pynacl
-chardet==3.0.4
-    # via
-    #   -r requirements/requirements.txt
-    #   requests
-cryptography==3.2
-    # via paramiko
-decorator==4.4.2
-    # via
-    #   ipython
-    #   traitlets
-distro==1.5.0
-    # via docker-compose
-docker-compose==1.28.2
-    # via -r requirements/dev.in
-docker[ssh]==4.4.1
-    # via docker-compose
-dockerpty==0.4.1
-    # via docker-compose
-docopt==0.6.2
-    # via docker-compose
-gunicorn==20.0.4
-    # via -r requirements/requirements.txt
-h-pyramid-sentry==1.2.1
-    # via -r requirements/requirements.txt
-h-vialib==1.0.1
-    # via -r requirements/requirements.txt
-hupper==1.10.2
-    # via
-    #   -r requirements/requirements.txt
-    #   pyramid
-idna==2.10
-    # via
-    #   -r requirements/requirements.txt
-    #   requests
-importlib-metadata==1.7.0
-    # via jsonschema
-ipdb==0.13.4
-    # via -r requirements/dev.in
-ipython-genutils==0.2.0
-    # via traitlets
-ipython==7.16.1
-    # via
-    #   -r requirements/dev.in
-    #   ipdb
-jedi==0.17.2
-    # via ipython
-jinja2==2.11.2
-    # via
-    #   -r requirements/requirements.txt
-    #   pyramid-jinja2
-jsonschema==3.2.0
-    # via docker-compose
-markupsafe==1.1.1
-    # via
-    #   -r requirements/requirements.txt
-    #   jinja2
-    #   pyramid-jinja2
-newrelic==6.0.1.155
-    # via -r requirements/requirements.txt
-paramiko==2.7.2
-    # via docker
-parso==0.7.1
-    # via jedi
-pastedeploy==2.1.0
-    # via
-    #   -r requirements/requirements.txt
-    #   plaster-pastedeploy
-pexpect==4.8.0
-    # via ipython
-pickleshare==0.7.5
-    # via ipython
-plaster-pastedeploy==0.7
-    # via
-    #   -r requirements/requirements.txt
-    #   pyramid
-plaster==1.0
-    # via
-    #   -r requirements/requirements.txt
-    #   plaster-pastedeploy
-    #   pyramid
-prompt-toolkit==3.0.7
-    # via ipython
-ptyprocess==0.6.0
-    # via pexpect
-pycparser==2.20
-    # via cffi
-pygments==2.7.1
-    # via ipython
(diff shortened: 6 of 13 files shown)


via/views/view_pdf.py
```diff
@@ -16,8 +19,22 @@
...
+    original_pdf_url = context.url()
+
+    checkmate = CheckmateClient(request.registry.settings["checkmate_url"])
+
+    blocked = None
+    try:
+        blocked = checkmate.check_url(original_pdf_url, allow_all=True)
+
+    except CheckmateException as err:
+        logging.exception("Failed to check url against checkmate with error: %s", err)
```
marcospri: I think it makes sense to log the exception but allow the request to continue at least now that via3 is a different service than via

via/views/view_pdf.py
```diff
@@ -16,8 +19,22 @@
 def view_pdf(context, request):
     """HTML page with client and the PDF embedded."""
 
+    original_pdf_url = context.url()
+
+    checkmate = CheckmateClient(request.registry.settings["checkmate_url"])
```
marcospri: I just dropped all of this here in the view but maybe there's a better pyramid mechanism I could use.

I'd say https://github.com/hypothesis/checkmate/issues/135 would do something very simlar (same?) so I could create a function to avoid duplication.

seanh: I think the view is fine.

I'm sure there's some Pyramid thing that could be done where a view can just declare (for example in its `@view_config`) that it wants Checkmate checking, and the check would happen automatically and the view would not even be called if the check failed. But I don't think it's necessary.

We're going to want to add the same Checkmate support to a second view in Via 3 as well (see https://github.com/hypothesis/checkmate/issues/135). At that point it might be worth putting this code into a shared helper function that just returns the `HTTPTemporaryRedirect` or `None`. Or actually doing a fancier Pyramid integration.

via/views/view_pdf.py
seanh: Do you actually need to pass `err` here? I think `logging.exception()` logs the current exception automatically

seanh: Yeah, I think "fail open" is the right approach for LMS's private instance of Via 3. Whereas for public Via, I think we want it to crash if calling Checkmate fails

marcospri: > Do you actually need to pass `err` here? I think `logging.exception()` logs the current exception automatically

I think you are right,, it will create a long message string too, will just remove that part.

@seanh It does by default but can't confirm 100% looking at the pyramid-sentry repo, would `logging.exception` create one in sentry?

seanh: I _think_ it will due to Sentry's logging integration (which is enabled by default): https://docs.sentry.io/platforms/python/guides/logging/

via/views/blocker.py
```diff
@@ -0,0 +1,35 @@
+"""Helper view decorators to integrate with checkmate's API."""
+import logging
+
+from checkmatelib import CheckmateClient, CheckmateException
+from pyramid.httpexceptions import HTTPTemporaryRedirect
+
+logger = logging.getLogger(__name__)
+
+
+def checkmate_block(view, url_param="url", allow_all=True):
```
marcospri, line 10: Ended up doing a decorator as it was just simpler than a plain function

If this approach works and we use it multiple services we could eventually move it to checkmatelib

seanh: Nice. Yeah I agree checkmatelib could provide Pyramid things like view derivers. It could even provide an actual Pyramid extension if that becomes necessary.

I **don't** think you need to change this (the decorator is fine) but just FYI there's a similar Pyramid feature named "view derivers" that can also be useful for doing this sort of thing: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#view-derivers View configs can pass arguments to view derivers, see: https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#custom-view-derivers

We have a view deriver in LMS for example: <https://github.com/hypothesis/lms/blob/ac8bbad5c6b96b9fff1ec80adc01d97aedfac7a3/lms/validation/__init__.py#L65-L95> Example of a view using it: <https://github.com/hypothesis/lms/blob/ac8bbad5c6b96b9fff1ec80adc01d97aedfac7a3/lms/views/api/canvas/authorize.py#L78>

tests/unit/conftest.py
```diff
@@ -86,3 +87,11 @@ def assert_cache_control(headers, cache_parts):
...
         headers["Cache-Control"].split(", ") == Any.list.containing(cache_parts).only()
     )
+
+
+@pytest.fixture()
+def CheckmateClient():
+    with mock.patch(
+        "via.views.blocker.CheckmateClient", spec_set=True
+    ) as CheckmateClient:
+        yield CheckmateClient
```
seanh: This should use our `patch` fixture, which'll handle the patching and unpatching and autospeccing etc:

```suggestion
@pytest.fixture()
def CheckmateClient(patch):
    return patch("via.views.blocker.CheckmateClient")
```

via/views/blocker.py
marcospri: Deriver looks like the pyramid-y way of doing it but it's also very similar so I'd say we can convert it into one once we move it out to checkmatelib or extension.

tests/unit/static_content_test.py
```diff
@@ -34,7 +34,9 @@ def test_get_static_content(self, url, mime_type, test_app):
 
         assert_cache_control(response.headers, ["public", "max-age=60"])
 
-    def test_immutable_contents(self, test_app):
+    def test_immutable_contents(self, test_app, CheckmateClient):
+        CheckmateClient.return_value.check_url.return_value = None
```
seanh: Oof, this whole file is actually functional tests, and the `test_app` fixture it uses belongs in the functional tests fixtures. Not sure how this happened but we should write a PR to move these out into `tests/functional/`, these aren't unit tests.

I'm not sure what functional tests for Via 3 should do about the fact that Via 3 calls Checkmate. Reaching in to `views/blocker.py` and patching the `CheckmateClient` is one option, but it means this functional test is tied to an implementation detail of `blocker.py`.

Another option would be to use [httpretty](https://httpretty.readthedocs.io/) (which is [already installed in Via 3's tests](https://github.com/hypothesis/checkmate/blob/90a9eec4ee0dbcda85cd33f8ed1a7293d47390dc/tests/unit/conftest.py#L108-L121)) to fake the response from Checkmate:

```suggestion
    def test_immutable_contents(self, test_app):
        import httpretty
        httpretty.register_uri(
            httpretty.GET,
            "http://localhost:9099/api/check",
            status=204,
        )
```

(With the `import httpretty` moved to the top of the file, of course.) This means that we're testing the app in integration with `checkmatelib`, and the test is no longer dependent on `blocker.py` internals. On the other hand, the test now has knowledge of what Checkmate's public API is.

I think the ideal solution here would be for `checkmatelib` to provide test helper functions for app's tests to call. So Via 3's tests would call something like `checkmatelib.allow("example.com")` and that would do whatever is necessary (mocking, registering httpretty URIs, ...) to make `check_url()` response that `"example.com"` is allowed. That way the Checkmate-specific knowledge all lives in checkmatelib. It's probably not worth doing this for now thoughl

seanh: The `httpretty.register_uri()` should be done in the `get_salt()` helper below, rather than in the test. The `get_salt()` helper is what actually causes the requests to Checkmate to get sent

tests/unit/conftest.py
seanh: This should be moved into `blocker_test.py` as well, since it's patching a name in `blocker.py`. I don't think tests for _other_ files (like `view_pdf.py`) should patch a name in `blocker.py` since that couples `view_pdf.py`'s tests to internal details of `blocker.py`.

`static_content_test.py` and `view_pdf_test.py` should use `httpretty` instead of this patch, see my other comments.

It's not quite the same thing as you're doing here, but we follow this pattern where `test_foo.py` should _only_ patch names in the corresponding `foo.py` module's namespace, we avoid patching names elsewhere or having global patch fixtures 'cos it can get really confusing. See this comment for why: https://github.com/hypothesis/via3/pull/222#discussion_r469865055

tests/unit/via/views/view_pdf_test.py
```diff
@@ -52,13 +52,24 @@ def test_we_escape_quote_literals_in_urls_to_prevent_XSS(
             f"{pyramid_settings['nginx_server']}/proxy/static/a\"b"
         )
 
-    def test_caching_is_disabled(self, test_app):
+    def test_caching_is_disabled(self, test_app, CheckmateClient):
+        CheckmateClient.return_value.check_url.return_value = None
```
seanh: Ugh, the existing test here is a functional test mixed in with the unit tests. It's using `test_app.get(...)` which actually sends a simulated request through the entire app at the WSGI level (it uses [webtest](https://docs.pylonsproject.org/projects/webtest/en/latest/)), so it's running the entire Via 3 app and Pyramid and everything. That's a functional test, not a unit test for the view.

I think it should use `httpretty` rather than mocking `CheckmateClient`, same as I commented on `static_content_test.py`.

And like `static_content_test.py` we should send a follow-up PR to move it out into `tests/functional/` because it's not a unit test.

tests/unit/via/views/view_pdf_test.py
```diff
@@ -52,13 +52,24 @@ def test_we_escape_quote_literals_in_urls_to_prevent_XSS(
...
         )
 
+    def test_blocked_by_checkmate(self, test_app, CheckmateClient):
+        block = Mock(presentation_url="http://checkmate.com/blocked")
+        CheckmateClient.return_value.check_url.return_value = block
+
+        response = test_app.get("/pdf?url=http://example.com/foo.pdf")
+
+        assert response.status_code == 307
+        assert response.location == block.presentation_url
```
seanh: This is a functional test so it doesn't belong here. I can see that you were following the example of the existing ` test_caching_is_disabled()` above but unfortunately that test is wrong. Options:

1. Just don't test this. We have unit tests for the `checkmate_block()` view decorator. Applying the view decorator to the view is purely declarative, it's just `@view_config(..., decorator=checkmate_block)`, so we don't need to test that.

2. If you _do_ want to test it add a functional test in `tests/functional/` for it. That would basically be the same test as you've added here but in `tests/functional/`, and I think it'd use `httpretty` to set up Checkmate's response rather than mocking `CheckmateClient`.

A view's configuration (whatever is declared in its `@view_config`) isn't part of what that view's unit tests cover. For example one of the things in view config is `route_name="foo"` and other predicates the decide when that view is called and when it isn't. But the view's unit tests don't include a test that _"When a `GET` request is sent to `/pdf/` it calls the `view_pdf` view"_. Similarly `renderer="json"` in a `@view_config` will turn the view's reponse into JSON (serializing a dict that the view returns into a JSON string, setting the Content-Type header to application/json, ...) but we don't write view unit tests for that. If we wanted to test these things we'd write functional tests.

tests/unit/via/views/test_blocker.py
```diff
@@ -0,0 +1,53 @@
...
+        CheckmateClient.return_value.check_url.side_effect = CheckmateException()
+        request = make_request(params={"url": url})
+
+        response = dummy_view(None, request)
+
+        # Request continues despite Checkmate errors
+        assert response.status_code == 200
+        assert response.text == "ok"
+
+    @pytest.fixture()
```
seanh:
```suggestion
    @pytest.fixture
```

tests/unit/via/views/test_blocker.py
```diff
@@ -0,0 +1,53 @@
+from unittest.mock import Mock
```
seanh: Should be renamed to `blocker_test.py` (we put the `_test` on the end because that works better with autocompletion)

tests/unit/via/views/test_blocker.py
```diff
@@ -0,0 +1,53 @@
...
+
+        response = dummy_view(None, request)
+
+        # Request continues despite Checkmate errors
+        assert response.status_code == 200
+        assert response.text == "ok"
+
+    @pytest.fixture()
+    def block(self):
+        return Mock(presentation_url="http://checkmate/blocked")
```
seanh: I think this is a mock of checkmatelib's `BlockResponse`? If so, might want to autospec it:

```python
return mock.create_autospec(checkmatelib.client.BlockResponse, instance=True, spec_set=True)
```

The reason is that the mock will catch breakage if checkmatelib's interface changes.

It looks to me like `BlockResponse` is public in checkmatelib (since `checkmatelib.CheckmateClient.check_url()` returns them) so I think checkmatelib should expose `BlockResponse` in its `__init__.py` as it does for `CheckmateClient` and `CheckmateException`: https://github.com/hypothesis/checkmatelib/blob/6b94ebff539d28e919b939841372d7453aae3ac4/src/checkmatelib/__init__.py. Then it'd just be `create_autospec(checkmatelib.BlockResponse, ...)`.

If we can't autospec this mock we should at least manually spec it (see <https://www.seanh.cc/2017/03/17/the-problem-with-mocks/#passing-a-list-of-strings-to-spec_set>), unspecced mocks are dangerous 'cos they'll respond successfully to any method call.

seanh (changes requested):
Looks good and I think the view decorator is a nice solution :+1:

Some requests (see comments for details):

- [x] Rename `test_blocker.py` to `blocker_test.py`
- [x] Tweak the `CheckmateClient` fixture to use `patch`
- [x] Move the `CheckmateClient` fixture into `blocker_test.py`
- [x] Change `static_content_test.py` and `view_pdf_test.py` to use `httpretty` instead of the `CheckmateClient` fixture
- [x] Spec the `block` fixture

Stuff that we can probably leave for follow-up PRs:

* functests need to be added to Via 3
* Any tests that use the `test_app` fixture are functional tests, and need to be moved into `tests/functional/`:
  * All of `test_static_content.py`
  * The existing `test_caching_is_disabled()` and your new `test_blocked_by_checkmate()` in `view_pdf_test.py`
  * Anything else that uses `test_app`

tests/unit/via/views/test_blocker.py
seanh: Probably rename the fixture to `block_response` as well, following the `BlockResponse` class's name

via/views/blocker.py
```diff
@@ -0,0 +1,35 @@
+"""Helper view decorators to integrate with checkmate's API."""
```
seanh:
```suggestion
"""View decorators to integrate with checkmate's API."""
```

via/views/blocker.py
```diff
@@ -0,0 +1,35 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def checkmate_block(view, url_param="url", allow_all=True):
+    """Intended to be used as a decorator for pyramid views.
+
+    The view must accept a url a query param.
+
+
```
seanh:
```suggestion
```

tests/unit/via/views/view_pdf_test.py
```diff
@@ -52,13 +53,34 @@ def test_we_escape_quote_literals_in_urls_to_prevent_XSS(
             f"{pyramid_settings['nginx_server']}/proxy/static/a\"b"
         )
 
+    @httpretty.activate
```
seanh:
```suggestion
```

I think this is unnecessary? We have a global pytest fixture that activates it for every test

tests/unit/via/views/view_pdf_test.py
```diff
@@ -52,13 +53,34 @@ def test_we_escape_quote_literals_in_urls_to_prevent_XSS(
...
+            status=204,
+        )
+
         response = test_app.get("/pdf?url=http://example.com/foo.pdf")
 
         assert_cache_control(
             response.headers, ["max-age=0", "must-revalidate", "no-cache", "no-store"]
         )
 
+    @httpretty.activate
```
seanh:
```suggestion
```

seanh (approved):
Looks good :+1: I suggested some minor changes but let's get this one merged.

We need to add the `CHECKMATE_URL` envvar to QA and prod before deploying it, right?

And we can test it on QA by making sure that:

1. A blocklisted domain is blocked. For example this <https://qa-via3.hypothes.is/pdf?url=https://vimeo.com/> should be blocked (currently it is not blocked and loads PDF.js, but that fails because vimeo.com isn't a PDF file)

2. A non-blockedlisted domain is not blocked. E.g. https://qa-via3.hypothes.is/pdf?url=https://h.readthedocs.io/_/downloads/en/latest/pdf/

And test that our QA test assignments in Canvas still work

via/views/blocker.py
jon-betts: A smattering of helpers and things seems like a good thing to include

PR #323, opened 2021-01-27
Require /proxy/static/ links to be signed

seanh:
See:

* https://www.nginx.com/blog/securing-urls-secure-link-module-nginx-plus/
* http://nginx.org/en/docs/http/ngx_http_secure_link_module.html#secure_link

Testing
-------

Unsigned requests are refused with 400:

    $ curl -sSL -D - 'http://localhost:9083/proxy/static/https://example.com' -o /dev/null
    HTTP/1.1 400 Bad Request
    Server: nginx/1.17.6
    Date: Wed, 27 Jan 2021 16:28:36 GMT
    Content-Type: text/html
    Content-Length: 157
    Connection: close

Generate a signature over a non-expired timestamp (`1640995200`):

    $ echo -n '1640995200/proxy/static/https://example.com seekrit' | openssl md5 -binary | openssl base64 | tr +/ -_ | tr -d =
    QwRvb9F8yBIsS1lR91dqGw

Signed requests work:

    $ curl -sSL -D - 'http://localhost:9083/proxy/static/https://example.com?md5=QwRvb9F8yBIsS1lR91dqGw&expires=1640995200' -o /dev/null
    HTTP/1.1 200 OK
    ...

If you try breaking the request (change the URL to be proxied, the MD5, or the timestamp, or generate a new MD5 with the wrong secret, etc) it'll 400.

Generate a signature over an expired timestamp (`1611764687`):

    $ echo -n '1611764687/proxy/static/https://example.com seekrit' | openssl md5 -binary | openssl base64 | tr +/ -_ | tr -d =
    SCAi0S5wCfadeuM8A_Hc6w

Expired links are refused with a 404 (this is supposed to be a 410, but another part of our `nginx.conf` turns 410s into 404s):

    $  curl -sSL -D - 'http://localhost:9083/proxy/static/https://example.com?md5=SCAi0S5wCfadeuM8A_Hc6w&expires=1611764687' -o /dev/null
    HTTP/1.1 404 Not Found
    ...


conf/nginx/nginx.conf
```diff
@@ -35,6 +35,31 @@ http {
...
 
         location ~ /proxy/static/ {
+            # $secure_link will be set to:
+            #
+            # "" if the MD5 in the URL is missing or wrong
+            #
+            # "0" if the MD5 is valid but the expires time has passed
+            #
+            # "1" if the MD5 is valid and the expires time has not passed
+            secure_link $arg_md5,$arg_expires;
```
seanh: I think we can replace `$arg_md5` and `$arg_expires` here with `$arg_WHATEVER` in order to change the names of the `md5` and `expires` params.

conf/nginx/nginx.conf
```diff
@@ -35,6 +35,31 @@ http {
...
+
+            # The string that we expect the MD5 to be a hash of:
+            #
+            # $secure_link_expires is the value of the link's `expires` param ($arg_expires)
+            #
+            # $uri is the requested path, e.g. "/proxy/static/https://example.com"
+            #
+            # $nginx_secure_link_secret comes from the NGINX_SECURE_LINK_SECRET environment variable.
+            secure_link_md5 "$secure_link_expires$uri $nginx_secure_link_secret";
```
seanh: We may want to include the origin part (`https://via3.hypothes.is/`) in this so that links generated by Via 3 QA can't be used on Via 3 prod (`$uri` is just the path part). I'd have to figure out which NGINX `$variable` to use for that. Dev, QA and prod will use different secrets so the links won't be transferable anyway, so maybe we needn't bother.

conf/nginx/nginx.conf
jon-betts: Yep, was just going to say. For other things we use `via.WHATEVER` for the params. Not sure how the dot would translate, but it seems relatively unlikely to clash with other sites.

In `checkmate` we used `sec` for the param where it's completely under our control, just so we have something short. This is as we are adding params to an arbitrarily long URL, it's good to bloat it as little as possible. Maybe `via.sec` and `via.exp` if dots work?

The `secure_link_secret` looks like the way to do it if you want a URL based one. So we calculate the hash and send it as the URL, then it either goes on it's merry way, or gets discarded. Might be a good way to go, although as you say, no expiry.

seanh: Yeah, I think we need them to expire unfortunately

conf/nginx/nginx.conf
```diff
@@ -26,16 +26,50 @@ http {
...
+        location ~ ^/proxy/static/(?<sec>[a-zA-Z0-9-_]+)/(?<exp>\d+)/ {
             proxy_intercept_errors on;
+
+            # $secure_link will be set to:
+            #
+            # "" if the MD5 digest in the URL ($sec) is missing or wrong
+            #
+            # "0" if the digest is valid but the timestamp in the URL ($exp) has expired
+            #
+            # "1" if both the digest and the timestamp are valid
```
jon-betts, line 45:
```suggestion
            # $secure_link will be set to:
            #     "" if the MD5 digest in the URL ($sec) is missing or wrong
            #     "0" if the digest is valid but the timestamp in the URL ($exp) has expired
            #     "1" if both the digest and the timestamp are valid
```

For de eyes

conf/nginx/nginx.conf
```diff
@@ -26,16 +26,50 @@ http {
...
+
+            # The string that we expect the MD5 digest in the URL ($sec) to be a hash of:
+            #
+            # $secure_link_expires is the timestamp in the URL ($exp)
+            #
+            # $url_to_be_proxied is the URL to be proxied, e.g. "https://example.com/foo.pdf", *without* the query string part
+            #
+            # $uri is the URL to be proxied.
+            #
+            # $nginx_secure_link_secret comes from the NGINX_SECURE_LINK_SECRET environment variable.
```
jon-betts, line 63:
```suggestion
            # The string that we expect the MD5 digest in the URL ($sec) to be a hash of:
            #     `$secure_link_expires` is the timestamp in the URL ($exp)
            #     `$url_to_be_proxied` is the URL to be proxied, e.g. "https://example.com/foo.pdf", *without* the query string part
            #     `$uri` is the URL to be proxied.
            #     `$nginx_secure_link_secret` comes from the NGINX_SECURE_LINK_SECRET environment variable.
```

via/views/view_pdf.py
```diff
@@ -31,6 +34,31 @@ def view_pdf(context, request):
...
+    hash_.update(f"/proxy/static/{exp}/{url} {secret}".encode("utf-8"))
+    sec = hash_.digest()
+    sec = b64encode(sec)
+    sec = sec.replace(b"+", b"-")
+    sec = sec.replace(b"/", b"_")
+    sec = sec.replace(b"=", b"")
+    sec = sec.decode()
+
+    # Construct the URL.
+    return _string_literal(f"{nginx_server}/proxy/static/{sec}/{exp}/{url}")
```
jon-betts, line 59: Is any part of this general to NGINX. Could we extract an "nginx_hash" function from this or something?

jon-betts (commented):
Not a review, just some comments from the sidelines

via/views/view_pdf.py
```diff
@@ -31,6 +34,31 @@ def view_pdf(context, request):
...
+    exp = int(exp)
+
+    # Compute the digest to put into the URL.
+    hash_ = hashlib.md5()
+    hash_.update(f"/proxy/static/{exp}/{url} {secret}".encode("utf-8"))
+    sec = hash_.digest()
+    sec = b64encode(sec)
+    sec = sec.replace(b"+", b"-")
+    sec = sec.replace(b"/", b"_")
+    sec = sec.replace(b"=", b"")
```
marcospri, line 55: Could use https://docs.python.org/3/library/base64.html#base64.urlsafe_b64encode if it's ok to have "="

seanh:
Closing in favour of https://github.com/hypothesis/via3/pull/353

PR #324, opened 2021-01-29
Split unit and functional tests

marcospri:
Limited to move the existing ones using test_app to the test/functional

Coverage was not affected by the move.


tox.ini
```diff
@@ -60,6 +61,7 @@ commands =
     tests: coverage run -m pytest {posargs:tests/unit/}
     tests: -coverage combine
     tests: coverage report
+    functests: pytest {posargs:tests/functional/}
```
seanh, line 65: We should add `functests` to GitHub Actions here: <https://github.com/hypothesis/via3/blob/54e660f4999e7fedc7e98189eb07591fea032925/.github/workflows/continuous-integration.yml#L27>

seanh: We could also change Jenkins to run the functests rather than the tests, here: https://github.com/hypothesis/via3/blob/54e660f4999e7fedc7e98189eb07591fea032925/Jenkinsfile#L31

We usually run some subset of the tests/checks on Jenkins rather than all of them. For h and lms we run the functests. For bouncer and checkmate we run the unit tests because they don't have functests. It's just meant as a final sanity check.

seanh (changes requested):
Looks good, just needs to be added to GitHub Actions and Jenkins. Thanks for this

seanh (approved):
:+1: Let's merge it

PR #339, opened 2021-02-04
Remove some headers that we're sending to gunicorn

seanh:
I don't think these are necessary? But I need to test to make sure that we don't break anything.

It looks like we also have some code in `headers.py` to prevent sending these headers on to third-parties, might be able to remove that too.


seanh:
Rebased

PR #341, opened 2021-02-04
Use nginx's default pidfile location

seanh:
`pid` is `logs/nginx.pid` by default:

http://nginx.org/en/docs/ngx_core_module.html#pid

I don't see why we need to change this? It looks like it might have been copied into Via 3 from h's nginx, but I don't see why we need to change it in h either.


robertknight (approved):
I think you're right about this being copied from h. The location of the pidfile in h dates back to the introduction of the nginx configuration in 03b365eb5f15f580d59f9cae2cde85089305ce0d and no explicit explanation for the custom location was given there.

PR #342, opened 2021-02-04
Use nginx's default number of worker connections

seanh:
...and open files.

I think it probably does make sense to tune these settings to match with our hardware, and I think nginx's defaults for these might be very conservative, so this PR is more of a placeholder for discussion rather than a serious suggestion that we just remove these.

* If we're going to set these I'd like there to be a comment explaining why the values were chosen
* I don't know why the particular numbers here were chosen
* I suspect they were chosen for reasons that never made sense or that at least don't make sense anymore (I think they came from Via 2 which was an entirely different beast, an OpenResty-based thing that did both PDF and HTML proxying, and probably ran on different EC2 instance types
* Are `worker_connections` and `worker_rlimit_nofile` the only two settings we should be tuning? We might want to consult an nginx tuning guide

### DIFF
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -9,15 +9,14 @@ pid /var/lib/hypothesis/nginx.pid;
 # Log errors to stderr: http://nginx.org/en/docs/ngx_core_module.html#error_log
 error_log /dev/stderr;
 
-# How many files a worker can open at once: http://nginx.org/en/docs/ngx_core_module.html#worker_rlimit_nofile
-# This has to be greater than or equal to worker_connections below.
-# TODO: Why have we chosen this number?
-worker_rlimit_nofile 7192;
-
 events {
     # How many connections a worker can open at once: http://nginx.org/en/docs/ngx_core_module.html#worker_connections
-    # TODO: Why have we chosen this number?
-    worker_connections  1024;
+    #
+    # This number hasn't been chosen for any particularly good reason. It's
+    # just an educated guess. The default of 512 seemed low and increasing it
+    # by 4x seemed like it might increase performance with little risk of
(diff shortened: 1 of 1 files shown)


robertknight:
References I found related to these settings:

https://github.com/hypothesis/deployment/pull/142 - A change to the value in h
https://github.com/hypothesis/deployment/issues/141#issuecomment-708443623 - The associated issue in the deployment repo

In the case of h the need to increase the limits were related to the WebSocket server which doesn't apply here.

The default for `worker_connections` is 512 according to http://nginx.org/en/docs/ngx_core_module.html#worker_connections.

The default for `worker_rlimit_nofile` must be higher than `worker_connections` and nginx doesn't specify a default, so I think it is whatever the process gets by default. In a production Via 3 container these are:

```
~ $ ulimit -a
-f: file size (blocks)             unlimited
-t: cpu time (seconds)             unlimited
-d: data seg size (kb)             unlimited
-s: stack size (kb)                8192
-c: core file size (blocks)        0
-m: resident set size (kb)         unlimited
-l: locked memory (kb)             64
-p: processes                      7745
-n: file descriptors               65824
-v: address space (kb)             unlimited
-w: locks                          unlimited
-e: scheduling priority            0
-r: real-time priority             0
```

So the "file descriptors" limit is already way higher than what was configured here, in which case we don't need to adjust it.  The default `worker_connections` value may be on the low side from what I can tell and that increasing it by a factor of 2x or 4x is probably low risk.

What would be useful information here is to find out the peak values that are reached in practice today.

seanh:
Rebased

seanh:
Adjusted this based on @robertknight 's comments. I'm gonna deploy it and see if anything breaks

PR #344, opened 2021-02-04
NGINX: Don't use Google's DNS

seanh:
When Via 3's NGINX needs to connect to the third-party servers it's being asked to proxy, don't use Google's public 8.8.8.8 DNS to look them up.

We do actually need to set `resolver` to something--by deleting it this commit actually breaks Via 3 at least in dev. So this is a placeholder for a discussion.  What DNS resolver should it use?

Note that it has to work both in production and in dev. We can potentially set this to different values in prod and dev though (or set
it to prod only not dev).

### DIFF
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -20,11 +20,6 @@ events {
 }
 
 http {
-    # Configure the DNS that nginx uses to connect to the servers it's proxying.
-    # http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver
-    # TODO: Do we need to change this? Why have we chosen this value?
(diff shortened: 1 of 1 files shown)


jon-betts (commented):
I think this sounds very wrong to me. Presumably AWS has DNS resolution we should be using inside AWS rather than making external calls.

I think any external traffic is likely to:

 * Be slower
 * Incur bandwidth issues
 * Miss out on any special AWS sauce that might have been added (who knows what)
 
 This very much isn't my area though

seanh:
I'm gonna close this for now as we're unsure what the solution is. I've created an ops spike issue for next sprint: https://github.com/hypothesis/playbook/issues/569

PR #348, opened 2021-02-04
Rename `direct_proxy.conf` to just `proxy.conf`

seanh:
I'm not sure the "direct" really means anything. Also add a docstring to the top of the file.

### DIFF
--- a/conf/nginx/includes/proxy.conf
+++ b/conf/nginx/includes/proxy.conf
@@ -1,3 +1,11 @@
+# Proxy to the URL that's given in the path.
+#
+# For example if the request is:
+#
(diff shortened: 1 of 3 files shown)


jon-betts (changes requested):
The "direct" I think was conceived of when we imagined there'd be inline rewriting too. 

My only feedback is that might want to scope it down further and say "pdf_proxy" to indicate it's intended purpose. Although this currently is a general purpose proxying thing, it doesn't mean we want it to be.

If there was a way to scope it to PDFs only, I think we'd do it. So maybe reveal the intent?

seanh:
Rebased

PR #349, opened 2021-02-04
Follow redirects internally

seanh:
When the third-party server that we're proxying responds with a 3xx redirect, on master we rewrite the redirect's `Location` header so that it begins with `https://via3.hypothes.is/proxy/static/` and return the rewritten redirect to the browser.

The problem with this is that in the future we want to use NGINX's secure links module to require `/proxy/static/` URLs to be signed. In the case of redirects that would mean that NGINX would have to sign the rewritten `Location` URL before sending it back to the browser, but NGINX's secure links module can only verify signatures, it can't sign things.

So this commit changes our NGINX to follow redirects internally instead of sending them back to the browser. When the third-party server that we're proxying responds with a 3xx NGINX will follow the redirect itself, downloading the file from the redirects `Location` header and returning it to the browser. 3xx's responses from third-party servers are "hidden" from browsers.

This works when the third-party server responds with:

* An absolute redirect

* A relative redirect

* A chain of multiple redirects (NGINX will follow each in turn)

* A too-long chain of redirects or an infinite cycle of redirects. After 10 redirects NGINX will stop and send a 500 back to the browser. (Unfortunately I don't think we can customize the status code of this response.)

* Redirect URLs with query params in them

* A redirect or chain of redirects where one of the `Location` URLs responds with an error. Our usual error handling kicks in where we hide the third-party error responses and status codes from the browser.

### DIFF
--- a/conf/nginx/includes/cors.conf
+++ b/conf/nginx/includes/cors.conf
@@ -1,0 +1,7 @@
+# Prevent third-party servers from adding CORS controls.
+
+proxy_hide_header "Access-Control-Allow-Origin";
+proxy_hide_header "Access-Control-Allow-Credentials";
+proxy_hide_header "Access-Control-Allow-Methods";
+proxy_hide_header "Access-Control-Allow-Headers";
+proxy_hide_header "Access-Control-Expose-Headers";
--- a/conf/nginx/includes/direct_proxy.conf
+++ b/conf/nginx/includes/direct_proxy.conf
@@ -23,18 +23,3 @@ rewrite ^/proxy/static/(.*)$ $1? break;
 # The proxy directly to the resulting URL
 # Add back on the args which we stripped above so we don't get them twice
 proxy_pass $uri$is_args$args;
-
-# If the proxied server returns a redirect response change the URL in the
-# response's `Location` header to be proxied through us.
-# For example `Location: https://example.com` might get changed to
-# `Location: https://via3.hypothes.is/proxy/static/https://example.com`.
-# http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect
-proxy_redirect ~^(https?:\/\/)(.*)$ $original_scheme://$http_host/proxy/static/$1$2;
-proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/https://$proxy_host$1;
-
-# Prevent the third party from adding CORS controls
-proxy_hide_header "Access-Control-Allow-Origin";
-proxy_hide_header "Access-Control-Allow-Credentials";
-proxy_hide_header "Access-Control-Allow-Methods";
-proxy_hide_header "Access-Control-Allow-Headers";
-proxy_hide_header "Access-Control-Expose-Headers";
--- a/conf/nginx/includes/errors.conf
+++ b/conf/nginx/includes/errors.conf
@@ -1,0 +1,37 @@
+# Follow chains of multiple redirects (when the requested URL redirects to
+# another URL which in turn redirects to another).
+#
+# This is limited to following up to 10 redirects.
+#
+# recursive_error_pages is also needed to enable our 4xx and 5xx error handling
+# to kick in when the requested URL redirects to a URL that 4xx's or 5xx's.
+#
+# http://nginx.org/en/docs/http/ngx_http_core_module.html#recursive_error_pages
+recursive_error_pages on;
+
+# Intercept 3xx, 4xx and 5xx responses from the servers we're
+# proxying and process them through the error_page directives
+# below: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_intercept_errors
+proxy_intercept_errors on;
+
+# Don't pass 3xx responses from the servers we're proxying back to
+# the browser. Instead process them through the @handle_redirect
+# location below.
+# http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page
+error_page 301 302 307 = @handle_redirect;
+
+# Don't pass 4xx or 5xx responses from the servers we're proxying
+# back to the browser. Instead process them through the @proxy_*
+# locations defined below.
+#
+# If users are proxying a third-party server that is returning
+# error responses we don't want Via 3 to return those same 4xx or
+# 5xx status codes to the browser because this would mess up our
+# monitoring and alerting. For example it might trigger an alarm
+# that Via 3 is unhealthy because it's sending too many error
+# responses.
+# http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page
+error_page 404 410 = @proxy_not_found;
+error_page 420 429 = @proxy_too_many_requests;
+error_page 400 401 402 403 405 406 407 408 409 411 412 413 414 415 416 417 418 422 423 424 425 426 428 431 444 449 450 451 = @proxy_client_error;
+error_page 500 501 502 503 504 505 506 507 508 509 510 511 598 599 = @proxy_upstream_error;
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -28,26 +28,6 @@ http {
     # Log accesses to stdout: http://nginx.org/en/docs/http/ngx_http_log_module.html#access_log
     access_log /dev/stdout;
 
-    # Set $original_scheme to the protocol that the browser used to connect to Via
-    # ("http" or "https"), based on the X-Forwarded-Proto header. We'll need
-    # this variable later.
-    #
-    # $original_scheme is needed because AWS's load balancer connects to the
-    # app server using HTTP so nginx's default $scheme is always "http" in
-    # production.
-    #
-    # If there's no X-Forwarded-Proto header $original_scheme falls back to
-    # $scheme ("http").
-    #
-    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
-    # http://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_
-    # http://nginx.org/en/docs/http/ngx_http_map_module.html#map
-    # http://nginx.org/en/docs/http/ngx_http_core_module.html#var_scheme
-    map $http_x_forwarded_proto $original_scheme {
-        "" $scheme;
-        default $http_x_forwarded_proto;
-    }
-
     include includes/app_upstream.conf;
 
     server {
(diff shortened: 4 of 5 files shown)


conf/nginx/includes/cors.conf
```diff
@@ -0,0 +1,7 @@
+# Prevent third-party servers from adding CORS controls.
+
+proxy_hide_header "Access-Control-Allow-Origin";
+proxy_hide_header "Access-Control-Allow-Credentials";
+proxy_hide_header "Access-Control-Allow-Methods";
+proxy_hide_header "Access-Control-Allow-Headers";
+proxy_hide_header "Access-Control-Expose-Headers";
```
seanh, line 7: This was being applied to the initial proxy request (in `direct_proxy.conf`) but would not be applied to the subsequent requests when following redirects. I moved it out into a separate file, so the `@handle_redirect` location can `include` it so that it applies to all the requests.

_However_, I'm pretty sure this is more unnecessary that should be removed anyway! https://hypothes-is.slack.com/archives/C4K6M7P5E/p1612461980452000

conf/nginx/includes/direct_proxy.conf
```diff
@@ -23,18 +23,3 @@ rewrite ^/proxy/static/(.*)$ $1? break;
...
 # Add back on the args which we stripped above so we don't get them twice
 proxy_pass $uri$is_args$args;
-
-# If the proxied server returns a redirect response change the URL in the
-# response's `Location` header to be proxied through us.
-# For example `Location: https://example.com` might get changed to
-# `Location: https://via3.hypothes.is/proxy/static/https://example.com`.
-# http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect
-proxy_redirect ~^(https?:\/\/)(.*)$ $original_scheme://$http_host/proxy/static/$1$2;
-proxy_redirect ~^(.*)$ $original_scheme://$http_host/proxy/static/https://$proxy_host$1;
```
seanh, line 33: This is to do with rewriting redirect `Location` headers before we return them to the browser. No longer needed since we no longer return third-party redirects to the browser.

conf/nginx/includes/direct_proxy.conf
```diff
@@ -23,18 +23,3 @@ rewrite ^/proxy/static/(.*)$ $1? break;
...
-
-# Prevent the third party from adding CORS controls
-proxy_hide_header "Access-Control-Allow-Origin";
-proxy_hide_header "Access-Control-Allow-Credentials";
-proxy_hide_header "Access-Control-Allow-Methods";
-proxy_hide_header "Access-Control-Allow-Headers";
-proxy_hide_header "Access-Control-Expose-Headers";
```
seanh, line 40: This got moved into a separate file, but appears to be unnecessary nonsense anyway (see other comment above)

conf/nginx/includes/errors.conf
```diff
@@ -0,0 +1,37 @@
...
+# error responses we don't want Via 3 to return those same 4xx or
+# 5xx status codes to the browser because this would mess up our
+# monitoring and alerting. For example it might trigger an alarm
+# that Via 3 is unhealthy because it's sending too many error
+# responses.
+# http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page
+error_page 404 410 = @proxy_not_found;
+error_page 420 429 = @proxy_too_many_requests;
+error_page 400 401 402 403 405 406 407 408 409 411 412 413 414 415 416 417 418 422 423 424 425 426 428 431 444 449 450 451 = @proxy_client_error;
+error_page 500 501 502 503 504 505 506 507 508 509 510 511 598 599 = @proxy_upstream_error;
```
seanh, line 37: This is the same code as we had before for making 3xx's, 4xx's and 5xx's go through the `@handle_redirect` and `@proxy_*` locations. Just moved into a snippet because it needs to be `include`'d twice now.

It's a shame that this code for directing the 3/4/5xx's to the `@foo` locations lives in this file, but the definitions of the `@foo` locations themselves live in another file. I'm not sure there's much that can be done about it though

The one new thing is the `recursive_error_pages on;` near the top.

conf/nginx/nginx.conf
```diff
@@ -28,26 +28,6 @@ http {
...
-    #
-    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
-    # http://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_
-    # http://nginx.org/en/docs/http/ngx_http_map_module.html#map
-    # http://nginx.org/en/docs/http/ngx_http_core_module.html#var_scheme
-    map $http_x_forwarded_proto $original_scheme {
-        "" $scheme;
-        default $http_x_forwarded_proto;
-    }
-
```
seanh, line 50: This sets the `$original_scheme` variable which was used to rewrite the `Location` headers of redirect responses but since we no longer rewrite those we no longer need this

conf/nginx/nginx.conf
```diff
@@ -105,13 +60,21 @@ http {
...
+            # http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_server_name
+            proxy_ssl_server_name on;
+
+            # Just using $upstream_http_location in the proxy_pass below
+            # doesn't work. We have to save it into this variable first.
+            set $saved_redirect_location '$upstream_http_location';
+
+            # Follow the redirect internally, now proxying to the URL given in
+            # the redirect's Location header.
+            proxy_pass $saved_redirect_location;
```
seanh, line 77: And this is how we follow redirects internally.

It's important to note that the `include includes/direct_proxy.conf;` has been deleted from the `@handle_redirect` so everything in `direct_proxy.conf` now applies to the _first_ request to the third-party server only and not to the subsequent requests when we follow redirects internally. This works in my testing so far but is it the right thing to do? Or should everything that applies to the original request also apply to the subsequent requests?

Here's the code that applies to the original request but not subsequent ones:

```nginx
# Replace the default url-decoded $uri with the $request_uri which is as
# we received it without url-decoding
rewrite ^ $request_uri;

# Redirect Google Drive requests to the API using our key
# Some weird things:
# * A rule that starts with http will _immediately_ issue a proxy_redirect
# * So we seemingly pointlessly capture (https) and replace it
# * This way the _result_ starts with https, but the _rule_ doesn't and we can carry on
# * NGINX will _merge_ queries you add, not replace them unless you put '?' on the end
rewrite ^/proxy/static/(https)://drive.google.com/uc\?id=(.*)&export=download$ $1://www.googleapis.com/drive/v3/files/$2?key=$google_api_key&alt=media? break;

# Remove our URL part off of the front
# The ? at the end means we strip off the args part (which we will add back on later)
# I have no idea why this works to be honest but it seems to
rewrite ^/proxy/static/(.*)$ $1? break;

# The proxy directly to the resulting URL
# Add back on the args which we stripped above so we don't get them twice
proxy_pass $uri$is_args$args;
```

seanh: For example if a page redirects to a Google Drive file, currently on this branch that will cause us to download that Google Drive file _without_ rewriting it to googleapis.com. We might want to apply that Google Drive rewriting to the `Location` URLs as well.

Apart from that, the rest of this looks fine to me

conf/nginx/nginx.conf
```diff
@@ -105,13 +60,21 @@ http {
...
-            proxy_hide_header "Cache-Control";
-            add_header "Cache-Control" "no-cache, no-store, must-revalidate";
+            # Pass the server name through when connecting to proxied HTTPS servers.
+            # This is needed to make proxying of some sites work.
+            # http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_server_name
+            proxy_ssl_server_name on;
+
+            # Just using $upstream_http_location in the proxy_pass below
+            # doesn't work. We have to save it into this variable first.
+            set $saved_redirect_location '$upstream_http_location';
```
jon-betts, line 73: ... why!?

jon-betts:
Testing results so far:

* Basic PDF - All good (http://localhost:9083/proxy/static/http://africau.edu/images/default/sample.pdf)
* Canvas PDF assignment - All good 
* Canvas file assignment - All good
* Canvas file with comma in the name - All good
* Canvas Google Drive assignment - All good
* Errors (http://localhost:9083/proxy/static/https://httpbin.org/status/501 etc)
   * 501 -> 409 - All good 
   * 500 -> 409 - All good 
   * 400 -> 400 - All good 
   * 401 -> 400 - All good 
   * 451 -> 400 - All good 
   * 404 -> 404 - All good 
   * 410 -> 404 - All good 
   * 420 -> 429 - All good

seanh:
### Testing

You need to run my test app for testing this: https://github.com/hypothesis/pdf_redriects (I think the typo in the name is probably appropriate). Then:

- [ ] The four localhost assignments (HTML, PDF, Canvas Files, Google Drive) in the [Sections Enabled](https://hypothesis.instructure.com/courses/125) test course should work

- [ ] The **localhost (make devdata) Redirects Test Assignment** assignments in the same course should all work as expected (some of these are expected to crash)

You can also test getting URLs directly in a terminal with `curl` or whatever:

- [x] http://localhost:9083/proxy/static/http://172.17.0.1:7077/pdf should return a PDF
- [x] http://localhost:9083/proxy/static/http://172.17.0.1:7077/absolute_redirect should return a PDF immediately, without redirecting you
- [x] http://localhost:9083/proxy/static/http://172.17.0.1:7077/relative_redirect should return a PDF immediately, without redirecting you
- [x] http://localhost:9083/proxy/static/http://172.17.0.1:7077/redirect_n_times/9/1 should return a PDF immediately, without redirecting you
- [x] http://localhost:9083/proxy/static/http://172.17.0.1:7077/redirect_n_times/11/1 should 500 because the chain of redirects is too long
- [x] http://localhost:9083/proxy/static/http://172.17.0.1:7077/crash should change the 500 Internal Server Error from the third-party server to a different error
- [x] http://localhost:9083/proxy/static/http://172.17.0.1:7077/redirect_to_crash should follow the redirect internally and then change the 500 Internal Server Error from the third-party server to a different error

conf/nginx/nginx.conf
jon-betts: Hmm, I don't know. I think we're probably fine catching the majority case for now (even though we've recently learned we aren't handling it well).

How many redirects are going to do this? Not many I'd bet.

conf/nginx/nginx.conf
```diff
@@ -105,13 +60,21 @@ http {
         }
```
jon-betts, line 61: Now the errors have been shipped out of this file, it's not immediately clear where `@handle_redirect` comes from.

It might be nice to add:

```
# @handle_redirect and the other targets defined here are used in `include/error.conf`
```

Or something...

jon-betts (approved):
This diff looks horrendous, but it's mostly stuff moving about. The things that have been added are actually a lot easier to understand than that which has been removed. The organisation into `includes/` also makes everything clearer (with perhaps the exception of where `@handle_redirects` and co come from)

I'd suggest adding a comment to give people a hint. In 6 months time when we next look at this I'm _going_ to forget and waste at least some time Googling to see if this is an NGINX built in thing.

conf/nginx/nginx.conf
seanh: Probably not many. I'm just gonna leave this as it is then

conf/nginx/nginx.conf
seanh: I think it might be because calling `proxy_pass` actually changes the upstream location? Dunno

conf/nginx/nginx.conf
seanh: Hmm, yeah, we could maybe have `include/error.conf` and `include/error_handlers.conf`. I'll give it a try

seanh: https://github.com/hypothesis/checkmate/issues/187

conf/nginx/nginx.conf
seanh: https://github.com/hypothesis/via3/issues/355

PR #352, opened 2021-02-05
Enable checkmate checks on route_by_content

marcospri:
Won't have any effect until `CHECKMATE_ENABLED` is enabled again.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import httpretty
 import pytest
 from h_matchers import Any
 
@@ -22,3 +23,19 @@ def assert_cache_control(headers, cache_parts):
     assert (
         headers["Cache-Control"].split(", ") == Any.list.containing(cache_parts).only()
     )
+
+
+@pytest.fixture(autouse=True)
+def httpretty_():
+    """Monkey-patch Python's socket core module to mock all HTTP responses.
+
+    We never want real HTTP requests to be sent by the tests so replace them
+    all with mock responses. This handles requests sent using the standard
+    urllib2 library and the third-party httplib2 and requests libraries.
+    """
+    httpretty.enable(allow_net_connect=False)
+
+    yield
+
+    httpretty.disable()
+    httpretty.reset()
--- a/tests/functional/api/views/route_by_content_test.py
+++ b/tests/functional/api/views/route_by_content_test.py
@@ -1,0 +1,64 @@
+from urllib.parse import quote_plus
+
+import httpretty
+import pytest
+from h_matchers import Any
+
+from tests.conftest import assert_cache_control
+
+
+class TestRouteByContent:
+    DEFAULT_OPTIONS = {
+        "via.client.openSidebar": "1",
+        "via.client.requestConfigFromFrame.origin": "http://localhost",
+        "via.client.requestConfigFromFrame.ancestorLevel": "2",
+        "via.external_link_mode": "new-tab",
+    }
+
+    @pytest.mark.usefixtures("html_response")
+    def test_proxy_html(self, test_app):
+        target_url = "http://example.com"
+
+        response = test_app.get(f"/route?url={target_url}")
+
+        assert response.status_code == 302
+        query = dict(self.DEFAULT_OPTIONS)
+        query["via.sec"] = Any.string()
+        assert response.location == Any.url.matching(
+            f"https://viahtml3.hypothes.is/proxy/{target_url}/"
+        ).with_query(query)
+
+    @pytest.mark.usefixtures("pdf_response")
+    def test_proxy_pdf(self, test_app):
+        target_url = "http://example.com"
+
+        response = test_app.get(f"/route?url={target_url}")
+
+        assert response.status_code == 302
+        query = dict(self.DEFAULT_OPTIONS)
+        query["via.sec"] = Any.string()
+        query["url"] = target_url
+        assert response.location == Any.url.matching(
+            f"http://localhost/pdf?url={quote_plus(target_url)}"
+        ).with_query(query)
+        assert_cache_control(
+            response.headers, ["public", "max-age=300", "stale-while-revalidate=86400"]
+        )
+
+    @pytest.fixture
+    def html_response(self):
+        httpretty.register_uri(
+            httpretty.GET,
+            "http://example.com",
+            status=204,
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -19,3 +23,36 @@ def assert_cache_control(headers, cache_parts):
...
+
+
+@pytest.fixture
+def pdf_response(get_url_details):
+    get_url_details.return_value = ("application/pdf", 200)
+
+
+@pytest.fixture
+def html_response(get_url_details):
+    get_url_details.return_value = ("application/html", 200)
```
jon-betts: I'm not sure re-use here is the right play. It makes the relevant test very short, but it's also quite mysterious. We have a fixture we don't use or declare.

This might be a good example where repeating yourself locally is actually better? I'm not sure.

I'd say I just about side with you on this one. There's enough stuff here and we use it twice...

via/views/route_by_content/view.py
```diff
@@ -4,10 +4,14 @@
 from pyramid import view
 
 from via.get_url import get_url_details
+from via.views.blocker import checkmate_block
```
jon-betts: This confused me on first read. On the whole, unless otherwise stated most things in `product.views.` are a view. The exception so far being `exceptions.py` which is a mixed bag. Either it's exceptions related to the views (which I prefer in the parent, but opinion varies), or pyramid exception views.

We are branching out with directories for views when they contain lots of local classes. That wouldn't work here as it's intended to be shared.

I wonder if this would be better as `_decorators.py `or something so you have from `via.views._decorators import checkmate_block`, which heavily suggests it's not a view itself?

tests/functional/api/view_route_by_content_test.py
```diff
@@ -0,0 +1,26 @@
+from urllib.parse import quote_plus
+
+import pytest
+
+from tests.conftest import assert_cache_control
+
+
+class TestRouteByContent:
+    def test_proxy_html(self, test_app, html_response):
```
jon-betts:
```suggestion
    @pytest.mark.usefixtures("html_response")
    def test_proxy_html(self, test_app):
```

We normally use mark when we have fixtures that setup a situation rather than return a value. In fact, we intentionally _don't_ return a value in fixtures where we are trying to setup a situation, to help you tell the fact by looking at the fixture definition (no return value = pure side effect fixture)

This helps in a few ways:

 * It's a bit baffling if you look at the code and think "This arg isn't used, I'll remove it" then the test breaks
 * The value is just None, so if you debug, it's not clear why this happened

tests/functional/api/view_route_by_content_test.py
```diff
@@ -0,0 +1,26 @@
...
+        target_url = "http://example.com"
+
+        response = test_app.get(f"/route?url={target_url}")
+
+        assert response.status_code == 302
+        assert response.location == f"https://viahtml3.hypothes.is/proxy/{target_url}"
+
+    def test_proxy_pdf(self, test_app, pdf_response):
```
jon-betts:
```suggestion
    @pytest.mark.usefixtures("pdf_response")
    def test_proxy_pdf(self, test_app):
```

Ditto

jon-betts (changes requested):
This all looks pretty much fine, but I think we should follow the style elsewhere for fixtures which setup a situation, and use `@pytest.mark.usefixture`

It looks like it's gone out in a different PR, but the view decorator here doesn't seem follow our standard pattern for naming things in the views. I'd expect this to be a view, rather than contain tools for views. Not sure if this is something we want to fix here or not. I'm surprised this didn't get any comment at the time, but maybe I was asleep at the wheel.

via/views/route_by_content/view.py
marcospri: Changed it to `decorators.py` (which we could break further if it grows too much)

tests/conftest.py
marcospri: Not sure either, I think the name it's relatively self explanatory so having them in conftest it's not too confusing.

tests/unit/via/views/blocker_test.py
```diff
@@ -5,7 +5,7 @@
 from checkmatelib import CheckmateException
 from pyramid.response import Response
 
-from via.views.blocker import checkmate_block
```
seanh: Does this file need to be renamed to `decorators_test.py` as well?

tests/conftest.py
seanh: It looks like you moved these three fixtures into `tests/conftest.py` because your functional test (in `view_route_by_content_test.py`) needs to use them. But I think a functional test should probably work by faking the third-party server's response using `httpretty`, rather than by patching part of the code. Having functional tests use patch fixtures seems too brittle.

Then you won't need to move `patch` and `autopatcher`, those can go back into `tests/unit/conftest.py`

tests/functional/api/view_route_by_content_test.py
```diff
@@ -0,0 +1,28 @@
...
+    def test_proxy_pdf(self, test_app):
+        target_url = "http://example.com"
+
+        response = test_app.get(f"/route?url={target_url}")
+
+        assert response.status_code == 302
+        assert response.location == f"http://localhost/pdf?url={quote_plus(target_url)}"
+        assert_cache_control(
+            response.headers, ["public", "max-age=300", "stale-while-revalidate=86400"]
+        )
```
seanh: I think you should move the `html_response` and `pdf_response` fixtures into this test class, and they should work by setting up `httpretty` so that requests to `http://example.com` respond with the required content-type headers

seanh (changes requested):

tests/functional/api/view_route_by_content_test.py
jon-betts: FYI, we stopped needing them in the place they originally came from in https://github.com/hypothesis/via3/pull/369, so this actually won't even be duplication.

tests/functional/api/view_route_by_content_test.py
```diff
@@ -0,0 +1,64 @@
+from urllib.parse import quote_plus
```
seanh: I wonder if the file should by `views/route_by_content_test.py` or just `route_by_content_test.py`, rather than `view_route_by_content_test.py`

seanh (approved):

PR #353, opened 2021-02-05
Require nginx links to be signed

seanh:
Depends on: [Follow redirects internally](https://github.com/hypothesis/via3/pull/349)

This PR:

1. Requires requests to `/proxy/static/` to be signed
2. Changes the `/pdf` route to sign the `/proxy/static/` URLs that it generates

This doesn't actually secure anything since the `/pdf` endpoint itself is still open and will generate a signed `/proxy/static` URL to any URL that you want.

## Testing

You need to run my test app for testing this: https://github.com/hypothesis/pdf_redriects (I think the typo in the name is probably appropriate). Then:

- [x] The localhost assignments (HTML, PDF, Canvas Files, Google Drive) in the [Sections Enabled](https://hypothesis.instructure.com/courses/125) test course should work

- [x] The **localhost (make devdata) Redirects Test Assignment** assignments in the same course should all work as expected (some of these are expected to crash)

### DIFF
--- a/conf/nginx/includes/direct_proxy.conf
+++ b/conf/nginx/includes/direct_proxy.conf
@@ -3,23 +3,61 @@
 # http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_server_name
 proxy_ssl_server_name on;
 
-# Replace the default url-decoded $uri with the $request_uri which is as
-# we received it without url-decoding
-rewrite ^ $request_uri;
+# Rewrite drive.google.com URLs to googleapis.com ones that use our API key.
+#
+# This is because Google gives us a much higher rate limit / quota when we use
+# our API key.
+#
+# Notes:
+#
+# * If a `rewrite`'s replacement string starts with "http://", "https://" or
+#   "$scheme" nginx immediately stops processing after this rewrite rule and
+#   returns a redirect to the client. We don't want this to happen so we
+#   capture the "https" in the `rewrite`'s regex so that we can use it as "$1"
+#   in the replacement string!
+#
+# * If a `rewrite`'s replacement string contains query params nginx also
+#   appends the _previous_ query params (from the pre-rewrite $uri) after them.
+#   We don't want this. The `?` at the end of the replacement string disables it.
+#
+# * The `break` at the end causes nginx to skip any subsequent rewrite module
+#   directives (e.g. other `rewrite`'s). If the request matches this rewrite
+#   (i.e. it's an https://drive.google.com request) then this will be the
+#   *last* rewrite that gets processed for the request.
+#
+#   Subsequent non-rewrite module directives will still get processed.
+#
+#   The `break` also causes nginx to stop processing at the end of this
+#   `location { ... }` and send the `proxy_pass`'d response back to the
+#   browser, rather than re-starting its processing using the rewritten $uri to
+#   search for a matching `location { ... }` again.
+#
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#break
+rewrite ^(https)://drive.google.com/uc\?id=(.*)&export=download$ $1://www.googleapis.com/drive/v3/files/$2?key=$google_api_key&alt=media? break;
 
-# Redirect Google Drive requests to the API using our key
-# Some weird things:
-# * A rule that starts with http will _immediately_ issue a proxy_redirect
-# * So we seemingly pointlessly capture (https) and replace it
-# * This way the _result_ starts with https, but the _rule_ doesn't and we can carry on
-# * NGINX will _merge_ queries you add, not replace them unless you put '?' on the end
-rewrite ^/proxy/static/(https)://drive.google.com/uc\?id=(.*)&export=download$ $1://www.googleapis.com/drive/v3/files/$2?key=$google_api_key&alt=media? break;
+# Remove the query params from the URI so we can add them back again later.
+#
+# If the request had matched the drive.google.com rewrite above its query
+# params would have been removed. If the request does *not* match the
+# drive.google.com rewrite then it will match this rewrite instead. We need to
+# strip the query params from non-drive.google.com requests as well so that we
+# can unconditionally re-add them in the proxy_pass below.
+#
+# The `break` is necessary to cause nginx to stop processing at the end of this
+# `location { ... }` and send the `proxy_pass`'d response back to the browser,
+# rather than re-starting its processing using the rewritten $uri to search for
+# a matching `location { ... }` again.
+#
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#break
+rewrite ^(.*)$ $1? break;
 
-# Remove our URL part off of the front
-# The ? at the end means we strip off the args part (which we will add back on later)
-# I have no idea why this works to be honest but it seems to
-rewrite ^/proxy/static/(.*)$ $1? break;
-
-# The proxy directly to the resulting URL
-# Add back on the args which we stripped above so we don't get them twice
+# Proxy to the resulting $uri (after all the `rewrite`s above).
+#
+# Query params have been stripped from $uri so re-add them with $is_args$args.
+#
+# http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass
+# http://nginx.org/en/docs/http/ngx_http_core_module.html#var_is_args
+# http://nginx.org/en/docs/http/ngx_http_core_module.html#var_args
 proxy_pass $uri$is_args$args;
--- a/conf/nginx/includes/secure_links.conf
+++ b/conf/nginx/includes/secure_links.conf
@@ -1,0 +1,47 @@
+# Require the request to be signed with an MD5 and expiration in the URL,
+# responding with an error if the request isn't signed.
+#
+# This uses nginx's built-in secure link module:
+#
+# https://www.nginx.com/blog/securing-urls-secure-link-module-nginx-plus/
+# http://nginx.org/en/docs/http/ngx_http_secure_link_module.html
+
+# Give nginx's secure link module the checksum ($sec) and expiration time
+# ($exp) to verify.
+#
+# $sec and $exp are extracted from the requested URI by the `location` regex in nginx.conf.
+#
+# http://nginx.org/en/docs/http/ngx_http_secure_link_module.html#secure_link
+secure_link $sec,$exp;
+
+# Replace the $uri variable (which is URL-decoded) with the raw version from $request_uri.
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
+rewrite ^ $request_uri;
+
+# Remove the /proxy/static/<SEC>/<EXP> prefix from the $uri, we don't need it anymore.
+# (And we do need easy access to $uri without the prefix because
+# that's the third-party URL that we need to proxy.)
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
+rewrite ^/proxy/static/[^/]+/[^/]+/(.*)$ $1;
+
+# The string that we expect the MD5 digest in the URL ($sec) to be a hash of:
+#
+#   $secure_link_expires is the timestamp in the URL (the same as $exp)
+#   $uri is the URL to be proxied
+#   $nginx_secure_link_secret comes from the NGINX_SECURE_LINK_SECRET environment variable
+#
+# http://nginx.org/en/docs/http/ngx_http_secure_link_module.html#secure_link_md5
+secure_link_md5 "/proxy/static/$secure_link_expires/$uri $nginx_secure_link_secret";
+
+# $secure_link will be set to:
+#
+#   "" if the MD5 digest in the URL ($sec) is missing or wrong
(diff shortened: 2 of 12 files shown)


conf/nginx/includes/direct_proxy.conf
```diff
@@ -3,23 +3,61 @@
...
+#   Subsequent non-rewrite module directives will still get processed.
+#
+#   The `break` also causes nginx to stop processing at the end of this
+#   `location { ... }` and send the `proxy_pass`'d response back to the
+#   browser, rather than re-starting its processing using the rewritten $uri to
+#   search for a matching `location { ... }` again.
+#
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#break
+rewrite ^(https)://drive.google.com/uc\?id=(.*)&export=download$ $1://www.googleapis.com/drive/v3/files/$2?key=$google_api_key&alt=media? break;
```
seanh, line 37: There's quite a lot going on in this rewrite so I added a massive comment to it :man_shrugging:

conf/nginx/includes/secure_links.conf
```diff
@@ -0,0 +1,44 @@
...
+# https://www.nginx.com/blog/securing-urls-secure-link-module-nginx-plus/
+# http://nginx.org/en/docs/http/ngx_http_secure_link_module.html
+
+# $secure_link will be set to:
+#
+#   "" if the MD5 digest in the URL ($sec) is missing or wrong
+#   "0" if the digest is valid but the timestamp in the URL ($exp) has expired
+#   "1" if both the digest and the timestamp are valid
+#
+# $sec and $exp are extracted from the requested URI by the `location` regex above.
```
jon-betts: What regex above?

conf/nginx/includes/secure_links.conf
```diff
@@ -0,0 +1,44 @@
...
+#
+# http://nginx.org/en/docs/http/ngx_http_secure_link_module.html#secure_link
+secure_link $sec,$exp;
```
jon-betts, line 15: It looks like this whole comment applies to this line, but I don't think it does. What does this line actually do?

conf/nginx/includes/secure_links.conf
```diff
@@ -0,0 +1,44 @@
...
+# This uses nginx's built-in secure link module:
+#
+# https://www.nginx.com/blog/securing-urls-secure-link-module-nginx-plus/
+# http://nginx.org/en/docs/http/ngx_http_secure_link_module.html
+
+# $secure_link will be set to:
+#
+#   "" if the MD5 digest in the URL ($sec) is missing or wrong
+#   "0" if the digest is valid but the timestamp in the URL ($exp) has expired
+#   "1" if both the digest and the timestamp are valid
```
jon-betts: Would this comment be more appropriate at the bottom where the values are being set?

conf/nginx/includes/direct_proxy.conf
```diff
@@ -3,23 +3,61 @@
...
+#   directives (e.g. other `rewrite`'s). If the request matches this rewrite
+#   (i.e. it's an https://drive.google.com request) then this will be the
+#   *last* rewrite that gets processed for the request.
+#
+#   Subsequent non-rewrite module directives will still get processed.
+#
+#   The `break` also causes nginx to stop processing at the end of this
+#   `location { ... }` and send the `proxy_pass`'d response back to the
+#   browser, rather than re-starting its processing using the rewritten $uri to
+#   search for a matching `location { ... }` again.
```
jon-betts, line 33: :+1: Super useful summary of absolutely baffling config.

conf/nginx/includes/direct_proxy.conf
```diff
@@ -3,23 +3,61 @@
...
+# can unconditionally re-add them in the proxy_pass below.
+#
+# The `break` is necessary to cause nginx to stop processing at the end of this
+# `location { ... }` and send the `proxy_pass`'d response back to the browser,
+# rather than re-starting its processing using the rewritten $uri to search for
+# a matching `location { ... }` again.
+#
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite
+# http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#break
+rewrite ^(.*)$ $1? break;
```
jon-betts, line 54: NGINX seems to have a lot of really unhelpful default behaviors. How about a version of `rewrite` that isn't `rewrite_and_go_back_or_do_a_bunch_of_other_non_configurable_things`?

docker-compose.yml
```diff
@@ -6,14 +6,16 @@ services:
...
       - ./conf/nginx/includes/cors.conf:/etc/nginx/includes/cors.conf:ro
       - ./conf/nginx/includes/direct_proxy.conf:/etc/nginx/includes/direct_proxy.conf:ro
       - ./conf/nginx/includes/errors.conf:/etc/nginx/includes/errors.conf:ro
       - ./conf/nginx/includes/robots.conf:/etc/nginx/includes/robots.conf:ro
+      - ./conf/nginx/includes/secure_links.conf:/etc/nginx/includes/secure_links.conf:ro
       - ./conf/nginx/includes.dev/app_upstream.conf:/etc/nginx/includes/app_upstream.conf:ro
       - ./conf/nginx/dev_host_bridge.sh:/etc/nginx/dev_host_bridge.sh:ro
       - ./conf/nginx/envsubst.conf.template:/var/lib/hypothesis/nginx_envsubst.conf.template:ro
-    command: /bin/sh -c "/etc/nginx/dev_host_bridge.sh && envsubst '$${GOOGLE_API_KEY}' < /var/lib/hypothesis/nginx_envsubst.conf.template > /var/lib/hypothesis/nginx_envsubst.conf && exec nginx"
+    command: /bin/sh -c "/etc/nginx/dev_host_bridge.sh && envsubst '$${GOOGLE_API_KEY}:$${NGINX_SECURE_LINK_SECRET}' < /var/lib/hypothesis/nginx_envsubst.conf.template > /var/lib/hypothesis/nginx_envsubst.conf && exec nginx"
```
jon-betts, line 20: I hate this so much.... but there doesn't seem to be a better way

tox.ini
```diff
@@ -27,6 +27,7 @@ setenv =
     dev: NEW_RELIC_ENVIRONMENT = {env:NEW_RELIC_ENVIRONMENT:dev}
     dev: SENTRY_ENVIRONMENT = {env:SENTRY_ENVIRONMENT:dev}
     dev: CHECKMATE_URL = http://localhost:9099
+    dev: NGINX_SECURE_LINK_SECRET = seekrit
```
jon-betts: Our trad secret is "not_a_secret"

via/views/view_pdf.py
```diff
@@ -34,6 +38,31 @@ def view_pdf(context, request):
...
+    hash_.update(f"/proxy/static/{exp}/{url} {secret}".encode("utf-8"))
+    sec = hash_.digest()
+    sec = b64encode(sec)
+    sec = sec.replace(b"+", b"-")
+    sec = sec.replace(b"/", b"_")
+    sec = sec.replace(b"=", b"")
+    sec = sec.decode()
+
+    # Construct the URL.
+    return _string_literal(f"{nginx_server}/proxy/static/{sec}/{exp}/{url}")
```
jon-betts, line 73: Does this need to be shipped out to `h-vialib` like we said? If LMS is coming in it's not going directly to the proxy point is it? Just the view. I think this might be fine here.

jon-betts (commented):
Some comments after a quick eyeball through. I've not tested it yet, as my machine is a bit bust and needs a reboot

conf/nginx/includes/secure_links.conf
seanh: Ugh. it's in the `location <regex> { ... }` but that's in another file that `include`'s this one

conf/nginx/includes/secure_links.conf
seanh: It tells nginx what MD5 and expiry time has been given in the request. I think of it kind of like a function call that takes two arguments: the MD5 and the expiry time. It's up to us to decide how to extract the MD5 and exp from the request, and then we pass them to nginx here

conf/nginx/includes/direct_proxy.conf
seanh: Yeah. I tried refactoring the code to be less confusing, but it's hard to get it to work in a less baffling way, and changing it is risky

via/views/view_pdf.py
seanh: I think there could well be some shared code between LMS which needs to generate signed links for Via 3's Python endpoints, and Via 3 which needs to verify those signed links and also needs to generate signed links for its own NGINX endpoint. We could leave this here at first though, and pull it out into h-vialib when we actually have some duplicate code to de-duplicate

jon-betts:
Testing results:

* [x] http://localhost:9083/route?url=http://172.17.0.1:7077/pdf should return a PDF
* [x] http://localhost:9083/route?url=http://172.17.0.1:7077/absolute_redirect should return a PDF immediately, without redirecting you
* [x] http://localhost:9083/route?url=http://172.17.0.1:7077/relative_redirect should return a PDF immediately, without redirecting you
* [x] http://localhost:9083/route?url=http://172.17.0.1:7077/redirect_n_times/9/1 should return a PDF immediately, without redirecting you
* [x] http://localhost:9083/route?url=http://172.17.0.1:7077/redirect_n_times/11/1 should 500 because the chain of redirects is too long
* **X** http://localhost:9083/route?url=http://172.17.0.1:7077/crash should change the 500 Internal Server Error from the third-party server to a different error - This doesn't really work as it gets directed off to ViaHTML before this can happen
* **X** http://localhost:9083/route?url=http://172.17.0.1:7077/redirect_to_crash should follow the redirect internally and then change the 500 Internal Server Error from the third-party server to a different error - Ditto with above

jon-betts:
I've run this through the main stuff in Canvas and it all looks ok.

conf/nginx/includes/secure_links.conf
seanh: Fixed

conf/nginx/includes/secure_links.conf
seanh: Yeah moved this further down

tests/unit/via/views/view_pdf_test.py
```diff
@@ -23,6 +25,49 @@ def test_it_passes_through_static_config(self, call_view_pdf, pyramid_settings):
...
+        datetime.datetime.utcnow.return_value = now
+
+        response = call_view_pdf("https://example.com/foo/bar.pdf?q=s")
+
+        signed_url = json.loads(response["pdf_url"])
+        signed_url_parts = signed_url.split("/")
+        signature = signed_url_parts[5]
+        expiry = signed_url_parts[6]
+        assert signature == expected_signature
+        assert expiry == expected_expiry
```
seanh: Once we've extracted the timestamp-generation and md5-generation out into separate functions, we'll be able to split this test up

via/views/view_pdf.py
```diff
@@ -34,6 +42,29 @@ def view_pdf(context, request):
...
```
seanh: I've just shoved this all into one helper function now for speed, but in follow-up PRs I think we could split this into a few separately-tested functions. One for generating the timestamp, another for generating the hash, another for generating the URL

tests/unit/via/views/view_pdf_test.py
jon-betts: ... pretty much my feedback yeah

via/views/view_pdf.py
```diff
@@ -34,6 +42,29 @@ def view_pdf(context, request):
...
 
 
+def _pdf_url(url, nginx_server, secret):
+    # Compute the expiry time to put into the URL.
+    now = datetime.datetime.utcnow()
+    exp = (
+        datetime.datetime(now.year, now.month, now.day, now.hour, tzinfo=timezone.utc)
+        + datetime.timedelta(hours=2)
+    ).timestamp()
+    exp = int(exp)
```
jon-betts: If we can get this sorted in https://github.com/hypothesis/h-vialib/pull/4/ 

I think this could be replaced with:

```
from h_vialib.secure import quantized_expiry

exp = int(quantized_expiry(max_age=60 * 60 * 2))
```

via/views/view_pdf.py
```diff
@@ -34,6 +42,29 @@ def view_pdf(context, request):
...
+
+    # Compute the digest to put into the URL.
+    hash_ = hashlib.md5()
+    hash_.update(f"/proxy/static/{exp}/{url} {secret}".encode("utf-8"))
+    sec = hash_.digest()
+    sec = b64encode(sec)
+    sec = sec.replace(b"+", b"-")
+    sec = sec.replace(b"/", b"_")
+    sec = sec.replace(b"=", b"")
```
jon-betts, line 68: This is a bit random. Are there some NGINX docs saying you should do this we could link to?

jon-betts (commented):
This seems fine, but the tests are really tightly bound to the hashing function. I'd probably make that a top level component and test it separately. We can also steal some stuff from my `h-vialib` PR if that's ok.

tox.ini
seanh: Done

via/views/view_pdf.py
seanh: Added some comments

via/views/view_pdf.py
seanh: Done

PR #357, opened 2021-02-08
Check for signed URLs in /pdf and /route

marcospri:
To quickly check locally:

- LMS in master branch
- via3 in this branch

`https://hypothesis.instructure.com/courses/125/assignments/874` should load

Change VIA_SECRET in either LMS or via3 to no longer match each other (and `make dev` again)

`https://hypothesis.instructure.com/courses/125/assignments/874` should return a 401

### DIFF
--- a/via/app.py
+++ b/via/app.py
@@ -9,7 +9,13 @@ from whitenoise import WhiteNoise
 from via.cache_buster import PathCacheBuster
 from via.sentry_filters import SENTRY_FILTERS
 
-REQUIRED_PARAMS = ["client_embed_url", "nginx_server", "via_html_url", "checkmate_url"]
+REQUIRED_PARAMS = [
+    "client_embed_url",
+    "nginx_server",
+    "via_html_url",
+    "checkmate_url",
+    "via_secret",
+]
 
 
 def load_settings(settings):
@@ -32,6 +38,7 @@ def load_settings(settings):
     settings["h_pyramid_sentry.filters"] = SENTRY_FILTERS
 
     settings["checkmate_enabled"] = asbool(os.environ.get("CHECKMATE_ENABLED"))
+    settings["signed_urls_required"] = asbool(os.environ.get("SIGNED_URLS_REQUIRED"))
     settings["nginx_secure_link_secret"] = os.environ["NGINX_SECURE_LINK_SECRET"]
 
     return settings
--- a/via/views/decorators.py
+++ b/via/views/decorators.py
@@ -2,7 +2,9 @@
 import logging
 
 from checkmatelib import CheckmateClient, CheckmateException
-from pyramid.httpexceptions import HTTPTemporaryRedirect
+from h_vialib.exceptions import TokenException
+from h_vialib.secure.url import ViaSecureURL
+from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPUnauthorized
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


via/app.py
```diff
@@ -32,6 +38,7 @@ def load_settings(settings):
     settings["h_pyramid_sentry.filters"] = SENTRY_FILTERS
 
     settings["checkmate_enabled"] = asbool(os.environ.get("CHECKMATE_ENABLED"))
+    settings["signed_urls_enabled"] = asbool(os.environ.get("SIGNED_URLS_ENABLED"))
```
seanh: Maybe `"signed_urls_required"`?

via/views/blocker.py
```diff
@@ -35,3 +37,33 @@ def view_wrapper(context, request):
...
+        url_token = request.params[url_param]
+        signed_url = urlparse(request.url)
+        options = dict(request.params)
+        del options[url_param]
+        signed_url = signed_url._replace(query=urlencode(options)).geturl()
+
+        try:
+            secure_token.verify(request.params["sec"], signed_url)
+        except (MissingToken, InvalidToken):
+            return HTTPUnauthorized()
```
jon-betts:
```suggestion
def secure_url_token(view, url_param="via.sec"):
    def view_wrapper(context, request):
        if not request.registry.settings["signed_urls_enabled"]:
            return view(context, request)

        secure_token = ViaSecureURLToken(request.registry.settings["via_secret"])
        
        try:
            secure_token.verify(request.params.get(url_param), request.url)
        except TokenException:
            return HTTPUnauthorized()
            
        return None
```

Quite a few things here:

* `ViaSecureURLToken` already raises `MissingToken` if passed None, as we don't need to check for that
* We don't need to clean the URL as:
    *  ViaSecureURLToken already does
    * There's nothing in the URL in the first place to clean. The URL param is totally separate from the token
* We use `via.*` for via configuration parameters. It seems like it might simplify things to do this in general
* We have a unified base class for the exceptions now!

jon-betts (commented):
More unhelpful comments!

via/views/blocker.py
jon-betts: I'm not sure about the names of the decorators here. "secure_url_token" reads as a noun to me rather than an action. "has_secure_url_token" or "with_url_signature" or something?

via/views/decorators.py
```diff
@@ -35,3 +37,27 @@ def view_wrapper(context, request):
...
+    """Intended to be used as a decorator for pyramid views.
+
+    Checks the validity of URL signature present in the `url_param` query parameter.
+
+    :param url_param Name of the query param containing the signature
+    """
+
+    def view_wrapper(context, request):
+        signature = request.params.get(signature_param)
+        if not signature and not request.registry.settings["signed_urls_enabled"]:
```
marcospri: This leaves it as "while `disabled` it's still mandatory if sent" If enabled it will be mandatory regardless.

We could change it back to plain disabled/enabled if we want to be more conservative.

via/views/decorators.py
```diff
@@ -35,3 +37,27 @@ def view_wrapper(context, request):
...
+            return view(context, request)
+
+        secure_token = ViaSecureURL(request.registry.settings["via_secret"])
+        try:
+            secure_token.verify(request.url)
+        except TokenException:
+            return HTTPUnauthorized()
```
marcospri, line 54: Not sure if adding some logging here would be helpful if there are problems? I thought that this in particular might generate a lot of noise.

via/views/decorators.py
```diff
@@ -35,3 +37,27 @@ def view_wrapper(context, request):
...
     return view_wrapper
+
+
+def has_secure_url_token(view, signature_param="via.sec"):
+    """Intended to be used as a decorator for pyramid views.
+
+    Checks the validity of URL signature present in the `url_param` query parameter.
+
+    :param url_param Name of the query param containing the signature
+    """
```
seanh:
```suggestion
    """Require the request to have a valid signature in the "via.sec" query param."""
```

Why have the `signature_param` argument? I don't think the view config can actually use it

via/views/decorators.py
seanh: I think we'll deploy this and get it working once, and then it'll just work forever. So no need for logging

seanh (approved):
This looks simple enough to me.

I haven't tested it yet, I'll give it some testing locally before we deploy it.

I think the URL signing is actually an authentication thing, so the "right" way to implement this is in the authentication policy and with a `permission="signed"` in each view config, rather than with a view decorator. But this works so I think it's good enough for now

seanh:
Gave it a quick test locally and it seems to work :+1:

via/views/decorators.py
jon-betts: It's not something you have a choice about any more

PR #362, opened 2021-02-08
Fix the NGINX_SECURE_LINK_SECRET envvar in Docker

seanh:


jon-betts (approved):
Looks the same as the dockerfile command, so hopefully this is good?

PR #369, opened 2021-02-10
Use ViaClient to sign requests to ViaHTML

jon-betts:
We should wait on: https://github.com/hypothesis/h-vialib/pull/15 for this

## Testing notes

For this to work locally you need:

 * ViaHTML on: https://github.com/hypothesis/viahtml/pull/47

## Review notes

* This includes the ViaClient as a very thin pyramid service
* In route by content, we now use this to format all the URLs (including those which redirect to ourselves)
* This is a little indirect, but means that we are dog-fooding our client, and the client cannot diverge from Via 3
* It also means it's much simpler to handle, as all the logic of bodging together signed URLs is in the client
* This _also_ means we can't diverge from the signing method there either
* This eliminated quite a bit of code and tests


marcospri (dismissed):

tests/unit/via/app_test.py
```diff
@@ -28,11 +26,6 @@ def test_app(configurator, pyramid, os, pyramid_settings):
     )
 
     pyramid.config.Configurator.assert_called_once_with(settings=expected_settings)
-    assert configurator.include.call_args_list == [
-        mock.call("pyramid_jinja2"),
-        mock.call("via.views"),
-        mock.call("h_pyramid_sentry"),
-    ]
```
jon-betts, line 35: I think we're moving to just not doing this

via/services/via_client.py
```diff
@@ -0,0 +1,22 @@
+"""A thin wrapper around ViaClient to make it a service."""
+from h_vialib import ViaClient
+
+
+class ViaClientService(ViaClient):
+    """Place holder for a service, in case we want to diverge."""
```
jon-betts, line 6: Not sure if this is good or dumb. We could use the actual real class as the service but:

* It leads to a little confusion about where it's coming from
* Makes our services file look a bit dumb
* We can't diverge from the interface

via/services/via_client.py
```diff
@@ -0,0 +1,22 @@
...
+def factory(_context, request):
+    """Create a ViaClientService object from the request."""
+
+    settings = request.registry.settings
+
+    return ViaClientService(
+        # Where we are coming from
+        host_url=request.host_url,
+        # Where we are going for general / HTML specifically
+        service_url=request.host_url,
```
jon-betts, line 18: If we could work this out at app start time, this doesn't have to be a factory, it can be a singleton

seanh (approved):

PR #380, opened 2021-02-18
Return json on /_status endpoint to align with other services

### LINKED ISSUE #367: Via3 - Improve /_status endpoint
A small request to update the `/_status` endpoint for Via3.

If we can implement the method we have taken with ViaHTML [status.py](https://github.com/hypothesis/viahtml/blob/master/viahtml/views/status.py) then I can write a NewRelic synthetic monitor to load the page and check the response contains:
```
{"status": "okay"}
``` 

Can you let me know once this is complete and I will perform the necessary configuration in NewRelic.

marcospri:
Return json on /_status

```
curl 'http://localhost:9082/_status
{"status": "okay"}%
```
```
curl -I 'http://localhost:9082/_status
HTTP/1.1 200 OK
Server: gunicorn/20.0.4
Date: Thu, 18 Feb 2021 09:08:16 GMT
Connection: close
Content-Length: 18
```


tests/functional/api/view_monitoring_test.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from tests.conftest import assert_cache_control
+
+
```
seanh:
```suggestion
```

That's weird: both of these imports are unused, but `make lint` isn't complaining. PyLint should spot this. It turns out we're not linting the functional tests: `tests/functional/` and all of its subdirectories need to contain `__init__.py` files otherwise PyLint ignores them.

@marcospri Can you copy [this test from LMS](https://github.com/hypothesis/lms/blob/master/tests/unit/test_missing_init_files.py) into Via 3, to make sure that we don't get anymore missing `__init__.py` files? Then add the missing `__init__.py`'s and fix any lint issues that they reveal. Probably best done as a separate PR, and then rebase _this_ PR once it's merged.

via/views/monitoring.py
```diff
@@ -1,9 +1,9 @@
...
-from pyramid import response, view
+from pyramid import view
 
 
-@view.view_config(route_name="get_status")
+@view.view_config(route_name="get_status", renderer="json")
 def get_status(_request):
     """Status endpoint."""
-    return response.Response(status_int=200, status="200 OK", content_type="text/plain")
+    return {"status": "okay"}
```
seanh: I realise it already exists on master but this file should **really** be called `status.py`

seanh (changes requested):
@marcospri This looks great but there's a lint issue with the functional tests. I think that needs to be fixed in a separate PR first, and then this PR re-based (and the lint warnings in this PR fixed).

If you feel like renaming `monitoring.py` to `status.py` that would be great too

seanh (approved):

PR #389, opened 2021-02-24
Fix handling of relative redirects in NGINX

seanh:
Handle PDF hosts that return relative redirects, such as Box.com.

I thought this was working but it turns out that the test app I had used to generate relative redirects to PDF files was actually returning absolute ones! Relative redirects were not working.

Box.com is an example of a PDF host whose PDF download URLs return chains of three redirects where the first redirect's `Location` header contains a relative URL and the subsequent redirects have absolute `Location`'s. This commit fixes Via 3 with these Box.com URLs.

See https://hypothes-is.slack.com/archives/C4K6M7P5E/p1614112183073900

Testing:

I've created a [localhost (make devdata) Box.com PDF Assignment](https://hypothesis.instructure.com/courses/125/assignments/1504) that doesn't work on master, but does work on this branch.

It's also worth testing all the other localhost assignments in Canvas, and running the [pdf_redirects](https://github.com/hypothesis/pdf_redriects) app and testing the localhost PDF redirects assignments.

### DIFF
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -42,7 +42,29 @@ http {
         # http://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes
         merge_slashes off;
 
-        location ~ ^/proxy/static/(?<sec>[a-zA-Z0-9-_]+)/(?<exp>\d+)/ {
+        # This `location` matches request paths like:
+        #
+        #     /proxy/static/<sec>/<exp>/<proxy_scheme>/<proxy_hostname><proxy_path>
+        #
+        # For example:
+        #
+        #     /proxy/static/abc123/456def/https://foo.example.com/bar/gar.pdf
+        #
+        # The following variables are extracted from the request path for later use:
+        #
+        # ?sec: The checksum for the NGINX secure link module (see includes/secure_links.conf),
+        # e.g. "abc123"
+        #
+        # ?exp: The timestamp for the NGINX secure link module (see includes/secure_links.conf)
+        # e.g. "456def"
+        #
+        # ?proxy_scheme: The scheme part of the URL-to-be-proxied, "http://" or "https://"
+        #
+        # ?proxy_hostname: The hostname part of the URL-to-be-proxied, e.g. "foo.example.com"
+        # Not to be confused with NGINX's builtin $proxy_host variable!
+        #
+        # ?proxy_path: The path part of the URL-to-be-proxied, e.g. "/bar/gar.pdf"
+        location ~ ^/proxy/static/(?<sec>[a-zA-Z0-9-_]+)/(?<exp>\d+)/(?<proxy_scheme>https?://)(?<proxy_hostname>[^/]+)(?<proxy_path>.*) {
             # We don't want our URLs that proxy third-party pages to show in Google.
             include includes/robots.conf;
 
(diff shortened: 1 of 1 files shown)


conf/nginx/nginx.conf
```diff
@@ -69,9 +91,35 @@ http {
...
+            #
+            # http://nginx.org/en/docs/http/ngx_http_upstream_module.html#var_upstream_http_
+            #
+            # Note that we're using NGINX's `if` in a `location` context here,
+            # which is said to be evil. But it seems to work, and I don't know
+            # of any other way to handle both absolute and relative redirects
+            # together.
+            #
+            # http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if
+            # https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/
```
robertknight, line 111: This blog post warns mainly about the use of `if` in a location context. https://cheat.readthedocs.io/en/latest/nginx.html says:

> You CANNOT put any directive you want inside the if, only rewrite directives like set, rewrite, return, etc.

We _are_ using `set` here. However it also says:

> The values of variables you set this way can ONLY be used in if conditions, and maybe rewrite directives; don’t try to use them elsewhere.

That makes it unclear to me what is expected to happen if the variable is used with `proxy_pass`.

robertknight, line 111: http://agentzh.blogspot.com/2011/03/how-nginx-location-if-works.html is an article which has a better explanation of what is going on with `if` and why it doesn't always behave as expected. It gives an example which is very similar to what is used here that should work because of the way that `if` directives interact with nginx rewriting.

This article does raise a query however - what happens with the `add_header` block here? Does that still get applied if either of the `if` conditions match?

conf/nginx/nginx.conf
```diff
@@ -42,7 +42,29 @@ http {
...
+        #
+        # ?exp: The timestamp for the NGINX secure link module (see includes/secure_links.conf)
+        # e.g. "456def"
+        #
+        # ?proxy_scheme: The scheme part of the URL-to-be-proxied, "http://" or "https://"
+        #
+        # ?proxy_hostname: The hostname part of the URL-to-be-proxied, e.g. "foo.example.com"
+        # Not to be confused with NGINX's builtin $proxy_host variable!
+        #
+        # ?proxy_path: The path part of the URL-to-be-proxied, e.g. "/bar/gar.pdf"
```
robertknight, line 66: Will `proxy_path` end up capturing the query string as well?

robertknight (commented):
My initial comment from looking at this is that it seems sensible, but I don't currently understand nginx query processing well enough to know whether the `if` conditions might have unexpected consequences. The example in http://agentzh.blogspot.com/2011/03/how-nginx-location-if-works.html suggests that it will probably work.

I also had a query about whether `proxy_path` contains the query string or not. It doesn't look like `proxy_path` is directly referenced, although it is captured. Is there anything that does use it?

conf/nginx/nginx.conf
seanh: I think so, but we better test that

conf/nginx/nginx.conf
seanh: :man_shrugging: It seems to work. And the main problem is: I don't know of any other way to handle this

seanh: I can check. The `add_header` is just for debugging and it's not even very useful for that, so we can afford to lose it

seanh: Looks like the header is still being added

conf/nginx/nginx.conf
seanh: Tested, it does preserve the query string

PR #392, opened 2021-02-26
Pass blocked_for param along to checkmate

marcospri:


tox.ini
```diff
@@ -27,6 +27,7 @@ setenv =
     dev: NEW_RELIC_ENVIRONMENT = {env:NEW_RELIC_ENVIRONMENT:dev}
     dev: SENTRY_ENVIRONMENT = {env:SENTRY_ENVIRONMENT:dev}
     dev: CHECKMATE_URL = http://localhost:9099
+    dev: CHECKMATE_ENABLED = {env:CHECKMATE_ENABLED:0}
```
seanh:
```suggestion
    dev: CHECKMATE_ENABLED = {env:CHECKMATE_ENABLED:1}
```

Checkmate will be enabled in prod so it should be enabled by default in dev too. (If you don't have Checkmate running then Via 3's requests to Checkmate will time out and then it'll allow the URL to be proxied.)

seanh (changes requested):

PR #402, opened 2021-03-02
Remove checkmate_enabled flag

marcospri:
Waiting for CHECKMATE_ENABLE to be set to `1` on prod before deploying.

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -1,3 +1,4 @@
+import httpretty
 import pytest
 import webtest
 
@@ -7,3 +8,10 @@ from via.app import create_app
 @pytest.fixture
 def test_app(pyramid_settings):
     return webtest.TestApp(create_app(None, **pyramid_settings))
+
+
+@pytest.fixture
+def checkmate_pass(pyramid_settings):
+    httpretty.register_uri(
+        httpretty.GET, "http://localhost:9099/api/check", status=204, body=None
+    )
--- a/tests/unit/via/views/route_by_content_test.py
+++ b/tests/unit/via/views/route_by_content_test.py
@@ -47,22 +47,6 @@ class TestRouteByContent:
 
         get_url_details.assert_called_once_with(url, Any.instance_of(EnvironHeaders))
 
-    def test_blocked_for_passed_to_viahtml(
-        self, call_route_by_content, via_client_service, get_url_details
-    ):
-        get_url_details.return_value = ("html", 200)
-        url = "http://example.com/path%2C?a=b"
-
-        call_route_by_content(
-            url,
-            params={"via.blocked_for": "lms"},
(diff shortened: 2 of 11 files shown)


tests/functional/conftest.py
```diff
@@ -7,3 +8,14 @@
...
     return webtest.TestApp(create_app(None, **pyramid_settings))
+
+
+@pytest.fixture
+def checkmate_pass(pyramid_settings):
+    httpretty.register_uri(
+        httpretty.GET,
+        "http://localhost:9099/api/check",
+        status=204,
+        body="""{"data": []}""",
```
seanh:
```suggestion
        body=None,
```

Checkmate's 204 responses actually have no body

tests/unit/via/views/route_by_content_test.py
```diff
@@ -47,22 +47,6 @@ def test_we_call_third_parties_correctly(
...
-
-        call_route_by_content(
-            url,
-            params={"via.blocked_for": "lms"},
-            settings={"checkmate_enabled": False},
-        )
-
-        via_client_service.url_for.assert_called_once_with(
-            url, content_type="html", options={"via.blocked_for": "lms"}
-        )
```
seanh, line 64: Did you mean to delete this one? It should still pass `blocked_for` to Via right?

seanh (changes requested):

tests/unit/via/views/route_by_content_test.py
marcospri: Not sure, if a checkmate check already happen do we need to pass that to viahtml? That was my original thinking

seanh: Hmm. Does Via HTML still do a Checkmate check? I think when Via 3 redirects to Via HTML, Via HTML still does its own Checkmate check in which case, in theory, the request could pass Via 3's check but get blocked by Via HTML's one (for example if Via 3 fails to reach Checkmate so allows the request through, but then Via HTML succeeds in reaching Checkmate). So I think Via HTML could still generate a Checkmate block page, so it needs the `blocked_for` param.

It does seem like there's a simple optimisation we could do in Via HTML here though: when it sees (from `Referer`) that the request came from Via 3 (or from LMS, which is what I think `Referer` will refer to) it can just trust it and skip any Checkmate check

seanh (approved):

tests/unit/via/views/route_by_content_test.py
seanh: One issue with Via HTML skipping its Checkmate check if the `Referer` is LMS is that it'd allow `curl` requests that send `Referer: lms.hypothes.is` to skip the blocklist. Currently scripts can bypass the LMS-only lockdown and the allow-list by just sending the appropriate `Referer` header, but at least they can't skip the blocklist that way

PR #410, opened 2021-03-10
Adding a demo script for accessing Google Drive files

jon-betts:
One possible solution to: https://github.com/hypothesis/via3/issues/405

The idea is to round robin multiple service accounts to share the user quota out.

This requires `google_drive_demo.json` which contains multiple Google credentials, but they could be used for anything, so I don't want to put them in git.


jon-betts:
The parent ticket has been closed, so this spike is done. For conclusions / follow up see https://github.com/hypothesis/via3/issues/405

PR #418, opened 2021-03-23
nginx - update nginx.conf resolver parameter

indigobravo:
This commit updates the value of the resolver parameter used in `nginx.conf`. The value provided is the hard coded to `169.254.169.253` which is the AWS VPC DNS. This will break functionality if running outside of AWS.

### DIFF
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -22,8 +22,12 @@ events {
 http {
     # Configure the DNS that nginx uses to connect to the servers it's proxying.
     # http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver
-    # TODO: Do we need to change this? Why have we chosen this value?
-    resolver 8.8.8.8 ipv6=off;
+    # Here we are setting the value of the resolver to the AWS VPC DNS server.
+    # Each AWS VPC has a DNS server available at the CIDR base + 2 address.
+    # 169.254.169.253 is mapped to what ever that value is meaning we can deploy
+    # in other AWS VPCs without updating the resolver value. The only requirement
+    # is to have DNS support enabled for the VPC.
+    resolver 169.254.169.253 ipv6=off;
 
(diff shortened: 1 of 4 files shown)


conf/nginx/nginx.conf
```diff
@@ -22,8 +22,12 @@ events {
...
     # Configure the DNS that nginx uses to connect to the servers it's proxying.
     # http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver
-    # TODO: Do we need to change this? Why have we chosen this value?
-    resolver 8.8.8.8 ipv6=off;
+    # Here we are setting the value of the resolver to the AWS VPC DNS server.
+    # Each AWS VPC has a DNS server available at the CIDR base + 2 address.
+    # 169.254.169.253 is mapped to what ever that value is meaning we can deploy
+    # in other AWS VPCs without updating the resolver value. The only requirement
+    # is to have DNS support enabled for the VPC.
+    resolver 169.254.169.253 ipv6=off;
```
indigobravo: Hard code the AWS VNC DNS resolver. Each AWS VPC has a DNS server available at the CIDR base + 2 address. 169.254.169.253 is mapped to whatever that value is meaning we can deploy in other AWS VPCs without updating the resolver value. However, this will break functionality when running outside of AWS.

indigobravo:
Internal thread discussing this change: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1616415118027100

seanh:
I think we should change this to read the resolver string from an envvar. In QA and prod we'll set that envvar to `169.254.169.253` and in dev we'll default it to `8.8.8.8`. See [`envsubst.conf.template`](https://github.com/hypothesis/via3/blob/master/conf/nginx/envsubst.conf.template) for other envvars that're already used in our NGINX config. @jon-betts might be able to take this on?

jon-betts:
I've swapped this to use an include instead. We never want to configure this to anything other than the AWS setting in live from what I can see. This should work for production and dev without adding another config var.

seanh (approved):

PR #420, opened 2021-03-25
Alarm if requests to Checkmate are failing

seanh:
See:

https://github.com/hypothesis/checkmate/issues/190
https://hypothes-is.slack.com/archives/CR3E3S7K8/p1616692590056200?thread_ts=1616581242.031600&cid=CR3E3S7K8

For some reason Via 3 seems to be doing its own thing about various stylistic choices, naming and imports etc. I've fixed the parts of the code I touched to match with the rest of our codebase.

### DIFF
--- a/.pydocstyle
+++ b/.pydocstyle
@@ -1,8 +1,5 @@
 [pydocstyle]
-ignore = D104,  # Missing docstring in public package
-                # Prevents us from having useless comments in __init__.py
-         D105,  # Missing docstring in magic method
-         D107,  # Missing docstring in __init__
+ignore = D100, D101, D102, D103, D104, D105, D106, D107,  # Missing docstrings
          D202,  # "No blank lines allowed after function docstring" conflicts with
                 # the Black code formatter, which insists on inserting blank lines
                 # after function docstrings.
--- a/.pylintrc
+++ b/.pylintrc
@@ -20,7 +20,10 @@ enable=
 disable=bad-continuation,
         missing-type-doc,
         missing-return-type-doc,
-        too-few-public-methods
+        too-few-public-methods,
+        # Docstrings are encouraged but we don't want to enforce that everything
+        # must have a docstring.
+        missing-docstring,
 
 [REPORTS]
 output-format=colorized
--- a/tests/functional/api/view_status_test.py
+++ b/tests/functional/api/view_status_test.py
@@ -1,5 +1,10 @@
-class TestViewMonitoring:
-    def test_status_view(self, test_app):
+import pytest
+
+pytestmark = pytest.mark.usefixtures("checkmate_pass")
+
+
+class TestStatus:
+    def test_it(self, test_app):
         response = test_app.get("/_status")
 
         assert response.json == {"status": "okay"}
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -13,5 +13,5 @@ def test_app(pyramid_settings):
 @pytest.fixture
 def checkmate_pass(pyramid_settings):
     httpretty.register_uri(
-        httpretty.GET, "http://localhost:9099/api/check", status=204, body=None
+        httpretty.GET, "http://localhost:9099/api/check", status=204, body=""
     )
--- a/tests/unit/via/views/__init___test.py
+++ b/tests/unit/via/views/__init___test.py
@@ -11,7 +11,7 @@ class TestIncludeMe:
         views.includeme(config)
 
         assert config.add_route.call_args_list == [
-            mock.call("get_status", "/_status"),
+            mock.call("status", "/_status"),
             mock.call("view_pdf", "/pdf", factory=URLResource),
             mock.call("route_by_content", "/route", factory=URLResource),
             mock.call("debug_headers", "/debug/headers"),
--- a/tests/unit/via/views/status_test.py
+++ b/tests/unit/via/views/status_test.py
@@ -1,10 +1,32 @@
-from via.views.status import get_status
+import pytest
+from checkmatelib import CheckmateException
+
+from via.views.status import status
 
 
-class TestStatusRoute:
-    def test_status_returns_200_response(self, make_request):
+class TestStatus:
+    def test_it_returns_200_response(self, make_request):
         request = make_request("/_status")
 
-        result = get_status(request)
+        result = status(request)
 
         assert result == {"status": "okay"}
+
+    def test_it_reports_failure_if_calling_checkmate_fails(
+        self, checkmate, make_request
+    ):
+        checkmate.check_url.side_effect = CheckmateException
+
+        result = status(make_request("/_status"))
+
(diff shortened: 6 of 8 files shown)


.pydocstyle
```diff
@@ -1,8 +1,5 @@
 [pydocstyle]
-ignore = D104,  # Missing docstring in public package
-                # Prevents us from having useless comments in __init__.py
-         D105,  # Missing docstring in magic method
-         D107,  # Missing docstring in __init__
+ignore = D100, D101, D102, D103, D104, D105, D106, D107,  # Missing docstrings
```
seanh, line 2: Although we totally encourage docstrings we don't force ourselves to put them on every single thing as that results in pointless docstrings. Via 3 had some of the docstring warnings ignored but not all of them. Ignore them all as we do in other projects

.pylintrc
```diff
@@ -20,7 +20,10 @@ enable=
 disable=bad-continuation,
         missing-type-doc,
         missing-return-type-doc,
-        too-few-public-methods
+        too-few-public-methods,
+        # Docstrings are encouraged but we don't want to enforce that everything
+        # must have a docstring.
+        missing-docstring,
```
seanh, line 26: Ignore this, again as we do for other projects. There's doubly no reason to have PyLint check for missing docstrings if we're also running pydocstyle.

tests/functional/api/view_status_test.py
```diff
@@ -1,5 +1,10 @@
-class TestViewMonitoring:
-    def test_status_view(self, test_app):
+import pytest
+
+pytestmark = pytest.mark.usefixtures("checkmate_pass")
+
+
+class TestStatus:
+    def test_it(self, test_app):
```
seanh, line 7: Fixed up the naming here to match our usual naming style. The function under test is called `status()` so the test class is `TestStatus` and the test methods are called `test_it_*()`. When there's only one test like here, it can just be `test_it()`

tests/functional/conftest.py
```diff
@@ -13,5 +13,5 @@ def test_app(pyramid_settings):
 @pytest.fixture
 def checkmate_pass(pyramid_settings):
     httpretty.register_uri(
-        httpretty.GET, "http://localhost:9099/api/check", status=204, body=None
+        httpretty.GET, "http://localhost:9099/api/check", status=204, body=""
```
seanh, line 16: `body=None` causes `httpretty` to crash (which makes me wonder whether this fixture was ever working before)

tests/unit/via/views/__init___test.py
```diff
@@ -11,7 +11,7 @@ def test_include_me(self):
         views.includeme(config)
 
         assert config.add_route.call_args_list == [
-            mock.call("get_status", "/_status"),
+            mock.call("status", "/_status"),
```
seanh, line 14: I renamed the route to just `"status"` as per our normal style. Not sure what the point of `"get_"` was

tests/unit/via/views/status_test.py
```diff
@@ -1,10 +1,32 @@
...
+import pytest
+from checkmatelib import CheckmateException
 
+from via.views.status import status
 
-class TestStatusRoute:
-    def test_status_returns_200_response(self, make_request):
+
+class TestStatus:
+    def test_it_returns_200_response(self, make_request):
```
seanh, line 8: Again fixing the test class and method name to match our usual style. There's no point in having tests called `TestStatus.test_status_*`. `test_it_*` gives you more room for the important part of the test name

via/views/status.py
```diff
@@ -1,9 +1,25 @@
...
 
+logger = logging.getLogger(__name__)
 
-@view.view_config(route_name="get_status", renderer="json")
-def get_status(_request):
-    """Status endpoint."""
-    return {"status": "okay"}
+
+@view_config(route_name="status", renderer="json")
+def status(request):
```
seanh, line 10: Again fixed the function naming style.

via/views/status.py
```diff
@@ -1,9 +1,25 @@
-"""Monitoring views."""
```
seanh, line 1: Pointless docstring. It's also wrong: `status.py` isn't a general "monitoring views" (plural) module. It's just the module that contains the single status view. But there's really no need for a docstring here, just delete it

via/views/status.py
```diff
@@ -1,9 +1,25 @@
...
 
-from pyramid import view
+from checkmatelib import CheckmateClient, CheckmateException
+from pyramid.view import view_config
 
+logger = logging.getLogger(__name__)
 
-@view.view_config(route_name="get_status", renderer="json")
-def get_status(_request):
-    """Status endpoint."""
```
seanh, line 8: Another pointless docstring (and also technically wrong: the function is a view not an endpoint)

via/views/status.py
```diff
@@ -1,9 +1,25 @@
-"""Monitoring views."""
+import logging
 
-from pyramid import view
+from checkmatelib import CheckmateClient, CheckmateException
+from pyramid.view import view_config
```
seanh, line 4: For some reason Via 3 keeps doing `from pyramid import view` and then using `@view.view_config` everywhere. All our apps (and every Pyramid app ever) do `from pyramid.view import view_config` and then `@view_config`. I fixed it for this one module at least

via/views/status.py
```diff
@@ -1,9 +1,25 @@
...
+
+    status_ = "okay"
+
+    try:
+        checkmate.check_url("https://example.com/")
+    except CheckmateException:
+        status_ = "checkmate_failure"
+        logger.exception("Checkmate request failed")
+
+    return {"status": status_}
```
marcospri, line 25: Do we need to return necessarily a {"status": str} structure?

Maybe something a bit more complex would be eaiser once there's more stuff to check

```
{
"api": okey,
"checkmate": ko
... 
"viahtml": ok"
....
}
````

seanh: I think we need to return `{"status": status}` because there are existing ops scripts that need that (and I can't be bothered to change them).

We want one simple ops script to work for all our apps so `{"status": status}` is probably a good, simple, universal pattern anyway. We _could_ add more fields just for human eyes (not used by the ops scripts), e.g. `{"status": "failure", "checkmate": "failed", "some_other_thing": "ok"}`. But to be honest `{"status": <whatever_string_we_want>}` is enough for our needs, and I've see cleverer status endpoints quickly get out of hand in the past, with people making them complicated enough that the code and its tests were actually time-consuming to maintain and were a hazard in production because requests to the (publicly available) status endpoint could take up too many resources. Example: <https://github.com/hypothesis/h-periodic/pull/49> (250-lines of status endpoints and tests, and it didn't even work!). So we should probably stick to `{"status": status}` just to encourage us to restrain ourselves.

If we want to know more detail about what's working and what's not about an app there are better ways: Papertrail logs, Sentry, New Relic, Elastic Beanstalk's monitoring.

marcospri (approved):

PR #426, opened 2021-03-30
Use built-in `tojson` to escape JSON strings

robertknight:
_This is preparation for fixing https://github.com/hypothesis/via3/issues/372_.

Use Jinja2's built in `tojson` filter to escape JSON data rather than
doing it ourselves in the view function:

 - It avoids the view functions needing to deal with escaping issues and
   moves the escaping closer to the sink, reducing the risk of
   context-inappropriate escaping.

 - This avoids a potential hazard specific to escaping JSON
   data in `<script>` contexts [1], where additional escaping is required (see
   the `htmlsafe_json_dumps` function in Jinja2 [2])

[1] https://sophiebits.com/2012/08/03/preventing-xss-json.html
[2] https://github.com/pallets/jinja/blob/8ab9db26f36a976212debd02a7a46265fa6dd1c1/src/jinja2/utils.py#L613

### DIFF
--- a/tests/unit/via/views/view_pdf_test.py
+++ b/tests/unit/via/views/view_pdf_test.py
@@ -1,10 +1,8 @@
-import json
 from datetime import datetime, timedelta, timezone
 from unittest.mock import sentinel
 
 import pytest
 from h_matchers import Any
-from markupsafe import Markup
 
 from via.resources import URLResource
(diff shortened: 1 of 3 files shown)


jon-betts (approved):
Yep, this is much nicer and neater. I think this was "right idea, wrong implementation" before.

PR #427, opened 2021-03-31
Set correct `originalUrl` in PDF.js config

### LINKED ISSUE #372: PDF annotations have incorrect `target_uri`
**Update (2021-03-19): via3.hypothes.is was replaced by via4.hypothes.is. The issue of annotations having the wrong `target_uri` when annotated through Via persists. See [comment](https://github.com/hypothesis/via3/issues/372#issuecomment-802622990) - @robertknight**

I found mentions of via3 URL in the `target_uri` in the annotation table. This happens mostly for PDF, for example, `https://via3.hypothes.is/proxy/static/https://whc.unesco.org/archive/convention-en.pdf`

Ideally the `https://via3.hypothes.is/proxy/static/` shouldn’t be there.

A few more examples of these URL can be retrieved using this SQL query:  `select * from document_uri where claimant ~'via3';`

## Steps to reproduce

1. Go https://hypothesis.instructure.com/courses/125/assignments/878
2. Add an annotations

## Expected result

`target_uri` should be `https://h.readthedocs.io/_/downloads/en/latest/pdf/`

## Actual result

`target_uri` is `https://qa-via3.hypothes.is/proxy/static/fApI11LAAE8wQAJpdjyVJQ/1613066400/https://h.readthedocs.io/_/downloads/en/latest/pdf`

## Additional information

This issue also involves to the client and h. The client uses `document.location.href` as the target of the annotation. For PDF downloads this is the whole URL.

If there would be a way to eliminate the via3 portion when creating annotation it would mean that one could find annotations either in:

- `https://via3.hypothes.is/proxy/static/https://whc.unesco.org/archive/convention-en.pdf`, or
- `https://whc.unesco.org/archive/convention-en.pdf`

## Possible solutions

- The client strips the via3 portion of the URL. This is not very clean.
- Strip the via3 portion of the URL on the backend. Not very clean neither.
- I have heard there is some sort of unique signature on a PDF. Could that be a solution?

## Warning

Whatever we do, it should be backward compatible. People should be able to find their old annotations, even thought they were included using a different URL.

robertknight:
~~**Depends on https://github.com/hypothesis/via3/pull/426**~~

This PR fixes an issue that the `originalUrl` value in the PDF.js configuration was set incorrectly, causing the Hypothesis client to associate the wrong main URL with annotations. The URL associated with annotations should be the original URL of the PDF, not the proxied URL from which PDF.js actually loads the PDF.

The PR splits the existing `pdf_url` output from the `view_pdf` function into `proxied_pdf_url` and `original_pdf_url` which are then passed through to the PDF.js configuration as the `url` and `originalUrl` config params

You can observe the effect this has by clicking the "?" icon in the Hypothesis client and selecting "About this version" to view what the client uses as the main document URL for annotations:

**Before:**

Proxied PDF wrong URL

**After:**

Proxied PDF correct URL

Setting the correct URL is important in contexts like the Notebook which display the document's domain in the annotation card as well as linking to the document.

Although the URL being associated with documents is changing here, this should not break the connection to existing annotations because the PDF fingerprint is the same as before, and both the main URL and fingerprint are used when fetching annotations. There is value in fixing up the existing URLs that have been recorded incorrectly, because it will affect how these documents are displayed in the Notebook, but it is not essential to do before/alongside deploying this change.

Fixes #372


bin/minify_assets.py
```diff
@@ -80,7 +80,7 @@ def _find_files(cls, instructions):
 
     @classmethod
     def _execute(cls, path, handler, in_place=False):
-        with open(path) as handle:
+        with open(path, encoding="UTF-8") as handle:
```
robertknight: This method of setting the encoding works but has to be applied everywhere that `open` is called. I was surprised that UTF-8 was not the default encoding used in this context. https://docs.python.org/3/library/functions.html#open says:

> The default encoding is platform dependent (whatever locale.getpreferredencoding() returns)

Printing out the value of `locale.getpreferredencoding()` just before calling `open` prints `UTF-8`. The path for which an exception is thrown is `via/templates/pdf_wrapper.html.jinja2`. If I open a Python 3.6 shell and run `open("via/templates/pdf_wrapper.html.jinja2").read()` then it runs with no error. Perhaps something is altering the codec that is used at the point where this is called? I'm currently digging further...

robertknight: Ah, I found something interesting. The Python builtin docs for `open`, shown via `help(open)` state:

> In text mode, if encoding is not specified the encoding used is platform dependent: locale.getpreferredencoding(False) is called to get the current locale encoding.

Changing the `bin/minify-assets.py` script to:

```py
import locale
print(f"locale.getpreferredencoding(): {locale.getpreferredencoding()}")
print(f"locale.getpreferredencoding(False): {locale.getpreferredencoding(False)}")
open('via/templates/pdf_wrapper.html.jinja2').read()
```

And running `make build` outputs:

```
current encoding: UTF-8
current encoding: US-ASCII
Traceback (most recent call last):
  File "bin/minify_assets.py", line 6, in <module>
    open('via/templates/pdf_wrapper.html.jinja2').read()
  File "/Users/robert/.pyenv/versions/3.6.9/lib/python3.6/encodings/ascii.py", line 26, in deco
de
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 3038: ordinal not in range
(128)
```

The `False` argument is `do_setlocale` which says:

> On some systems, it is necessary to invoke setlocale() to obtain the user preferences, so this function is not thread-safe. If invoking setlocale is not necessary or desired, do_setlocale should be set to False.

So it looks like there is some global interpreter state that is different in these two situations.

robertknight: It looks like the reason for the difference is tox's environment filtering. In a normal Python shell `LC_ALL` is set to `en_GB.UTF-8` on my system. In a tox environment `LC_ALL` is not set. 

It looks like this problem can be resolved in Python >= 3.7 by enabling [UTF-8 mode](https://www.python.org/dev/peps/pep-0540/). This can be done in tox by setting the PYTHONUTF8 env var.

robertknight: Update: It seems that the way this should work is that the `LANG` env var is in tox's env var whitelist (`passenv`) by default and this value should enable the use of UTF-8. In my terminal (Alacritty) running `locale` reports `LANG` as empty string. If however I run `locale` in the default macOS Terminal app then `locale` reports `LANG` as `en_GB.UTF-8`. So it looks like I need to fix whatever is causing `LANG` not to be set on my end.

I've removed this commit from the PR for now.

jon-betts (changes requested):
This looks good, but I honestly don't know how you are supposed to test this in an isolated fashion any more. I had to run this all the way through Canvas to get the client to show up. Any tips definitely welcome.

I do have one request for a change which is around the naming of "original_pdf_url" and "proxied_pdf_url". For my money the second one is quite ambiguous. How I think it's meant to be read:

* (original_pdf)_url - The URL for the original PDF
* proxied_(pdf_url) - The proxy of the (original)PDF URL

However how my brain tends to read this:

* (proxied_pdf)_url - The URL of the proxied PDF (i.e. the original URL again)

I'd suggest either:

* pdf_url / proxy_pdf_url - The original takes the prime position and the new var is clearly different from it
* original_pdf_url / proxy_pdf_url - This kind of disambiguates as a "proxy pdf" isn't a thing, so it must be a proxy "pdf url". If you see what I mean...

Of the two, the first is probably easier and would mean that the code mostly agrees that in the absence of any qualifiers "the" URL is the original, which I think makes sense.

robertknight:
> This looks good, but I honestly don't know how you are supposed to test this in an isolated fashion any more. I had to run this all the way through Canvas to get the client to show up. Any tips definitely welcome.

I tested this by running client + h + via3 and visiting http://localhost:9083/pdf?url=https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf. Then I checked that the expected URLs were used a) to load the PDF and b) shown in the client's "About this version" page.

> I'd suggest either:
> pdf_url / proxy_pdf_url - The original takes the prime position and the new var is clearly different from it

OK, I've run with this suggestion. Given that the ambiguity here is tricky to avoid by naming alone I think the most important thing is to document the intended usage clearly and be consistent about the terminology.

PR #430, opened 2021-04-02
Prevent caching of the status end-point

jon-betts:
For: hypothesis/product-backlog#1198

This also includes a fix that allows the func tests to be run individually.


via/app.py
```diff
@@ -29,19 +34,19 @@ def load_settings(settings):
...
+
+        if value is None and options.get("required"):
             raise ValueError(f"Param {param} must be provided.")
 
     # Configure sentry
     settings["h_pyramid_sentry.filters"] = SENTRY_FILTERS
 
-    settings["checkmate_ignore_reasons"] = os.environ.get("CHECKMATE_IGNORE_REASONS")
-    settings["signed_urls_required"] = asbool(os.environ.get("SIGNED_URLS_REQUIRED"))
-    settings["nginx_secure_link_secret"] = os.environ["NGINX_SECURE_LINK_SECRET"]
```
jon-betts, line 43: These settings were added outside of the mechanism above, meaning these ones didn't respect things passed in via `settings`

jon-betts:
Fixing up the settings in the func tests has broken something the tests were relying to pass for now... not a Friday job though

tests/conftest.py
```diff
@@ -12,7 +12,7 @@ def pyramid_settings():
         "checkmate_url": "http://localhost:9099",
         "nginx_secure_link_secret": "not_a_secret",
         "via_secret": "not_a_secret",
-        "signed_urls_required": True,
+        "signed_urls_required": False,
```
jon-betts, line 15: Yeah, turns out this isn't actually the value we were using! This was being ignored and overwritten with the empty env var.

tox.ini
```diff
@@ -28,7 +28,7 @@ setenv =
     dev: SENTRY_ENVIRONMENT = {env:SENTRY_ENVIRONMENT:dev}
     dev: CHECKMATE_URL = http://localhost:9099
     dev: CHECKMATE_IGNORE_REASONS = {env:CHECKMATE_IGNORE_REASONS:publisher-blocked}
-    {dev,functests}: NGINX_SECURE_LINK_SECRET = not_a_secret
+    dev: NGINX_SECURE_LINK_SECRET = not_a_secret
```
jon-betts, line 31: This was the only value being set from the outside intentionally. Now it's all in the `conftest.py`

seanh (approved):

PR #435, opened 2021-04-09
Optionally use allow list

seanh:
Instead of just the blocklist.

Adds a `CHECKMATE_ALLOW_ALL` boolean envvar. This will be set to `True` for LMS's Via HTML and `False` for the public Via HTML.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,6 +28,7 @@ setenv =
     dev: SENTRY_ENVIRONMENT = {env:SENTRY_ENVIRONMENT:dev}
     dev: CHECKMATE_URL = http://localhost:9099
     dev: CHECKMATE_IGNORE_REASONS = {env:CHECKMATE_IGNORE_REASONS:publisher-blocked}
+    dev: CHECKMATE_ALLOW_ALL = {env:CHECKMATE_ALLOW_ALL:true}
     dev: NGINX_SECURE_LINK_SECRET = not_a_secret
     dev: VIA_SECRET = not_a_secret
     dev: CHECKMATE_API_KEY = dev_api_key
--- a/via/app.py
+++ b/via/app.py
@@ -20,6 +20,7 @@ PARAMETERS = {
     "nginx_secure_link_secret": {"required": True},
     # Optional
     "checkmate_ignore_reasons": {},
+    "checkmate_allow_all": {"formatter": asbool},
     "signed_urls_required": {"formatter": asbool},
 }
 
--- a/via/views/decorators.py
+++ b/via/views/decorators.py
@@ -23,7 +23,7 @@ def checkmate_block(view):
         try:
             blocked = checkmate.check_url(
                 url,
-                allow_all=True,
+                allow_all=request.registry.settings["checkmate_allow_all"],
                 blocked_for=blocked_for,
                 ignore_reasons=request.registry.settings["checkmate_ignore_reasons"],
             )
--- a/via/views/decorators.py
+++ b/via/views/decorators.py
@@ -5,13 +5,10 @@ from h_vialib.secure.url import ViaSecureURL
 from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPUnauthorized
 
 
-def checkmate_block(view, url_param="url", allow_all=True):
+def checkmate_block(view):
     """Intended to be used as a decorator for pyramid views.
 
     The view must accept a url a query param.
-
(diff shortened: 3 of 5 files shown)


via/views/decorators.py
```diff
@@ -5,13 +5,10 @@
 from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPUnauthorized
 
 
-def checkmate_block(view, url_param="url", allow_all=True):
```
seanh, line 8: I removed them in a separate commit but non-default values were never passed for either of these args by either the code or the tests. Nor _can_ they be passed because this function is a view decorator, it's called by Pyramid not by us

tox.ini
```diff
@@ -28,6 +28,7 @@ setenv =
     dev: SENTRY_ENVIRONMENT = {env:SENTRY_ENVIRONMENT:dev}
     dev: CHECKMATE_URL = http://localhost:9099
     dev: CHECKMATE_IGNORE_REASONS = {env:CHECKMATE_IGNORE_REASONS:publisher-blocked}
+    dev: CHECKMATE_ALLOW_ALL = {env:CHECKMATE_ALLOW_ALL:true}
```
seanh, line 31: So by default the allow-list will _not_ be used in dev, which is more convenient. But you can turn it on with an envvar

via/app.py
```diff
@@ -20,6 +20,7 @@
     "nginx_secure_link_secret": {"required": True},
     # Optional
     "checkmate_ignore_reasons": {},
+    "checkmate_allow_all": {"formatter": asbool},
```
seanh, line 23: `asbool` will cause `checkmate_allow_all` to be `False` if the envvar isn't set. So we'll have to add `CHECKMATE_ALLOW_ALL=true` to LMS's Via 3 in QA and prod before deploying this

via/views/decorators.py
```diff
@@ -20,14 +17,11 @@ def view_wrapper(context, request):
...
-
         try:
             blocked = checkmate.check_url(
-                url,
-                allow_all=allow_all,
-                blocked_for=blocked_for,
+                request.params["url"],
+                allow_all=request.registry.settings["checkmate_allow_all"],
+                blocked_for=request.params.get("via.blocked_for"),
                 ignore_reasons=request.registry.settings["checkmate_ignore_reasons"],
```
jon-betts, line 25: I think this all looks good, but it's kind of scratching at an itch I think is of my own making.

If this were a service, rather than a 3rd party object it feels like there's a lot going on in the call to `check_url` that you'd normally have in the construction of the service, and the call to `check_url` would only take a URL.

The split we have between these arguments is logical though:

 * The arguments to the constructor don't really change per call: Checkmate is where it is, no matter what URL you are checking)
 * The arguments to `check_url` _could_ change: You could conceivably want to check different URLs in different ways

However the issue I have is that _practically_, the arguments don't actually change from call to call in any case so far. I think the args as they are right now is actually logically correct for a library, but sit wrong for a service.

So some ideas to fix:

* Add a service factory that returns a `functools.partial` with all this stuff pre-filled or something (maybe a full service object)
* Change the definition of the `CheckmateClient` - seems kind of wrong and current usage specific
* Just not worry about it?

Of these I think I prefer adding a small service / doing nothing. A service would be quite small and easy to implement, but also seems like an annoying indirection, but it would make these tests very easy? Checkmate doesn't do anything expensive when you instantiate it (at the moment), so aside from the actual dict lookups themselves, there's nothing slow going on here that a service would measurably improve.

I honestly dunno...

via/app.py
jon-betts: Yeah we could add "default"... but nah. Lets not re-invent `argparse`

jon-betts (approved):
I think this PR is actually fine, but it's pinged some heeby jeebies I have about the number of static values we are reading from the `request.registry` on every call. Not because it's slow or anything, but it just feels like it's pointing out a flaw in the design.

Some incoherent ramblings below about that. Basically we could fix this with a small service along the lines of:

```python
def factory(context, registry):
    checkmate = CheckmateClient(
        request.registry.settings["checkmate_url"],
        request.registry.settings["checkmate_api_key"],
    )

    return functools.partial(
        checkmate.check_url,
        allow_all=request.registry.settings["checkmate_allow_all"],
        blocked_for=request.params.get("via.blocked_for"),
        ignore_reasons=request.registry.settings["checkmate_ignore_reasons"],
    )
```

PR #438, opened 2021-04-14
Upgrade h-vialib to 1.0.14

seanh:
In 1.0.14 h-vialib no longer adds the `via.client.requestConfigFromFrame.*` options to Via URLs by default, see: https://github.com/hypothesis/h-vialib/pull/23

As a result Via 3 will no longer add the `via.client.requestConfigFromFrame.*` when it generates Via HTML URLs to redirect the browser to.

This is correct:

1. For LMS's Via 3 the LMS app will already have generated the `via.client.requestConfigFromFrame.*` when it generates the Via 3 URL and Via 3 will blindly pass these on to Via HTML.

2. In the future when Via 3 is used directly (as public Via) (so when Via 3 is *not* being redirected to by the LMS app) we actually don't want the `via.client.requestConfigFromFrame.*` options. Those options are only supposed to be used in LMS

### DIFF
--- a/requirements/build.txt
+++ b/requirements/build.txt
@@ -4,7 +4,11 @@
 #
 #    pip-compile requirements/build.in
 #
-htmlmin==0.1.12           # via -r requirements/build.in
-rcssmin==1.0.6            # via -r requirements/build.in
-rjsmin==1.1.0             # via -r requirements/build.in
-whitenoise==5.2.0         # via -r requirements/build.in
+htmlmin==0.1.12
+    # via -r requirements/build.in
+rcssmin==1.0.6
+    # via -r requirements/build.in
+rjsmin==1.1.0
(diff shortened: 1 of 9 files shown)


marcospri:
~~Do we need https://github.com/hypothesis/devdata/blob/master/lms/devdata.env#L5 merged first?~~

Meant https://github.com/hypothesis/lms/pull/2565 but I see you mention there it depends on this one :+1:

seanh:
@marcospri Wrong link?

jon-betts (approved):
This looks good to me. The original idea was that new arguments could be added and passed through from caller to `h` client without having to alter Via and friends at all, so I think this actually moves us one step closer to that.

Now if we change our minds about what these arguments are called, or how they work we only have to alter LMS and the client.

PR #439, opened 2021-04-14
Use DummyRequest

seanh:
Same as recently done for Checkmate:

https://github.com/hypothesis/checkmate/pull/280

Replace the `make_request` fixture that returns a function for creating real `Request` objects with a `pyramid_request` fixture that returns a `DummyRequest`.

The `make_request` fixture was making it difficult for me to add tests that I need to add on another branch both because it doesn't integrate with pytest fixtures (fixture overriding etc) in quite the same way (since the fixture returns a function that returns a request, rather than the fixture returning a request directly) and because it's not using `DummyRequest` (so e.g. you can't easily change the request params, and lots of other things).

The `pyramid_request` fixture is also in the style familiar from the rest of our apps.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,10 +1,9 @@
 # pylint: disable=no-self-use
 """A place to put fixture functions that are useful application-wide."""
 import functools
-from urllib.parse import urlencode
 
 from pyramid import testing
-from pyramid.request import Request, apply_request_extensions
+from pyramid.request import apply_request_extensions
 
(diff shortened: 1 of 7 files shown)


jon-betts (approved):
This is good and the culmination of unravelling the tests that this project started with. The previous incarnation of `make_request` was already considerably less gnarly than the one before etc. Less code too!

Looking at this I think we should have probably jumped a few steps and gone directly to this instead of working incrementally, as the change is actually pretty simple. But both things have arguments for them.

PR #440, opened 2021-04-14
Add Checkmate to status endpoint

seanh:
Add an optional Checkmate check to the status endpoint.

### DIFF
--- a/via/app.py
+++ b/via/app.py
@@ -7,6 +7,7 @@ from pyramid.settings import asbool
 from whitenoise import WhiteNoise
 
 from via.cache_buster import PathCacheBuster
+from via.checkmate import checkmate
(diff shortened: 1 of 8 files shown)


via/app.py
```diff
@@ -70,6 +71,9 @@ def create_app(_=None, **settings):
     config.add_static_view(name="static", path="via:static")
     config.add_cache_buster("via:static", cachebust=cache_buster)
 
+    # Make the CheckmateClient object available as request.checkmate.
+    config.add_request_method(checkmate, reify=True)
```
seanh, line 75: Constructing the `CheckmateClient` once here and adding it as `request.checkmate` so that it doesn't have to be constructed in two different places

marcospri (approved):

PR #453, opened 2021-04-23
Reflect the document title of proxied HTML content

robertknight:
Listen for `metadatachange` messages sent via `window.postMessage` from
the viahtml iframe and update the document title of top frame to reflect
the current document in the proxied frame.

Currently the title is the only proxied metadata but the mechanism could
in future be extended to update the document URL, favicon etc.

Counterpart PR in Via 3: https://github.com/hypothesis/viahtml/pull/160

A note on JavaScript language compatibility: I am assuming a modern browser which is capable of running the Hypothesis client. See `browserslist` in the client repository's `package.json` file for a list of minimum supported browser versions.

Part of https://github.com/hypothesis/viahtml/issues/154

### DIFF
--- a/via/templates/proxy.html.jinja2
+++ b/via/templates/proxy.html.jinja2
@@ -24,8 +24,45 @@
       padding: 0;
     }
   </style>
+
+  <script>
+  /**
+   * Perform basic validation of the structure of a notification from the content
+   * frame.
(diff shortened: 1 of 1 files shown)


seanh (approved):
:+1: Works for me

PR #454, opened 2021-04-23
`npm` based HTML, CSS and JS minification

jon-betts:
For: https://github.com/hypothesis/via3/issues/444

The front-end ecosystem is likely to have more up-to-date and well supported methods for dealing with front-end assets than the Python ecosystem. 

This removes the following poorly supported Python dependencies:

* `rcssmin`
* `htmlmin`
* `rjsmin`

And replaces them with `npm` dependencies:

* `terser` - As used by the front-end team
* `sass` - As used by the front-end team
* ~[`html-minifier`](https://www.npmjs.com/package/html-minifier) - This is a new dependency as we don't do HTML minification in the front end~ Removed as review feedback

So in total we replace 3 python dependencies with ~1 new npm one~ two `npm` packages we already use.

## Review notes

 * Most of the lines of this change are just a big auto generated `package-lock.json` file
 * The minifications script now only accepts minification if the result is smaller
 * To keep the noise down I'll commit the minifications in a separate ticket if this is accepted
 
 ## Testing notes
 
  * Run `make build`
  * We might have to merge this to see how Jenkins behaves when building the image
  * If the image that Jenkins is running in has `npm` available, no changes may be required


bin/minify_assets.py
```diff
@@ -21,18 +18,45 @@
...
+
+def minify_html(file_name):
+    return check_output(
+        [
+            "./node_modules/.bin/html-minifier",
+            "--collapse-whitespace",
+            "--remove-optional-tags",
+            "--remove-redundant-attributes",
+            "--remove-script-type-attributes",
+            "--remove-tag-whitespace",
```
jon-betts: Don't remove comments as they include a copyright statement

package.json
```diff
@@ -0,0 +1,17 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/hypothesis/via3/issues"
+  },
+  "dependencies": {
+    "html-minifier": "^4.0.0",
+    "sass": "^1.32.11",
+    "terser": "^5.6.1"
+  }
+}
```
jon-betts, line 17: Generated by doing:

* `npm init`
* Answering a bunch of questions
* Removing some fields that don't make sense
* Running `npm install sass terser html-minifier`

robertknight: I think you'll want to also set the `private` field on this package to mark it as not-for-publishing.

bin/minify_assets.py
```diff
@@ -21,18 +18,45 @@
 )
 
 
-def _htmlmin(content):
-    return minify(content, remove_comments=True, remove_empty_space=True)
+def minify_js(file_name):
+    return check_output(
+        ["./node_modules/.bin/terser", "--compress", "--safari10", file_name]
+    ).decode("utf-8")
```
jon-betts, line 24: All of these command should be trivial to swap out now if we find better modules as long as:

 * They accept a file as input
 * Can output to STDOUT

package.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "Via 3",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/hypothesis/via3.git"
+  },
+  "bugs": {
+    "url": "https://github.com/hypothesis/via3/issues"
+  },
+  "dependencies": {
+    "html-minifier": "^4.0.0",
```
robertknight: `html-minifier` includes entirely separate CSS and JS minifiers as dependencies. This includes a copy of the precursor to terser [which doesn't support modern JS](https://github.com/kangax/html-minifier/issues/1040). Given the small amount of HTML content in Via 3 (viewer.html and pdf_wrapper.html.jinja2) my preference would be not to bother minifying it.

bin/minify_assets.py
```diff
@@ -21,18 +18,45 @@
...
 
     handlers = {
-        "js": jsmin,
-        "css": cssmin,
-        "html": _htmlmin,
-        "jinja2": _htmlmin,
+        "js": minify_js,
+        "css": minify_css,
+        "html": minify_html,
+        "jinja2": minify_html,
```
robertknight: Are there any ways in which applying HTML minification to a Jinja2 template could silently produce undesirable results?

robertknight (commented):
The JS and CSS minification changes look good. I would recommend that we drop the HTML minification because I don't think it provides a whole lot of value and it introduces some future maintenance hazards (eg. due to including old CSS and JS minifiers and because we're applying it to a Jinja template here).

bin/minify_assets.py
jon-betts: It's certainly possible if we had tags in comments etc. Maybe it's best not to bother for now.

robertknight (approved):
Looks like there was a formatting issue on CI, otherwise LGTM.

package.json
jon-betts: Done!

bin/minify_assets.py
```diff
@@ -21,19 +18,22 @@
...
 class Minifier:
     """Minifier for CSS, Javascript and HTML."""
 
-    handlers = {
-        "js": jsmin,
-        "css": cssmin,
-        "html": _htmlmin,
-        "jinja2": _htmlmin,
-    }
+    handlers = {"js": minify_js, "css": minify_css}
```
jon-betts, line 36: We don't need HTML minification any more as we don't use it (because it's potentially dangerous over Jinja files)

bin/minify_assets.py
jon-betts: I've removed the HTML stuff, we only ran it on one thing and that was a template, which didn't have issues (it worked), but easily could if you tried to do something which the minifier didn't understand in Jinja2.

marcospri:
> * We might have to merge this to see how Jenkins behaves when building the image
> * If the image that Jenkins is running in has `npm` available, no changes may be required

Looks good on local. As you say not sure if in CI everything will be where it should, but let's merge it and find out :+1:

marcospri (approved):

PR #459, opened 2021-04-27
Allow Google to index the front page

### LINKED ISSUE #168: Allow the new Via's front page to show in Google search results
Allow the front page at via.hypothes.is to show up in search results, while still preventing all other pages from showing.

This probably just means removing the `noindex` header from the front page, but leaving it on all other pages.

The `noindex` header **should still be used even on the front page, for LMS's private instances of Via**.

See:

* [Get Via 3 out of Google (and keep it out)](https://github.com/hypothesis/checkmate/issues/167)
* [Get Via HTML out of Google (and keep it out)](https://github.com/hypothesis/checkmate/issues/166)

seanh:
Fixes https://github.com/hypothesis/checkmate/issues/168

With this PR we're still using NGINX to add the `X-Robots-Tag: noindex, nofollow` header to the proxied responses when NGINX proxies the actual PDF files. This has to be done in NGINX as those responses are handled directly by NGINX, no Python:

https://github.com/hypothesis/via3/blob/de537006593acbf3d5a6160e2822d3e5173d2d0f/conf/nginx/nginx.conf#L67

But we no longer use NGINX to add the header to the responses from Python. Instead we're using a Pyramid tween to add the header to every response from the Pyramid app. This makes sure that all of the Pyramid app's pages, including error pages, have the `X-Robots-Tag: noindex, nofollow` header.

The tween allows individual responses to opt-out of the `X-Robots-Tag: noindex, nofollow` by setting their own `X-Robots-Tag` header, and the view for the front page does this.

There is one unintended consequence of this: the tween doesn't add the header to static file responses served by [WhiteNoise](http://whitenoise.evans.io/en/stable/) whereas the previous NGINX-based approach to adding the header to every Python response did. I don't think this is important.

### DIFF
--- a/via/tweens.py
+++ b/via/tweens.py
@@ -1,0 +1,15 @@
+def robots_tween_factory(handler, _registry):
+    def robots_tween(request):
+        response = handler(request)
+
+        # If the view hasn't set its own X-Robots-Tag header then set one that
+        # tells Google (and most other crawlers) not to index the page and not
+        # to follow links on the page.
+        #
+        # https://developers.google.com/search/reference/robots_meta_tag
+        if "X-Robots-Tag" not in response.headers:
+            response.headers["X-Robots-Tag"] = "noindex, nofollow"
+
+        return response
+
+    return robots_tween
--- a/via/views/index.py
+++ b/via/views/index.py
@@ -13,6 +13,8 @@ class IndexViews:
         if not self.enabled:
             return HTTPNotFound()
 
+        self.request.response.headers["X-Robots-Tag"] = "all"
+
         return {}
 
     @view_config(request_method="POST")
--- a/conf/nginx/nginx.conf
+++ b/conf/nginx/nginx.conf
@@ -159,12 +159,6 @@ http {
         }
 
         location / {
-            # At least for now we don't want Via 3's non-proxy pages (including
-            # the front page) to show in Google. Later (when Via 3 replaces the
-            # public Via service) we'll probably want at least the front page
-            # to be indexable.
-            include includes/robots.conf;
-
             # Proxy to the Gunicorn/Pyramid app.
             # http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass
             proxy_pass http://web;
--- a/tests/unit/via/tweens_test.py
+++ b/tests/unit/via/tweens_test.py
@@ -1,0 +1,39 @@
+from unittest.mock import create_autospec
+
+import pytest
+
+from via.tweens import robots_tween_factory
+
+
+class TestRobotsTween:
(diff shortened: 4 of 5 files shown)


via/views/index.py
```diff
@@ -13,6 +13,8 @@ def get(self):
         if not self.enabled:
             return HTTPNotFound()
 
+        self.request.response.headers["X-Robots-Tag"] = "all"
```
robertknight, line 16: This API is going to get a bit cumbersome if we want to allow other pages to be indexed in future. I think it would be ideal if we could specify a boolean flag on the view somehow (`allow_indexing=True`).

via/tweens.py
```diff
@@ -0,0 +1,10 @@
+def robots_tween_factory(handler, _registry):
```
robertknight, line 1: It would be useful to add a link to the Google docs that explain what `X-Robots-Tag` does and the valid values for it: https://developers.google.com/search/docs/advanced/robots/robots_meta_tag#xrobotstag

robertknight (approved):
This PR does what is expected. The API for an individual view to turn indexing on or off is a bit cumbersome and the meaning may not be obvious to someone who isn't familiar with `X-Robots-Tag`. Right now we only have a single view, so it is not a major issue. If we add more then it would be useful to be able to configure indexing via an `allow_indexing=True` flag in the `@view_config(...)` options for example.

via/views/index.py
seanh: I'm not sure whether it's possible to add a view config that'd be passed to a tween. Might be. Or we might need to change it to a view mapper or something rather than a tween.

This is arguably a Pyramid issue though. The `request.response` object is Pyramid's way for views (and other bits of code that have access to `request`) to communicate things about the desired responses to parts of the code that get run later in the request-processing pipeline (renderers, tweens, etc etc). Pyramid actually does provide convenience attrs for a bunch of common HTTP headers, so you can do `request.response.content_type = "application/json"` for example. See <https://docs.pylonsproject.org/projects/pyramid/en/latest/api/response.html#pyramid.response.Response>. But `X-Robots-Tag` isn't one of the headers covered. And `request.response.headers["Foo"] = "bar"` is Pyramid's interface for setting an arbitrary header. The way you read a setting in Pyramid is similarly a bit awkward (`foo = request.registry.settings["foo"]`) but I wouldn't recommend us developing our own more convenient way to access settings. Having said that, this does suggest that adding our own writeable `request.response.robots` attr would be one way to make this more convenient. I am tempted to just go with the normal Pyramid way here though, as awkward as it is.

via/tweens.py
seanh: :+1: Added a comment

PR #460, opened 2021-04-27
Add X-Robots-Tag header to error pages

seanh:
Also add the X-Robots-Tag header to any error responses served by the `/proxy/static` endpoint. See http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header

### DIFF
--- a/conf/nginx/includes/robots.conf
+++ b/conf/nginx/includes/robots.conf
@@ -2,4 +2,4 @@
 #
 # https://developers.google.com/search/reference/robots_meta_tag
 
-add_header "X-Robots-Tag" "noindex, nofollow";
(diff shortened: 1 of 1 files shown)


robertknight (approved):
To confirm my understanding of the change here: Previously we added `X-Robots-Tag` to these responses if the HTTP status was a success or redirect status (2xx, 3xx). With this PR the header is added for all responses regardless of status.

seanh:
> To confirm my understanding of the change here: Previously we added X-Robots-Tag to these responses if the HTTP status was a success or redirect status (2xx, 3xx). With this PR the header is added for all responses regardless of status.

Yep. I didn't think it was possible in NGINX to add headers to error responses but it turns out you can with `always`.

PR #461, opened 2021-04-27
Factor out an unnecessary network round trip

### LINKED ISSUE #451: Avoid the additional network round trip for legacy Via-style URLs
The `"proxy"` route should do the HTML-or-PDF detection itself and inject the PDF or Via HTML URL directly, instead of using the `"route_by_content"` route. See https://github.com/hypothesis/via3/pull/443#discussion_r615000066

seanh:
Fixes https://github.com/hypothesis/via3/issues/451

Problem
-------

URLs like `https://via.hypothes.is/https://example.com` are served by
the `proxy()` view which renders a page containing
`<iframe src="https://via.hypothes.is/route?url=https://example.com">`.
This iframe URL is served by Via's `route_by_content()` view which checks
whether `https://example.com` returns HTML or PDF and redirects the
iframe to either `https://via.hypothes.is/pdf?url=https://example.com`
or `https://viahtml.hypothes.is/proxy/https://example.com`.

This redirect of the iframe is an unnecessary network round trip.

Solution
--------

Change the `proxy()` view to do the check of whether `https://example.com` is
HTML or PDF itself, exactly as the `route_by_content()` view does, and
inject `https://via.hypothes.is/pdf?url=https://example.com` or
`https://viahtml.hypothes.is/proxy/https://example.com` as the iframe
URL directly instead of going through `route_by_content()`.

This requires some refactoring to move the HTML-or-PDF test out of the
`route_by_content()` view and into `ViaClientService` where it can be
called by both the `route_by_content()` and `proxy()` views.

`route_by_content()` is unchanged
---------------------------------

This PR doesn't change `route_by_content()`'s behavior, it still
redirects as before, it's just that the `proxy()` view no longer uses
it.

Arguably `route_by_content()` isn't needed anymore: the LMS app could
inject the `proxy()` route's
`https://via.hypothes.is/https://example.com` URL as the `src` of its
`iframe`. This would be less code and more consistency between Via in
the LMS and public contexts.

But it would mean an iframe-within-an-iframe in the LMS context and,
more importantly, it might create URL-encoding issues with things like
Canvas files with unusual characters in their filenames that're
important in the LMS context but not in the public context. There were
lots of tricky URL-encoding differences in LMS to do with encodings
being handled differently when the URL is in the path or in a query
string, and differences between Python and NGINX. After much difficulty
we got it all working, so now I don't want to change it.

### DIFF
--- a/tests/unit/via/services/via_client_test.py
+++ b/tests/unit/via/services/via_client_test.py
@@ -2,11 +2,53 @@ from unittest.mock import sentinel
 
 import pytest
 
-from via.services.via_client import factory
+from via.services.via_client import ViaClientService, factory
+
+
+class TestViaClientService:
+    @pytest.mark.parametrize(
+        "mime_type,is_pdf",
+        [
+            ("application/x-pdf", True),
+            ("application/pdf", True),
+            ("text/html", False),
+        ],
+    )
+    def test_is_pdf(self, mime_type, is_pdf, svc):
+        assert svc.is_pdf(mime_type) == is_pdf
+
+    @pytest.mark.parametrize(
+        "mime_type,expected_content_type",
+        [
+            ("application/pdf", "pdf"),
+            ("text/html", "html"),
+            (None, "html"),
+        ],
+    )
+    def test_url_for(self, mime_type, expected_content_type, svc, via_client):
+        params = {"foo": "bar"}
+
+        url = svc.url_for(sentinel.url, mime_type, params)
+
+        via_client.url_for.assert_called_once_with(
+            sentinel.url, expected_content_type, params
+        )
+        assert url == via_client.url_for.return_value
+
+    def test_url_for_pops_blocked_for_from_params(self, svc, via_client):
+        params = {"foo": "bar", "via.blocked_for": sentinel.blocked_for}
+
+        svc.url_for(sentinel.url, sentinel.mime_type, params)
+
+        assert via_client.url_for.call_args[1]["options"] == {"foo": "bar"}
+
+    @pytest.fixture
+    def svc(self, via_client):
+        return ViaClientService(via_client)
 
 
 class TestFactory:
-    def test_it(self, pyramid_request, ViaClientService):
+    def test_it(self, pyramid_request, ViaClient, ViaClientService):
         pyramid_request.registry.settings = {
             "via_html_url": sentinel.via_html_url,
             "via_secret": sentinel.via_secret,
@@ -14,13 +56,24 @@ class TestFactory:
 
         service = factory(sentinel.context, pyramid_request)
 
-        assert service == ViaClientService.return_value
-        ViaClientService.assert_called_once_with(
+        ViaClient.assert_called_once_with(
             service_url=pyramid_request.host_url,
             html_service_url=sentinel.via_html_url,
             secret=sentinel.via_secret,
         )
+        ViaClientService.assert_called_once_with(ViaClient.return_value)
+        assert service == ViaClientService.return_value
 
-    @pytest.fixture
+    @pytest.fixture(autouse=True)
     def ViaClientService(self, patch):
         return patch("via.services.via_client.ViaClientService")
+
+
+@pytest.fixture(autouse=True)
+def ViaClient(patch):
+    return patch("via.services.via_client.ViaClient")
+
+
+@pytest.fixture
+def via_client(ViaClient):
+    return ViaClient.return_value
--- a/tests/unit/via/services/via_client_test.py
+++ b/tests/unit/via/services/via_client_test.py
@@ -35,13 +35,6 @@ class TestViaClientService:
         )
         assert url == via_client.url_for.return_value
 
-    def test_url_for_pops_blocked_for_from_params(self, svc, via_client):
-        params = {"foo": "bar", "via.blocked_for": sentinel.blocked_for}
-
-        svc.url_for(sentinel.url, sentinel.mime_type, params)
-
-        assert via_client.url_for.call_args[1]["options"] == {"foo": "bar"}
-
     @pytest.fixture
     def svc(self, via_client):
         return ViaClientService(via_client)
--- a/tests/unit/via/views/proxy_test.py
+++ b/tests/unit/via/views/proxy_test.py
@@ -1,3 +1,5 @@
+from unittest.mock import sentinel
+
 import pytest
 
 from via.views.proxy import proxy
@@ -10,15 +12,27 @@ class TestProxy:
             ("/https://example.com/foo", "https://example.com/foo"),
             ("/http://example.com/foo", "http://example.com/foo"),
             ("/example.com/foo", "https://example.com/foo"),
+            (
+                "/https://example.com/foo?bar=gar&har=jar#car",
+                "https://example.com/foo?bar=gar&har=jar#car",
+            ),
         ],
     )
-    def test_it(self, pyramid_request, path, url_to_proxy):
+    def test_it(
+        self, pyramid_request, path, url_to_proxy, get_url_details, via_client_service
+    ):
         pyramid_request.path = path
 
         result = proxy(pyramid_request)
 
-        assert result == {
-            "src": pyramid_request.route_url(
-                "route_by_content", _query={"url": url_to_proxy}
-            )
-        }
+        get_url_details.assert_called_once_with(url_to_proxy)
+        via_client_service.url_for.assert_called_once_with(
+            url_to_proxy, sentinel.mime_type, pyramid_request.params
+        )
+        assert result == {"src": via_client_service.url_for.return_value}
+
+    @pytest.fixture(autouse=True)
+    def get_url_details(self, patch):
+        get_url_details = patch("via.views.proxy.get_url_details")
+        get_url_details.return_value = (sentinel.mime_type, sentinel.status_code)
+        return get_url_details
--- a/via/services/via_client.py
+++ b/via/services/via_client.py
@@ -15,13 +15,11 @@ class ViaClientService:
 
     def url_for(self, url, mime_type, params):
         """Return a Via URL for the given `url`."""
-        content_type = "pdf" if self.is_pdf(mime_type) else "html"
-
-        options = dict(params)
-
-        options.pop("via.blocked_for", None)
-
-        return self.via_client.url_for(url, content_type=content_type, options=options)
+        return self.via_client.url_for(
+            url,
+            content_type="pdf" if self.is_pdf(mime_type) else "html",
+            options=params,
+        )
 
 
 def factory(_context, request):
--- a/via/services/via_client.py
+++ b/via/services/via_client.py
@@ -2,19 +2,33 @@
 from h_vialib import ViaClient
 
 
-class ViaClientService(ViaClient):
-    """Place holder for a service, in case we want to diverge."""
+class ViaClientService:
+    """A wrapper service for h_vialib.ViaClient."""
+
+    def __init__(self, via_client):
+        self.via_client = via_client
+
+    @staticmethod
+    def is_pdf(mime_type):
+        """Return True if the given MIME type is a PDF one."""
+        return mime_type in ("application/x-pdf", "application/pdf")
+
+    def url_for(self, url, mime_type, params):
+        """Return a Via URL for the given `url`."""
+        content_type = "pdf" if self.is_pdf(mime_type) else "html"
+
+        options = dict(params)
+
+        options.pop("via.blocked_for", None)
+
+        return self.via_client.url_for(url, content_type=content_type, options=options)
 
 
 def factory(_context, request):
-    """Create a ViaClientService object from the request."""
-
-    settings = request.registry.settings
-
     return ViaClientService(
-        # Where we are going for general / HTML specifically
-        service_url=request.host_url,
-        html_service_url=settings["via_html_url"],
-        # Secret used to sign everything
-        secret=settings["via_secret"],
+        via_client=ViaClient(
+            service_url=request.host_url,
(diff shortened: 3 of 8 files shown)


via/services/via_client.py
```diff
@@ -2,19 +2,34 @@
 from h_vialib import ViaClient
 
 
-class ViaClientService(ViaClient):
-    """Place holder for a service, in case we want to diverge."""
+class ViaClientService:
```
seanh, line 5: I changed `ViaClientService` to wrap a `ViaClient` instance by composition rather than by inheritance.

Generally composition is more flexible and simpler to understand, whereas inheritance tightly couples classes together and makes the code harder to understand. For example I'd have to look up `h_vialib.ViaClient`, in a different project, to know what all of `ViaClientService`'s methods are, or how to construct it. It also makes it impossible to unit test `ViaClientService` in isolation from `ViaClient` (you can't really mock out the superclass). See endless things that have been written about "composition over inheritance".

In this particular case I want `ViaClientService` to have a `url_for()` method that takes different arguments than `ViaClient.url_for()`. This causes PyLint to complain, rightly, about a method being overridden with different parameters. When subclassing, an instance of a subclass should be able to be used wherever an instance of the superclass is expected, so subclasses shouldn't change the parameters of superclass methods when overriding. But that puts me in a difficult position, I'd have to think of a different name for my method, other than the name I ideally want. There's no reason why we need to use inheritance here (we don't need polymorphism) so I just removed it.

Note that the methods of the underlying `ViaClient` instance are still accessible as `via_client_svc.via_client.*` (though we never need to access them anyway)

robertknight: 👍

via/services/via_client.py
```diff
@@ -2,19 +2,34 @@
...
 
+    def __init__(self, via_client):
+        self.via_client = via_client
 
-def factory(_context, request):
-    """Create a ViaClientService object from the request."""
+    @staticmethod
+    def is_pdf(mime_type):
+        """Return True if the given MIME type is a PDF one."""
+        return mime_type in ("application/x-pdf", "application/pdf")
```
robertknight, line 14: It looks like we're not doing any kind of MIME type sniffing on the content at present. If so, I expect at some point we'll run into an issue with a PDF that is served without a suitable header which will end up redirecting to ViaHTML.

via/services/via_client.py
```diff
@@ -2,19 +2,34 @@
...
+
+    def url_for(self, url, mime_type, params):
+        """Return a Via URL for the given `url`."""
+        content_type = "pdf" if self.is_pdf(mime_type) else "html"
+
+        options = dict(params)
 
-    settings = request.registry.settings
+        options.pop("url", None)
+        options.pop("via.blocked_for", None)
```
robertknight: From looking across repos it looks like the `blocked_for` param indicates whether it is the LMS or public Via service is requesting the URL?

Is there a reason only this particular Via param is dropped? What about other `via.`-prefixed query params?

via/services/via_client.py
```diff
@@ -2,19 +2,34 @@
...
+        return mime_type in ("application/x-pdf", "application/pdf")
+
+    def url_for(self, url, mime_type, params):
+        """Return a Via URL for the given `url`."""
+        content_type = "pdf" if self.is_pdf(mime_type) else "html"
+
+        options = dict(params)
 
-    settings = request.registry.settings
+        options.pop("url", None)
```
robertknight: Shouldn't the responsibility for dropping the `url` param live in the `route_by_content` URL? Do both the `proxy` and `route_by_content` views accept this param?

tests/unit/via/services/via_client_test.py
```diff
@@ -2,25 +2,81 @@
...
+        assert svc.is_pdf(mime_type) == is_pdf
+
+    @pytest.mark.parametrize(
+        "mime_type,expected_content_type",
+        [
+            ("application/pdf", "pdf"),
+            ("text/html", "html"),
+        ],
+    )
+    def test_url_for(self, mime_type, expected_content_type, svc, via_client):
```
robertknight, line 28: Can the `mime_type` passed into `url_for` be empty or `None`?

tests/unit/via/services/via_client_test.py
```diff
@@ -2,25 +2,81 @@
...
+            sentinel.url, expected_content_type, params
+        )
+        assert url == via_client.url_for.return_value
+
+    def test_url_for_pops_url_and_blocked_for_from_params(self, svc, via_client):
+        params = {
+            "foo": "bar",
+            "url": sentinel.url,
+            "via.blocked_for": sentinel.blocked_for,
+        }
```
robertknight: I had a query about where the `url` param popping should live and why only the `blocked_for` via param is popped but not others.

tests/unit/via/services/via_client_test.py
```diff
@@ -2,25 +2,81 @@
...
 
-    @pytest.fixture
+    @pytest.fixture(autouse=True)
     def ViaClientService(self, patch):
         return patch("via.services.via_client.ViaClientService")
+
+
+@pytest.fixture(autouse=True)
+def ViaClient(patch):
+    return patch("via.services.via_client.ViaClient")
```
robertknight, line 67: Should this use a mock with autospec?

tests/unit/via/views/proxy_test.py
```diff
@@ -12,13 +14,21 @@ class TestProxy:
             ("/example.com/foo", "https://example.com/foo"),
         ],
     )
-    def test_it(self, pyramid_request, path, url_to_proxy):
```
robertknight, line 15: I think it would be a good idea to add some examples of URLs with query params and fragments here.

robertknight (commented):
I first look through this. Generally the change makes sense. I had some queries about tests, query params and how missing or empty MIME types are handled.

via/services/via_client.py
seanh: Same as it has always been with Via 3. I think this is the same in legacy Via too, isn't it?

I dunno if we'd _want_ to do content sniffing either as that'd mean downloading part of every file (that couldn't already be identified as PDF from a header) before redirecting and then downloading the file from the start again.

There might be something we can do about this edge case in Via HTML. If Via HTML starts downloading a file and realises that it's PDF then it could redirect to Via 3's PDF endpoint. Adds an extra redirect, but only in this edge case, and it's better than the file not working at all

via/services/via_client.py
seanh: This could be moved into `route_by_content()`, yeah

tests/unit/via/services/via_client_test.py
seanh: Looks like it can be, I'll add a test for that case

tests/unit/via/services/via_client_test.py
seanh: `patch()` does autospeccing automatically

tests/unit/via/views/proxy_test.py
seanh: Added an extra test case

via/services/via_client.py
seanh: Done

via/services/via_client.py
seanh: I looked into how the `blocked_for` param works.

TLDR: Via 3 should not be popping this param.

Details:

1. What this param ultimately does is change what text appears on Checkmate's block pages. If you include the `blocked_for` param when calling `checkmatelib`'s `check_url()` method then the block page shown at the returned presentation URL will either suggest download the Chrome extension (in the public context) or not (in the LMS context).

1. LMS includes the `via.blocked_for=lms` param when redirecting to Via 3: 

   https://github.com/hypothesis/lms/blob/2f23c04fd0d9a1ba41dab3034c752684e223389b/lms/views/helpers/_via.py#L21-L32

2. Via 3 forwards the `via.blocked_for` param when calling Checkmate's API:

   https://github.com/hypothesis/via3/blob/1c483048d98e8999740ced01fadb408796ca2ac4/via/views/decorators.py#L19

3. Via 3 does not forward the `via.blocked_for` param when redirecting to Via HTML:

   https://github.com/hypothesis/via3/blob/ce2a664bba8d00af48c799b5c07c8ecdbb5cef31/via/views/route_by_content.py#L26-L35

4. Via HTML reads the `via.blocked_for` param from a query param:

   https://github.com/hypothesis/viahtml/blob/b6bd9b4f5b005a1824452ce0863dc08b2e78a88f/viahtml/views/security.py#L30-L32

   This looks wrong: Via HTML seems to be expecting `via.blocked_for` in a query param but Via 3 explicitly _avoids_ forwarding that query param to Via HTML with a slightly mysterious comment saying `# Already passed this to checkmate. Viahtml will decide a new value if any.` LMS's QA and prod instances of Via HTML do not have any environment variable relating to this param, so it looks like Via HTML is definitely expecting to receive `blocked_for` in a query param. The comment and the popping of the `via.blocked_for` param comes from this PR: https://github.com/hypothesis/via3/pull/402. There's some discussion on that PR that the param should _not_ be popped, but it looks like the code was not changed to stop popping it.

   In practice I don't think this usually causes a problem. If a request is going to be blocked by Checkmate then it'll be blocked when Via 3 calls Checkmate and the redirect to Via HTML will never happen. If a request is _not_ blocked by Checkmate then Via 3 pops the `via.blocked_for` and Via HTML incorrectly fails to include `blocked_for` when calling Checkmate but, since Via 3 already did the same Checkmate check and it passed, the check is gonna pass when Via HTML does it too so the missing `blocked_for` param doesn't matter.

   There _is_ an edge-case where this can matter though. If Via 3's Checkmate check fails to get a response (say because of a network blip) but Via HTML's Checkmate check succeeds and returns a block response then the missing `blocked_for` param results in Checkmate's block page incorrectly suggesting that the user install the Chrome extension in the LMS context. You can test this by:

   1. Making sure you have Checkmate running locally

   2. Breaking your local Via 3's Checkmate URL:
   
      ```diff
      diff --git a/tox.ini b/tox.ini
      index ef71f47..8ad8733 100644
      --- a/tox.ini
      +++ b/tox.ini
      @@ -26,7 +26,7 @@ setenv =
           dev: NEW_RELIC_APP_NAME = {env:NEW_RELIC_APP_NAME:via3}
           dev: NEW_RELIC_ENVIRONMENT = {env:NEW_RELIC_ENVIRONMENT:dev}
           dev: SENTRY_ENVIRONMENT = {env:SENTRY_ENVIRONMENT:dev}
      -    dev: CHECKMATE_URL = http://localhost:9099
      +    dev: CHECKMATE_URL = http://localhost:9999
           dev: CHECKMATE_IGNORE_REASONS = {env:CHECKMATE_IGNORE_REASONS:publisher-blocked}
           dev: CHECKMATE_ALLOW_ALL = {env:CHECKMATE_ALLOW_ALL:true}
           dev: NGINX_SECURE_LINK_SECRET = not_a_secret
      ```

   3. Visiting [localhost (make devdata) Blocked Media HTML Assignment](https://hypothesis.instructure.com/courses/125/assignments/1523)

      Screenshot from 2021-04-28 12-27-04

Of course there is an obvious optimisation here: if Via 3 has already checked the URL then Via HTML doesn't need to check it at all. But we haven't implemented that optimisation. Until we do it's not correct for Via 3 to pop the `blocked_for` param. To implement the optimisation we'd need a secure way for Via 3 to identify itself when redirecting to Via HTML in order to say "don't bother calling Checkmate". It can't be just based on `Referer` as that would enable `curl` scripts and the like to bypass Checkmate's blocklist (not just the allow-list), which they can't do currently.

seanh: There was some discussion of this in Slack: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1619605700198300

seanh: Removed the popping: https://github.com/hypothesis/via3/pull/461/commits/dd38991c6ee1509d52a4acfa4c6b608ea99a749a

via/services/via_client.py
robertknight: > I dunno if we'd want to do content sniffing either as that'd mean downloading part of every file (that couldn't already be identified as PDF from a header) 

There is a fairly simple and standard MIME sniffing test for PDFs: Look for the exact string `%PDF` in the first 1024 bytes of the file. From past experience it was necessary to implement this in custom PDF viewers in order to match Adobe Acrobat's behavior. This is out of scope for this PR though.

via/services/via_client.py
robertknight: > To implement the optimisation we'd need a secure way for Via 3 to identify itself when redirecting to Via HTML in order to say "don't bother calling Checkmate"

A lesser but perhaps simpler optimization, depending on how much caching Checkmate already implements, is to just make sure Checkmate is as fast as possible for recently queried URLs.

seanh: Ironically the Via's call Checkmate on an internal checkmate.svc.hypothes.is URL that bypasses Cloudflare because we think that should be faster than going over the internet. But that means that there's no caching because we don't have any internal cache set up between Via and Checkmate. I don't think Cloudflare's cache would be of use here anyway though: you wouldn't normally cache these (JSON API requests that include an authentication token). I think in theory they can be cached with appropriate sharding. For example perhaps `Vary` on the `Authorization` header (which would require the different Via's to share the same API token to be effective, they don't currently). I don't think this could work with Cloudflare which doesn't support `Vary`. Alternatively we could decide to make the API public and remove the authentication from it, then (if we were also happy for Via's Checkmate requests to go over the internet) we could use Cloudflare's caching.

In practice I think Checkmate is very fast without caching

robertknight (approved):
Changes LGTM. I tested this locally and the iframe URL now refers to `/proxy` rather than `/route_by_content`.

PR #462, opened 2021-04-27
Add abuse policy link to front page

seanh:
Part of https://github.com/hypothesis/checkmate/issues/70.

The plan is to add it to Via 3 and Via HTML response headers as well, separate PRs coming for that.

### DIFF
--- a/via/templates/index.html.jinja2
+++ b/via/templates/index.html.jinja2
@@ -99,6 +99,7 @@
         <a href="https://web.hypothes.is/via-request/">ask us to add it to the allowed list.</a>
         See our
         <a href="https://web.hypothes.is/help/what-is-the-via-proxy/">help article</a>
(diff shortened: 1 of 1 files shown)


robertknight (approved):
LGTM. The typography on this page could do with some refinement, but we can address that separately.

seanh:
> The typography on this page could do with some refinement, but we can address that separately.

Yeah. Personally I think it's way too much text. I'd prefer a link or button to reveal this text, something like:

> [Paste a link to annotate]
>
> <kbd>Annotate</kbd>
>
> <details>
> <summary>Abuse Policy</summary>
> Via is a closed proxy that lets you annotate web pages and PDFs with Hypothesis. Only a small number of known sites can be annotated. If you'd like to annotate a site that isn't allowed please ask us to add it to the allowed list. See our help article for more information.
> </details>

PR #465, opened 2021-04-28
Show loading indicator while ViaHTML iframe is loading

### LINKED ISSUE #155: Indicate when the proxied page is loading
When Via HTML is loading a third-party page in an iframe it should display some sort of "loading" placeholder in place of the iframe while the third-party page is still loading. See: https://github.com/hypothesis/via3/pull/443#pullrequestreview-638734311

robertknight:
In Safari there is no indicator that anything is loading once the main
frame has loaded, the user just sees a blank white space where the
iframe is.

To provide a better experience add a loading _bar at the top of the page_. This is shown until the iframe reports the document metadata to the parent frame for the first time, corresponding
to the `DOMContentLoaded` event having been received in the iframe. This
is also when the tab title gets set. (_Edited: The spinner has been changed to a loading bar_)

~~The loading spinner HTML and SVG have been taken from the
`annotation.html.jinja2`, `bouncer.css` and `hypothesis-icon.svg` files
in the hypothesis/bouncer repository. All of the resources have been
inlined here, which optimizes loading a little, but we could quite
easily move them to external resources in future for maintainability.~~ (_Edit: The spinner was replaced with a red bar._)

Fixes https://github.com/hypothesis/viahtml/issues/155

### DIFF
--- a/via/templates/proxy.html.jinja2
+++ b/via/templates/proxy.html.jinja2
@@ -68,6 +68,17 @@
         transform: translate(-3px, -3px) rotate(359deg);
       }
     }
+
+    .js-loading-indicator {
+      opacity: 0;
+      transition: opacity 0.2s;
+    }
+    .js-loading-indicator.is-loading {
+      opacity: 1.0;
+    }
+    .js-loading-indicator.is-done {
+      opacity: 0.0;
+    }
   </style>
 
   <script>
@@ -92,7 +103,7 @@
   // metadata etc. in the current frame.
   window.addEventListener('message', event => {
     const contentFrame = document.querySelector('.js-content-frame');
-    const loadingIndicator = document.querySelector('.js-loading-spinner');
+    const loadingIndicator = document.querySelector('.js-loading-indicator');
 
     if (event.source !== contentFrame.contentWindow) {
       return;
@@ -107,7 +118,7 @@
       // Treat the first `metadatachange` message as a hint that the iframe is
       // done loading. This event is sent in response to `DOMContentLoaded` within the
       // ViaHTML iframe, so sub-resources on the page may still be loading.
-      loadingIndicator.style.display = 'none';
+      loadingIndicator.classList.add('is-done');
 
       document.title = `Via: ${message.metadata.title}`;
     }
@@ -123,7 +134,7 @@
        visible while the iframe loads. In Safari however there is no indication that anything is
        loading once the main frame has loaded.
     #}
-    <div class="js-loading-spinner">
+    <div class="js-loading-indicator">
       <div class="center spinner__stationary-ring">
         <svg alt="Hypothesis logo" class="center spinner_icon" width="24" height="28" viewBox="0 0 24 28">
           <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
@@ -133,5 +144,16 @@
         <div class="spinner__moving-ring"></div>
       </div>
     </div>
+    <script>
+      /**
+       * Show the loading spinner after a short delay. The delay avoids a flash
+       * of the indicator if the content loads very quickly.
+       */
+      function showLoadingIndicator() {
+        const loadingIndicator = document.querySelector('.js-loading-indicator');
+        setTimeout(() => loadingIndicator.classList.add('is-loading'), 500);
+      }
+      showLoadingIndicator();
+    </script>
   </body>
 </html>
--- a/via/templates/proxy.html.jinja2
+++ b/via/templates/proxy.html.jinja2
@@ -23,6 +23,51 @@
       margin: 0;
       padding: 0;
     }
+
+    .center {
+      position: absolute;
+      left: 50%;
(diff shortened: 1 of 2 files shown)


seanh:
A couple of issues in testing this:

It fails to hide the spinner when loading a PDF rather than an HTML page. The background colour behind the spinner is also wrong when loading a PDF (it's set by PDF.js):

Peek 2021-04-28 17-20

I think the background colour could also be an issue with HTML pages that set a non-white background.

For HTML pages also I notice that the spinner is still shown for a second on top of the page contents once they become visible:

Peek 2021-04-28 17-19

seanh:
Instead of showing a spinner in front of the third-party page or PDF, I wonder if a better approach might be to show a loading bar _above_ it? Either at the very top of the iframe, or above the iframe. Some sites show a simple blue line, a few pixels high, at the top of the page that moves from left to right as the page loads. The advantage of this is no clashing between our spinner and the site or PDF behind it. GitHub does this for example:

Peek 2021-04-28 17-24

The problem of course is that using a progress bar rather than a spinner requires the code to have some at least rough idea of _how much_ of the site has loaded, not just that it's still loading. There are bar-shaped loading indicators that simply use animation to indicate that "something is happening" without indicating how much progress has been made, this sort of thing:

Peek 2021-04-28 17-26

robertknight:
I have pushed to changes to address the issues above:

- The loading indicator no longer lingers in the PDF viewer. The fix is to make the PDF viewer iframe send the same document metadata message to the parent that a ViaHTML frame does. This also fixes the tab title in the PDF viewer.
- The display of the loading indicator has been delayed by .5s so that it doesn't flicker briefly if the content loads quickly
- The loading indicator now fades in and out to reduce the flicker effect

Regarding the loading indicator style, I considered a GitHub-style progress bar but I thought this might be too hidden on the initial page load. On GitHub this indicator is shown when transitioning from one GH page to another so there isn't the "blank slate" problem on first load. I think we'd have to try some different variants to be sure.

robertknight:
An issue that is not addressed here is what to do if the Via 3 request succeeds and renders the wrapper page containing an iframe, but the ViaHTML page fails to load for any reason or renders a page without the `banner_insert.html` header. From what I can find so far, there is no event that gets emitted if the iframe fails to load for any reason. If that is the case then the best we can do is to add a timeout.

Edit: Related issue about this: https://github.com/whatwg/html/issues/125. Our use case is very similar to the ones in the comments that mention using `postMessage` to communicate with a parent frame.

seanh:
Works with PDFs now :+1:

There are examples of the "bar at the top of the frame" design being used in initial page loads. PDF.js does it for example (sorry long/slow GIF, wait for it):

Peek 2021-04-29 17-44

The problem is that, in Safari, the user can be faced with a blank white window and nothing happening, right? If the _only_ thing in the blank white window is a loading bar at the top, the user will notice it.

I really think that showing a spinner with transparent elements over arbitrary third-party content is a problem. It creates visual clashing and legibility issues that will vary depending on the third-party site that's behind it, and it just sort of looks like broken CSS. Example:

Screenshot from 2021-04-29 17-39-49

Here's what it'll look like if the third-party site uses the same colour as we do:

Screenshot from 2021-04-29 17-47-42

seanh:
Maybe put another way: there's a reason why browsers show a tiny loading spinner up on the tab (where they have control of the background colour and surroundings), rather than a big one in the middle of the window

seanh:
@robertknight Do you want to push ahead with this design, or try out some variants? And do you want to do anything about the case when the iframe's contents fail to load?

robertknight:
I have changed the loading spinner to a solid bar that is similar to GitHub's, but using our brand color instead of blue and is slightly thicker to make it more visible. This avoids an issue where the spinner can obscure the page content if it takes a long time in between some content appearing and the `DOMContentLoaded` event being fired in the page. New version:

https://user-images.githubusercontent.com/2458/116975615-d5bf2a80-acb7-11eb-8d0e-dae1a166f2bc.mov

As with GitHub's bar, this one doesn't use an indeterminate state during the initial load. Instead it just grows to a fixed progress percentage and stops until loading is "done" at which point it grows to 100%.

To get a sense of what it looks like under different conditions, open Chrome devtools, go to the Network tab and enable connection throttling at various levels.

via/templates/proxy.html.jinja2
```diff
@@ -23,6 +23,77 @@
...
+        opacity: 1.0;
+      }
+    }
+
+    /* Once the content has loaded, set progress to 100% and then fade out the
+       progress bar.
+    */
+    .loading-bar.is-done {
+      animation: loading-bar-finish;
+      animation-duration: 1s;
```
robertknight, line 71: The animation is not actually `1s` long as it reaches the final state at 70%. Using 1s makes each 10% of the animation take 100ms, which is convenient.

seanh (approved):
Looks great!

PR #467, opened 2021-04-28
Add abuse policy NGINX request headers

seanh:
This adds X-Abuse-Policy and X-Complaints-To headers to the requests
that Via 3's NGINX sends to third-party servers when proxying their PDF
files.

This does not add the headers to the requests that the Python app sends
to determine with a URL is PDF or HTML, separate PR for that.

via_test_app contains a Python view that returns a PDF file and I was
able to use the view to verify that the request headers are received.

See this Slack thread for the choice of headers:

https://hypothes-is.slack.com/archives/C4K6M7P5E/p1619616535474700?thread_ts=1619615485.468500&cid=C4K6M7P5E

Part of hypothesis/checkmate#70

### DIFF
--- a/conf/nginx/includes/proxy.conf
+++ b/conf/nginx/includes/proxy.conf
@@ -11,6 +11,12 @@
 # http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_server_name
 proxy_ssl_server_name on;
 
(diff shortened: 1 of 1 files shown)


seanh:
Closing in favour of https://github.com/hypothesis/via3/pull/468

PR #468, opened 2021-04-28
Add abuse policy headers

seanh:
This adds `X-Abuse-Policy` and `X-Complaints-To` headers to:

* The requests that Via 3's Python app sends to third-party servers when testing whether a URL is HTML or PDF
* The requests that Via 3's NGINX sends to third-party servers when proxying PDF files
* All Via 3's responses (both the ones that're proxying third-party PDF files and all the non-proxy responses)

I was able to use via_test_app to verify that the request headers are received.

Part of https://github.com/hypothesis/checkmate/issues/70

See this Slack thread for the choice of headers: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1619616535474700?thread_ts=1619615485.468500&cid=C4K6M7P5E

### DIFF
--- a/via/get_url/details.py
+++ b/via/get_url/details.py
@@ -1,6 +1,7 @@
 """Retrieve details about a resource at a URL."""
 import cgi
 import re
+from collections import OrderedDict
 from functools import wraps
 
 import requests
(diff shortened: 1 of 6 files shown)


via/get_url/details.py
```diff
@@ -53,16 +54,20 @@ def get_url_details(url, headers=None):
     :raise UnhandledException: For all other request based errors
     """
     if headers is None:
-        headers = {}
+        headers = OrderedDict()
```
seanh, line 57: PyLint complained that the `headers = clean_headers(headers)` below was redefining the type of `headers` from `dict` to `OrderedDict`, so I had to make it an `OrderedDict` in the first place

via/get_url/details.py
```diff
@@ -53,16 +54,20 @@ def get_url_details(url, headers=None):
...
 
     if GOOGLE_DRIVE_REGEX.match(url):
         return "application/pdf", 200
 
+    headers = clean_headers(headers)
+    headers["X-Abuse-Policy"] = "https://web.hypothes.is/abuse-policy/"
+    headers["X-Complaints-To"] = "https://web.hypothes.is/report-abuse/"
```
jon-betts, line 65:
```suggestion
    # Pass our abuse policy in request headers for third-party site admins.
    headers["X-Abuse-Policy"] = "https://web.hypothes.is/abuse-policy/"
    headers["X-Complaints-To"] = "https://web.hypothes.is/report-abuse/"
```

via/get_url/details.py
jon-betts: Actually as of 3.7 dict iteration order is guaranteed to be in insertion order. So we can use our newfound newness and remove the `OrderedDict` from the function if we like!

jon-betts (approved):
Yep, this looks good. Good use of NGINX to get most of this done.

There's one suggestion that we can unleash the full power of Python 3.7 to the `OrderedDict` issue.

PR #473, opened 2021-05-04
Fix direct links in new Via implementation

### LINKED ISSUE #472: Direct links not working with the new Via
Original context: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1620139294309300?thread_ts=1620137299.302400&cid=C1MA4E9B9

Direct links to annotations are not working properly with the new Via. Via loads the page and embeds the client, but the client does not scroll to the annotation or group indicated by the `#annotations:` fragment.

The issue is that the `#annotations` fragment is not seen by the server-side part of Via 3 and is not included in the URL of the iframe that Via 3 renders.

----

**Steps to reproduce:**

In a browser without the Hypothesis Chrome extension installed:

1. Go to https://hypothes.is/search and expand one of the document items for an HTML (not PDF) document
2. Click on the "View annotations in context" link

**Expected result:**

The hyp.is service directs the user to Via, the client loads and scrolls to the annotation.

**Actual result:**

The hyp.is service directs the user to Via, the client loads but does not scroll to the annotation.

robertknight:
Direct links were not working in Via 3 because the `#annotations:` URL
fragment added to the Via URL by the bouncer service is not propagated
to the iframe and hence were never seen by the client.

This commit fixes the issue by using JS to update the fragment of the
iframe URL. This change assumes that:

 - The script will always run before the client has been able to load. I
   believe this is a safe assumption.

 - It is OK to blindly copy all fragments to the child iframe without
   any filtering. This has a benefit that it enables regular anchors in
   the page to work.

Fixes https://github.com/hypothesis/via3/issues/472

----

**Testing:**

1. Navigate to http://localhost:9083/example.com and make an annotation
2. Click on the annotation's timestamp and copy the annotation ID from the URL
3. Navigate to http://localhost:9083/example.com#annotations:<annotation ID>

This should load example.com and select the specified annotation in the client. The same steps should also work with a PDF (eg. https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf)

### DIFF
--- a/via/templates/proxy.html.jinja2
+++ b/via/templates/proxy.html.jinja2
@@ -160,7 +160,23 @@
         const loadingIndicator = document.querySelector('.js-loading-indicator');
         setTimeout(() => loadingIndicator.classList.add('is-loading'), 500);
       }
+
+      /**
+       * Copy the fragment from the top frame's URL to the iframe's URL.
+       *
+       * This is needed so that direct links to annotations, which use a
+       * `#annotations:` fragment, are respected by the client in the iframe.
+       */
+      function copyFragmentToFrame() {
+        const fragment = location.hash;
+        if (fragment) {
+          const iframe = document.querySelector('.js-content-frame');
+          iframe.src += fragment;
+        }
+      }
+
       showLoadingIndicator();
+      copyFragmentToFrame();
     </script>
   </body>
 </html>


seanh (approved):

via/templates/proxy.html.jinja2
```diff
@@ -160,7 +160,23 @@
...
+      function copyFragmentToFrame() {
+        const fragment = location.hash;
+        if (fragment) {
+          const iframe = document.querySelector('.js-content-frame');
+          iframe.src += fragment;
+        }
+      }
+
       showLoadingIndicator();
+      copyFragmentToFrame();
```
robertknight, line 179: Something I'm not doing here is copying changes to the fragment to the iframe _after_ the page loads. That's not needed to solve this particular problem.

PR #474, opened 2021-05-04
Prevent the browser extension from launching in Via

### LINKED ISSUE #471: Prevent browser extension from activating itself on Via 3 pages
Original context: https://hypothes-is.slack.com/archives/C1MA4E9B9/p1620137299302400

If a user follows a link to a Via page with a `#annotations:<ID>` fragment in the URL, this can trigger the browser extension to open. The extension won't realize that the page already includes the client because the content is inside an iframe and the extension only checks for the presence of the client in the top page.

seanh:
Fixes https://github.com/hypothesis/via3/issues/471

### DIFF
--- a/via/templates/proxy.html.jinja2
+++ b/via/templates/proxy.html.jinja2
@@ -5,6 +5,10 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>via.hypothes.is</title>
   <link rel="icon" href="favicon.ico" type="image/x-icon" />
+
+  {# Prevent the browser extension from injecting a second sidebar into Via pages. -#}
+  <link rel="sidebar" type="application/annotator+html">
+
   <style>
     body {
       margin: 0;


via/templates/proxy.html.jinja2
```diff
@@ -5,6 +5,10 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>via.hypothes.is</title>
   <link rel="icon" href="favicon.ico" type="image/x-icon" />
+
+  {# Prevent the browser extension from injecting a second sidebar into Via pages. -#}
+  <link rel="sidebar" type="application/annotator+html">
```
robertknight, line 10: It took me a while to find out where the code is that stops this injection. It is actually in the client in https://github.com/hypothesis/client/blob/626721051e11d846ca4a9b6a78a35b1db3eb890b/src/boot/boot.js#L142. Currently only the `type` attribute is considered, `rel` is ignored.

robertknight (approved):
This change prevents double injection of the sidebar. Due to an existing issue with the interaction between the browser extension and client it leaves the extension icons in an active state. I suggest adding a note that it is the client that currently checks for the presence of `link[type="application/annotator+html"]` and aborts startup if found.

PR #483, opened 2021-05-06
Fix query string and leading/trailing whitespace handling on the index page

### LINKED ISSUE #479: Trailing spaces in input field not trimmed
**Steps to reproduce:**

1. Go to the Via homepage
2. Enter a PDF URL followed by a trailing space (eg: "https://arxiv.org/pdf/2105.01649 ")
3. Click "Annotate"

**Expected result:**

The trailing space is ignored and does not appear in the URL shown in the URL bar or the URL that is used in annotations

**Actual result:**

The trailing space is URL encoded as `%20` and appears both in the URL and in URL shown in the document.

URL bar space

PDF URL space

### LINKED ISSUE #482: URL encoding issue prevents PDF from showing in Via3
**Describe the bug**
Via is failing to display a PDF at https://library.educause.edu/-/media/files/library/2021/4/2021hrteachinglearning.pdf?la=en&hash=C9DEC12398593F297CC634409DFF4B8C5A60B36E, but works (intermittently?) for the URL https://library.educause.edu/-/media/files/library/2021/4/2021hrteachinglearning.pdf.

The former link in Via causes the site to give an error:
> We’re sorry. Something went wrong.
> Our site is experiencing server issues or the page is configured incorrectly.

Per our Slack convo:
> The URL with query string params clearly returns a PDF, but Via3 is entering HTML proxying mode and trying to pass the wrong URL to the ViaHTML service.

>I can reproduce. Looks like a URL encoding issue somewhere. Via 3 ends up trying to check the MIME type of the wrong URL (https://library.educause.edu/-/media/files/library/2021/4/2021hrteachinglearning.pdf%3Fla=en&hash=C9DEC12398593F297CC634409DFF4B8C5A60B36E instead of https://library.educause.edu/-/media/files/library/2021/4/2021hrteachinglearning.pdf?la=en&hash=C9DEC12398593F297CC634409DFF4B8C5A60B36E) and gets the wrong result.

> The URLs above are the value of the url as seen by the proxy view in Via 3 when it checks the MIME type of the URL. Note the ? of the query string has been encoded to %3F

**To Reproduce**
1. Load  https://library.educause.edu/-/media/files/library/2021/4/2021hrteachinglearning.pdf. in Via
2. Load https://library.educause.edu/-/media/files/library/2021/4/2021hrteachinglearning.pdf?la=en&hash=C9DEC12398593F297CC634409DFF4B8C5A60B36E in Via
3. See error.

**Additional context**
https://hypothes-is.slack.com/archives/C2BLQDKHA/p1620223234407700
https://app.hubspot.com/contacts/6291320/ticket/401408913/

robertknight:
This PR fixes two issues on the index page:

- Leading or trailing whitespace in the input URL would cause an error when submitting the form
- Query string params in the URL on the index page were not handled correctly when generating the redirect URL: Submitting the URL `https://example.com?param=value` resulted in a redirect to `/https://example.com%3Fparam=value` instead of `/https://example.com?param=value`

Changes in this PR:

- Introduce a `url_from_user_input` helper in `via.views._helpers` that takes URLs from a "user input source" (eg. the `<URL>` part of a manually entered `https://via.hypothes.is/<URL>` input in URL bar, or the URL entered into the input box on the index page) and returns a normalized version that is suitable for passing to `requests` or ViaHTML. The normalized version strips leading/trailing whitespace and adds an `https://` protocol if missing
- On the index page, feed the input URL into `url_from_user_input` and then parse the result to separate out the parts before and after the query. The parts before are passed as the `url` path param when generating the redirect URL for the `proxy` view and the query params are passed as the `_query` to `route_url`

A side effect of these changes is that submitting `example.com` on the homepage will now redirect to `/https://example.com` instead of `/example.com` because the redirect URL is the normalized version.

Fixes https://github.com/hypothesis/via3/issues/482
Fixes https://github.com/hypothesis/via3/issues/479

### DIFF
--- a/tests/unit/via/views/index_test.py
+++ b/tests/unit/via/views/index_test.py
@@ -1,7 +1,8 @@
 import pytest
-from pyramid.httpexceptions import HTTPNotFound
+from pyramid.httpexceptions import HTTPFound, HTTPNotFound
 
 from tests.unit.matchers import temporary_redirect_to
+from via.views.exceptions import BadURL
 from via.views.index import IndexViews
 
 
@@ -18,6 +19,28 @@ class TestIndexViews:
             pyramid_request.route_url(route_name="proxy", url=url)
         )
 
+    def test_post_normalizes_input(self, views, pyramid_request, url_from_user_input):
+        url = pyramid_request.params["url"] = "example.com"
+        url_from_user_input.side_effect = ["https://normalized.com"]
+
+        redirect = views.post()
+
+        url_from_user_input.assert_called_with(url)
+        assert redirect == temporary_redirect_to(
+            pyramid_request.route_url(route_name="proxy", url="https://normalized.com")
+        )
+
+    def test_post_url_with_query_params(self, views, pyramid_request):
+        pyramid_request.params["url"] = "https://site.org?q1=value1&q2=value2"
+
+        redirect = views.post()
+
+        assert isinstance(redirect, HTTPFound)
+        assert (
+            redirect.location
+            == "http://example.com/https://site.org?q1=value1&q2=value2"
+        )
+
     def test_post_with_no_url(self, views, pyramid_request):
         assert "url" not in pyramid_request.params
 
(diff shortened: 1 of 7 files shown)


tests/unit/via/views/index_test.py
```diff
@@ -18,6 +19,28 @@ def test_post(self, views, pyramid_request):
             pyramid_request.route_url(route_name="proxy", url=url)
         )
 
+    def test_post_normalizes_input(self, views, pyramid_request, url_from_user_input):
+        url = pyramid_request.params["url"] = "example.com"
+        url_from_user_input.side_effect = ["https://normalized.com"]
```
seanh, line 24: Do you just want to set `.return_value` here?

seanh (approved):

tests/unit/via/views/index_test.py
robertknight: `side_effect` needs to be used here in order to override the `side_effect` from the fixture. If both `side_effect` and `return_value` are set, it seems that `side_effect` takes precedence.

PR #492, opened 2021-05-18
Remove references to via3

seanh:
Remove most references to "via3", "Via 3" or "viahtml3", renaming them to just via, Via and viahtml.

There are still a couple of references to "via3" in Makefile and Jenkinsfile as these have to do with pushing to Docker Hub. Changes to these need to be coordinated with other changes so they'll be done in a separate PR (https://github.com/hypothesis/via/pull/493).


jon-betts (approved):
Yep, the only ones I could spot are in the mentioned exceptions

Issue #543, opened 2021-07-28
Call Checkmate before sending `GET` requests to third-party servers


seanh:
Slack thread: https://hypothes-is.slack.com/archives/C074BUPEG/p1627378641012800?thread_ts=1627359091.009000&cid=C074BUPEG

### Problem

When a public Via URL like <https://via.hypothes.is/https://example.com> is visited Via [sends a `GET` request to example.com to determine whether the URL is HTML or PDF](https://github.com/hypothesis/via/blob/d967aef42a38f3b8fa68c2e2b77e2821f84e4634/via/views/proxy.py#L13) and it looks like it _doesn't_ check whether the URL is blocked by Checkmate before sending this request.

As a result blocked websites that are slow to respond to this `GET` can cause performance issues for Via.

### Solution

Via should call Checkmate before doing this `GET` request.

This may result in multiple Checkmate checks for a single URL if Checkmate gets checked again by Via's PDF viewer or by Via HTML. This is probably fine: Checkmate requests are fast. We probably don't need to optimize it.

If it turns out that we _do_ need to optimize this we can probably do so based on the `Referer` header (Via and Via HTML could know that Checkmate has already been checked if the `Refer` points to Via). See the Slack thread (linked above) for more details. But don't implement this optimisation unless necessary.

## PR

 * https://github.com/hypothesis/via/pull/552

jon-betts:
I _think_ this is working now, although it's a bit hard to tell. I'm seeing wacky behavior with the web archive save URL we had, but it's not hanging any more.

It just redirects back to the index page?

Other pages appear to be blocked as we'd expect.

PR #552, opened 2021-08-05
Check URLs with Checkmate before retrieving them on the proxy end-point

jon-betts:
For: https://github.com/hypothesis/via/issues/543

This was a lot more annoying than I was expected. It looks like the conversion to make Via work as the public interface didn't follow some of the style of what came before, so there are a number of patterns and assumptions which have been broken:

 * The URL does not reliably contain `url=` anymore
 * This means we can't have a single decorator which handles checkmate checks
 * We're still passing around a checkmate object but we don't use it much any more (only in status)
 * We've introduced a `views/_helpers.py` which replicates the purpose of the `resource.py` and feels like a step backwards to old ways of doing things. Checkmate doesn't really have a helpers pattern

This PR:

 * Removes some random empty files
 * Converts the decorator to be a method call on the checkmate object

### Random musings

Should this all be on the context object?

### Testing notes

 * Start `viahtml` (just in case), and `checkmate`
 * Run `CHECKMATE_ALLOW_ALL=False make dev`
 * Visit: http://localhost:9083/https://bad.example.com
 * On `master` you should get an error page
 * On this branch you should be blocked

### DIFF
--- a/tests/unit/via/views/proxy_test.py
+++ b/tests/unit/via/views/proxy_test.py
@@ -6,52 +6,38 @@ from via.views.proxy import proxy
 
 
 class TestProxy:
-    @pytest.mark.parametrize(
-        "path,url_to_proxy",
-        [
-            ("/https://example.com/foo", "https://example.com/foo"),
-            ("/http://example.com/foo", "http://example.com/foo"),
-            (
-                "/https://example.com/foo?bar=gar&har=jar#car",
-                "https://example.com/foo?bar=gar&har=jar#car",
-            ),
-        ],
-    )
     def test_it(
-        self, pyramid_request, path, url_to_proxy, get_url_details, via_client_service
+        self,
+        pyramid_request,
+        get_url_details,
+        via_client_service,
+        url_from_user_input,
+        raise_if_blocked,
     ):
-        pyramid_request.path = path
+        pyramid_request.path = "/https://example.org"
 
         result = proxy(pyramid_request)
 
-        get_url_details.assert_called_once_with(url_to_proxy)
+        url_from_user_input.assert_called_with("https://example.org")
+        raise_if_blocked.assert_called_once_with(
+            pyramid_request, url_from_user_input.return_value
+        )
+        get_url_details.assert_called_once_with(url_from_user_input.return_value)
         via_client_service.url_for.assert_called_once_with(
-            url_to_proxy, sentinel.mime_type, pyramid_request.params
+            url_from_user_input.return_value, sentinel.mime_type, pyramid_request.params
         )
         assert result == {"src": via_client_service.url_for.return_value}
-
-    def test_it_normalizes_url(
-        self, pyramid_request, get_url_details, via_client_service, url_from_user_input
-    ):
-        pyramid_request.path = "/https://example.org"
-        url_from_user_input.side_effect = ["https://normalized.com"]
-
-        proxy(pyramid_request)
-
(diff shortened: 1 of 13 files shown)


tests/unit/via/views/proxy_test.py
```diff
@@ -6,52 +6,29 @@
...
-    ):
-        pyramid_request.path = path
-
-        result = proxy(pyramid_request)
-
-        get_url_details.assert_called_once_with(url_to_proxy)
-        via_client_service.url_for.assert_called_once_with(
-            url_to_proxy, sentinel.mime_type, pyramid_request.params
-        )
-        assert result == {"src": via_client_service.url_for.return_value}
```
jon-betts, line 31: This all appears to be retesting `url_from_user_input`

tests/unit/via/views/proxy_test.py
```diff
@@ -6,52 +6,29 @@
...
-        get_url_details.return_value = (sentinel.mime_type, sentinel.status_code)
+        get_url_details.return_value = sentinel.mime_type, sentinel.status_code
         return get_url_details
 
     @pytest.fixture(autouse=True)
     def url_from_user_input(self, patch):
-        url_from_user_input = patch("via.views.proxy.url_from_user_input")
-        url_from_user_input.side_effect = lambda url: url
-        return url_from_user_input
+        return patch("via.views.proxy.url_from_user_input")
```
jon-betts, line 34: Without this weirdness the test works much more simply with `url_from_user_input.return_value`

via/views/proxy.py
```diff
@@ -10,6 +10,8 @@ def proxy(request):
     # Strip leading '/' and normalize URL.
     url = url_from_user_input(request.path[1:])
 
+    request.checkmate.raise_if_blocked(url)
```
marcospri, line 13: Were these ones not checkmate'd at all?

via/views/view_pdf.py
```diff
@@ -16,14 +16,18 @@
...
     http_cache=0,
-    decorator=(checkmate_block, has_secure_url_token),
+    decorator=(has_secure_url_token,),
 )
 def view_pdf(context, request):
     """HTML page with client and the PDF embedded."""
 
+    url = context.url()
+
+    request.checkmate.raise_if_blocked(url)
```
marcospri, line 26: Out of scope for this PR but I wonder what's the pyramid-y way of annotating the request  on `route_by_content` (or directly on `ViaCheckmateClient`) and skip a second check if it happens

marcospri (approved):

via/views/proxy.py
jon-betts: No. I'm guessing the theory was that we were "only" peeking at the data to see what it was / that we would check it properly when it got to the next step, but the truth is we can attempt to make a quick query with HEAD and the like, but we're not in control. The server can take as long as it likes.

In the case of the Wayback machine problems we've been having, the request is the message, the caller doesn't care about the content as they are trying to trigger it to download a website. So just issuing the call puts us on the hook for 30 seconds or something.

via/views/view_pdf.py
jon-betts: Yeah. In our initial solution to this we put a signed token on that say'd we checked it, but it caused some issue I can't remember and instead of solving it, we removed it.

I think our approach right now is:

 * Checkmate is super fast
 * It's better to check twice and not have an error
 * We can optimise later (or not) if we need to

PR #557, opened 2021-08-23
Disable replacement of 4xx and 5xx errors from upstream

robertknight:
We are having a production issue with requests for some PDFs from Google Drive
failing with generic nginx 400 responses, where the 400 response is
coming from the @proxy_client_error handler and is masking the upstream
error.

Disable the error replacement of 4xx and 5xx responses via the
/proxy/static/* route to enable debugging of the issue.

A bonus of this change is that invalid secure links now result in more useful error codes, eg. 410 for an expired link.

Slack thread: https://hypothes-is.slack.com/archives/C2BLQDKHA/p1629749442452600


robertknight:
I've tested this locally and it proxies the upstream error content and HTTP status as expected. I'm going to deploy to QA to see if we can get more information.

PR #558, opened 2021-08-24
Remove all overrides of upstream errors

robertknight:
Overriding upstream 4xx and 5xx responses with generic 400 responses made it difficult to debug interactions between Via and upstreams in production because all information about the original upstream response was lost.

This commit completes the change started in https://github.com/hypothesis/via/pull/557 by removing all such overrides.

We will find another way to solve the problem described in the comments where proxied 4xx or 5xx responses from upstream could confuse monitoring tools about the health of the service.

One possibility that I haven't implemented here would be to return a fixed status code for an upstream 4xx or 5xx error but include the original response status in an HTTP header or body. I don't know if this is possible to do in nginx.

----

On a broader note, this leads me to ask whether we should consider replacing nginx as the proxying system entirely with something Python-based. That will give us much easier control over all details of the response and error logging. It would also enable us to use New Relic's Python integration for more advanced alerting (eg. observing baseline deviations in the error response rate).

### DIFF
--- a/conf/nginx/includes/errors.conf
+++ b/conf/nginx/includes/errors.conf
@@ -10,8 +10,13 @@
 recursive_error_pages on;
 
 # Intercept 3xx, 4xx and 5xx responses from the servers we're
-# proxying and process them through the error_page directives
+# proxying so we can optionally process them through error_page directives
 # below: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_intercept_errors
+#
+# If no matching `error_page` is defined then the upstream error status and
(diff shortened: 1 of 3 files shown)


seanh (approved):
Let's merge this.

I think we're gonna need to re-implement Via's non-proxying of third-party error statuses so that users proxying broken third-party sites doesn't trigger Pager Duty false alarms. But it needs to be done in a way that preserves the third-party response's status and body somewhere for debugging. I don't think we'll be reviving the implementation that this PR deletes.

PR #563, opened 2021-08-26
PoC using a service account to authenticate google drive files

marcospri:


via/views/view_pdf.py
```diff
@@ -45,9 +50,45 @@ def view_pdf(context, request):
...
+    try:
+        pdf_contents = service.files().get_media(fileId=file_id).execute()
+        return Response(body=pdf_contents)
+    except:
+        google_api_key = request.registry.settings["google_api_key"]
+        response = requests.get(
+            f"https://www.googleapis.com/drive/v3/files/$2?key={google_api_key}&alt=media"
+        )
+        response.raise_for_status()
+        return Response(body=response.contents)
```
marcospri, line 83: All this block is not actually needed but left it here so I can write this comment:

- Either a file is shared with the service account (and then it can download it)
- Or the file was made public, so the service account also has access.

via/views/view_pdf.py
```diff
@@ -45,9 +50,45 @@ def view_pdf(context, request):
...
+        r"https://drive.google.com/uc\?id=(.*)&export=download", url
+    ).group(1)
+
+    credentials = service_account.Credentials.from_service_account_file(
+        SERVICE_ACCOUNT_FILE, scopes=SCOPES
+    )
+    service = build("drive", "v3", credentials=credentials)
+    try:
+        pdf_contents = service.files().get_media(fileId=file_id).execute()
+        return Response(body=pdf_contents)
```
marcospri, line 76: This could potentially kill a gunicorn worker.

PR #564, opened 2021-08-26
Serve Google Drive from Python using a service account

jon-betts:
To get around tighter apparent quotas for API key authentication we will try using service accounts which appear to have a higher value.

This requires us to secure the end-point in some way, so we can't be used to indescriminately spam things (avoiding Checkmate etc). I've re-used the same URL signing and verification as already used by the `route_by_content()` and `view_pdf()` views for this.

This is based on: https://github.com/hypothesis/via/pull/565

#### Watch out for caching

You probably want to open the browser's dev tools and check **Disable cache** when testing this, otherwise you can be getting cached responses and not actually calling the app.

#### Testing with Via directly

 * `make devdata`
 * `make dev`
 * Visit: http://0.0.0.0:9082/pdf?url=https%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1heF_Yjs_ZZyms4QpdG0P06fTEXWkMdVW%26export%3Ddownload
 * Or directly: http://0.0.0.0:9082/google_drive/1heF_Yjs_ZZyms4QpdG0P06fTEXWkMdVW
 * It should work... 
 * There should be caching information `Cache-Control`
 * Scramble some letters of the ID
 * You _should_ get a 404 (~this is giving me a 500... why?~ Not evaluating an iterator before starting a response)

#### Testing through the LMS app

You can also test this in the LMS app. Just open [localhost (make devdata) Google Drive Assignment](https://hypothesis.instructure.com/courses/125/assignments/876).


via/services/google_drive.py
```diff
@@ -0,0 +1,60 @@
...
+        downloader = MediaIoBaseDownload(handle, request)
+        done = False
+        while done is False:
+            status, done = downloader.next_chunk()
+            # print("Download %d%%." % int(status.progress() * 100))
+
+        return handle.getvalue()
+
+    _PUBLIC_URL_REGEX = re.compile(
+        r"^https://drive.google.com/uc\?id=(.*)&export=download$", re.IGNORECASE
```
marcospri, line 53: I wonder now if we should open this one a bit.
We'll start seeing some URLs with `resourceKey` which could be the first param too. Maybe just checking that `export=download` is there, not necessarily that is the first or the only param.

via/views/view_pdf.py
```diff
@@ -25,54 +23,31 @@ def view_pdf(context, request):
...
-    # http://nginx.org/en/docs/http/ngx_http_secure_link_module.html#secure_link_md5
-    hash_ = hashlib.md5()
-    hash_.update(hash_expression.encode("utf-8"))
-    sec = hash_.digest()
-    sec = b64encode(sec)
-    sec = sec.replace(b"+", b"-")
-    sec = sec.replace(b"/", b"_")
-    sec = sec.replace(b"=", b"")
-    sec = sec.decode()
+    if google_drive_api.is_available and (
```
marcospri: Is the idea that `service_account_info` could be toggled on/off from AWS? Maybe having a envvar feature flag could be easier?

via/services/google_drive.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # https://developers.google.com/drive/api/v3/manage-downloads#download_a_file_stored_on_google_drive
+        request = self._service.files().get_media(fileId=file_id)
+
+        handle = BytesIO()
+        downloader = MediaIoBaseDownload(handle, request)
+        done = False
+        while done is False:
+            status, done = downloader.next_chunk()
+            # print("Download %d%%." % int(status.progress() * 100))
```
marcospri: Any idea if pyramid's `response.app_iter` could take a generator from here? Or more specifically would doing that stream a response?

This is probably on the things to look at later territory anyway.

via/views/view_pdf.py
jon-betts: The idea is we'd read the settings (or not) from the environment variable, and then it's either on or not. So effectively this is the solution you suggest, but just tucked away inside the service.

via/services/google_drive.py
jon-betts: I think we should, but I'd want to see some examples before I did that. Also, I don't think it's strictly related.

via/services/google_drive.py
jon-betts: I think it's a good idea. I'll give it a go.

via/get_url/details.py
```diff
@@ -56,7 +52,7 @@ def get_url_details(url, headers=None):
     if headers is None:
         headers = OrderedDict()
 
-    if GOOGLE_DRIVE_REGEX.match(url):
+    if GoogleDriveAPI.google_drive_id(url):
```
seanh: :+1: Nice

README.md
```diff
@@ -77,6 +77,8 @@ Environment variables:
 | `CLIENT_EMBED_URL` | The URL of the client's embed script | `https://hypothes.is/embed.js`
 | `ENABLE_FRONT_PAGE` | Show a front page at the root URL | `true` |
 | `GOOGLE_API_KEY` | The API key to use to authenticate with the Google Drive API |
+| `GOOGLE_DRIVE_CREDENTIALS` | Service account credentials in JSON to authenticate with the Google Drive API |
+| `GOOGLE_DRIVE_IN_PYTHON` | Process Google Drive in Python (requires `GOOGLE_DRIVE_CREDENTIALS`) |
```
seanh:
```suggestion
| `GOOGLE_DRIVE_IN_PYTHON` | Process Google Drive in Python (feature flag) (requires `GOOGLE_DRIVE_CREDENTIALS`) |
```

README.md
```diff
@@ -77,6 +77,8 @@ Environment variables:
 | `CLIENT_EMBED_URL` | The URL of the client's embed script | `https://hypothes.is/embed.js`
 | `ENABLE_FRONT_PAGE` | Show a front page at the root URL | `true` |
 | `GOOGLE_API_KEY` | The API key to use to authenticate with the Google Drive API |
+| `GOOGLE_DRIVE_CREDENTIALS` | Service account credentials in JSON to authenticate with the Google Drive API |
```
seanh, line 80: Could we add a note about where to get these credentials from? I'm guessing the JSON file is downloaded from some Google console somewhere?

tests/unit/via/services/google_drive_test.py
```diff
@@ -0,0 +1,183 @@
...
+from h_matchers import Any
+from pyramid.httpexceptions import HTTPNotFound
+
+from via.exceptions import ConfigurationError, UpstreamServiceError
+from via.services import GoogleDriveAPI
+from via.services.google_drive import factory
+
+
+def build_http_exception(status_code, message=None):
+    """Create one of Googles annoying error classes."""
```
seanh:
```suggestion
    """Return one of Google's annoying exception objects."""
```

seanh: We'd normally put helpers like this at the bottom of the file (like we do with fixtures)

via/services/google_drive.py
```diff
@@ -0,0 +1,123 @@
+import json
+import re
+from io import BytesIO
+from json import JSONDecodeError
+from typing import ByteString, Iterator
+
+from google.oauth2.service_account import Credentials
+from googleapiclient.discovery import build
+from googleapiclient.errors import HttpError
+from googleapiclient.http import MediaIoBaseDownload
```
seanh: I wonder if we could do something like:

```suggestion
import google
import googleapiclient
```

then then use them in the code as `google.oauth2.service_account.Credentials`, `googleapiclient.discovery.build`.

It's more long-winded when you have to use the names in the code. But each name is only used once, and it makes it clear in the code when we're referring to Google library things. When seeing very short and generic names like `Credentials`, `build` and `HttpError` it's hard to remember exactly what they are and that they come from the Google libraries. You have to look back up at the imports to see. You have to keep in your head a list of all the names that come from the Google libs.

It'd also make the tests nicer because you'd be patching `"via.services.google_drive.google"` and `"via.services.google_drive.googleapiclient"` rather than having fixtures named `Credentials` and `build` (same problem: you can't really tell from the name what the fixture is).

In general I think the pattern of importing multiple short and generic names from a third-party package is one to avoid where possible. It can make the code more concise but makes it harder to read by adding a bunch of unclear names to the module's namespace, where the reader has to work to remember which names belong to what and which are defined locally.

**If doing this I think you might need to keep the `from googleapiclient.errors import HttpError` import separate** since your tests are using the real exception class (and I don't think you want to or can mock the exception class: I don't think mocking exceptions works).

via/services/google_drive.py
```diff
@@ -0,0 +1,123 @@
...
+
+from via.exceptions import ConfigurationError, UpstreamServiceError
+
+
+class GoogleDriveAPI:
+    """Simplified interface for interacting with Google Drive."""
+
+    SCOPES = [
+        # If we want metadata
+        "https://www.googleapis.com/auth/drive.metadata.readonly",
```
seanh, line 19: Are we using this one? If not it might be misleading to leave it here

via/app.py
```diff
@@ -22,6 +22,8 @@
     # Optional
     "checkmate_ignore_reasons": {},
     "checkmate_allow_all": {"formatter": asbool},
+    "google_drive_credentials": {},
```
seanh, line 25: Does the name "credentials" come from the Drive API or its docs? Based on the Drive API's `Credentials.from_service_account_info(...)` I'd think this'd be called a `google_drive_service_account_info`. Otherwise the name changes from "credentials" to "service account info" mid-way through the code flow, which makes it harder to follow

tests/unit/via/services/google_drive_test.py
```diff
@@ -0,0 +1,183 @@
...
+    def build(self, patch):
+        return patch("via.services.google_drive.build")
+
+    @pytest.fixture(autouse=True)
+    def Credentials(self, patch):
+        return patch("via.services.google_drive.Credentials")
+
+    @pytest.fixture(autouse=True)
+    def MediaIoBaseDownload(self, patch):
+        return patch("via.services.google_drive.MediaIoBaseDownload")
```
seanh: :+1: These were pretty easy to follow. We're breaking our rule by mocking the Google libs: we normally unit test without mocking third-party libs mostly so that we have more confidence that our code still works when we update the third-party lib. But in the case of these Google libs I guess they're going to send network requests to Google so I'm not sure we have much choice really.

We _could_ try using `httpretty` to mock out the network layer. I think that would likely prevent Google's code from sending real network requests. But then we'd have to set up fakes of the server responses _that Google's code expects internally_ which is terrible.

So I think this is probably the best call and autospeccing should help protect us somewhat against Google changing their Python API.

via/views/view_pdf.py
jon-betts: I've added a separate var, but it's still behind the `is_available` as it keeps all the logic in one place.

via/views/google_drive.py
```diff
@@ -0,0 +1,30 @@
...
+from via.services.google_drive import GoogleDriveAPI
+from via.services.secure_link import has_secure_url_token
+
+
+@view_config(
+    route_name="google_drive_file",
+    decorator=(has_secure_url_token,),
+)
+def get_file_content(_context, request):
+    """Download a file from Google Drive."""
```
seanh: You can fit the view config on one line (Black is formatting it onto two lines because of the trailing comma after the last item).

If you're not using `context` in a Pyramid view you can just omit it, you don't need to take a `_context` arg.

Also suggesting a tweak to the route and view name and docstring: I think "proxying" is the verb we tend use for this. I get that this isn't quite the same as what our currently-deployed NGINX-based solution for Google Drive does where it's literally proxying an HTTP response from Google. Still, I think it's close enough, it fills the same functional need as when we "proxy" a PDF from Canvas or Blackboard files or when Via HTML "proxies" a web page etc.

```suggestion
@view_config(route_name="proxy_google_drive_file", decorator=(has_secure_url_token,))
def proxy_google_drive_file(request):
    """Proxy a file from Google Drive."""
```

via/views/google_drive.py
```diff
@@ -0,0 +1,30 @@
...
+            # We'll just assume everything from Google Drive is a PDF
+            "Content-Type": "application/pdf",
+            # Add a very generous caching policy of half a day max-age, full
+            # day stale while revalidate.
+            "Cache-Control": "public, max-age=43200, stale-while-revalidate=86400",
+        }
+    )
+    # Add an iterable to stream the content instead of holding it all in memory
+    response.app_iter = request.find_service(GoogleDriveAPI).iter_file(
+        request.matchdict["file_id"]
```
seanh: I like the use of `matchdict` so that it's not possible for this view to even be called without a `file_id` (as opposed to if we used a query param)

via/views/google_drive.py
```diff
@@ -0,0 +1,30 @@
...
+)
+def get_file_content(_context, request):
+    """Download a file from Google Drive."""
+
+    response = request.response
+    response.headers.update(
+        {
+            "Content-Disposition": "inline",
+            # We'll just assume everything from Google Drive is a PDF
+            "Content-Type": "application/pdf",
```
seanh, line 18: Yep :+1: This is totally a safe assumption and I think will remain so for a long time

via/views/google_drive.py
```diff
@@ -0,0 +1,30 @@
...
+            # Add a very generous caching policy of half a day max-age, full
+            # day stale while revalidate.
+            "Cache-Control": "public, max-age=43200, stale-while-revalidate=86400",
```
seanh, line 21: :+1: I guess the nice thing about "proxying" this way is that we're not literally proxying a file download HTTP response from Google (like the currently deployed NGINX solution does) so there isn't the possibility of a bunch of caching-related headers from Google that we have to delete

tests/unit/via/views/google_drive_test.py
```diff
@@ -0,0 +1,22 @@
...
+
+        assert response.headers["Content-Disposition"] == "inline"
+        assert response.headers["Content-Type"] == "application/pdf"
+        assert (
+            response.headers["Cache-Control"]
+            == "public, max-age=43200, stale-while-revalidate=86400"
+        )
+
+        google_drive_api.iter_file.assert_called_once_with(sentinel.file_id)
+        assert response.app_iter == google_drive_api.iter_file.return_value
```
seanh: :+1: Nice and easy to test

via/services/google_drive.py
```diff
@@ -0,0 +1,123 @@
...
+        # To actually get the file
+        "https://www.googleapis.com/auth/drive.readonly",
+    ]
+
+    def __init__(self, service_account_info):
```
seanh, line 24: This is where the name "Google drive credentials" changes to "service account info" for the same object. I think it'd be easier to follow if we stuck to one name

via/services/google_drive.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @property
+    def is_available(self):
+        """Get whether the Google Drive API is available."""
+
+        return self._service is not None
+
+    _PUBLIC_URL_REGEX = re.compile(
+        r"^https://drive.google.com/uc\?id=(.*)&export=download$", re.IGNORECASE
+    )
```
seanh, line 54: I think the use of a regex here comes from [existing code](https://github.com/hypothesis/via/blob/f6383f688034d3af536407d5bba45983f4604a26/via/get_url/details.py#L19-L21) but just FYI [we need to replace this with proper URL parsing](https://github.com/hypothesis/via/issues/569). Can be a separate issue

via/services/secure_link.py
```diff
@@ -31,6 +42,14 @@ def request_is_valid(self, request):
 
         return True
 
+    def sign_url(self, url):
+        """Get a signed URL (if URL signing is enabled)."""
+
+        if self._signed_urls_required:
+            return self._via_secure_url.create(url)
+
+        return url
```
seanh, line 51: :+1: We should probably remove `signed_urls_required`. I'm guessing it's always on

seanh (changes requested):
This looks good and works for me locally. I added notes to the PR description about how to test it in Canvas.

I haven't tested the issue mentioned in the PR description about why it's 500'ing when it should be 404'ing. I'm gonna make some tea and then I'll take a look at that.

I'm suggesting that we try to be consistent about calling them "Google Drive credentials" or "service account info".

Suggested a tweak to the Google lib imports that I think would make the code easier to read even though it'll make some lines longer.

Suggested renaming the route, view, etc to "proxy Google Drive file".

Various other small tweaks suggested.

Do we have any way of testing that slow or large responses actually are getting streamed, and aren't piling up in memory somewhere in the pipeline? And I guess we'd have to actually test that in production as that might differ from dev.

Maybe if we upload a really big PDF to Drive?

via/services/google_drive.py
jon-betts: I'll give this a shot in this case, as I think it might help. but I'm wary of making general advice from it.

It could a bit of a trap to look at a specific situation and extrapolate policy as the decision here hangs on a number of local decisions and judgement calls:

* It increases one type of readability at the expense of another. What's the exact cut off?
* This causes a problem with patching, but in this case we consider that an acceptable trade off. When is it too much?
* We only use them once. What if we used them twice, when's the cut off?
* The names are "generic". At which point do names become too generic? What makes a name good enough?
* This is 3rd party, but presumably `BytesIO` isn't?

The point isn't that the questions above don't have answers, they do, but the answers are situational. Which is to say, we may be dealing with a number of distinct situations which always require judgement rather than an opportunity for a general rule.

Having been refactoring about the code base a lot recently, I've seen this much more over used than under. If we had to pick a side of the coin for a simple rule (without exercising judgement) I'd recommend never doing it rather than always doing it. The bad effects I've seen are usually about the bloat you mention, which often pushes a single line onto two, which doesn't sound like a lot but can make things a lot longer and harder to read sometimes. 

The patching is also a bit weird. It seems nice to patch something out all in one go, but it can lead to some unwieldy assertions in the end. Pytests chains of calls and return values can already make somethings quite long:

```python
        MediaIoBaseDownload.return_value.next_chunk.assert_has_calls(...
```

If it's fully qualified that becomes:

```python
        googleapiclient.http.MediaIoBaseDownload.return_value.next_chunk.assert_has_calls(...
```

I think you can still sort of deep patch into the name space in the file to select different bits to get around that, but that would get us back to where we started with the tests.

tests/unit/via/services/google_drive_test.py
jon-betts: I've moved this to the bottom and changed the name to "build_http_error" to mirror the class name.

via/services/google_drive.py
jon-betts: I'm working on a separate PR for this. I could have just smooshed them together honestly... but never mind. It's not a change actually related to this PR, so it's probably good form not to.

via/views/google_drive.py
jon-betts: Yeah I suppose it's a double edged sword. We're able to set all the values, but we also _need_ to set them all correctly.

I honestly don't know if the above is a good system, but I think it _should_ result in Cloudlflare keeping it and checking in with us a max of twice per day.

README.md
jon-betts: I was being a bit cagey about it here. Maybe a Stack Overflow question would be good. The actual answer to this is not something you can explain quickly unfortunately.

tests/unit/via/services/google_drive_test.py
jon-betts: Yeah, I couldn't see any sane way of doing this in this situation. Also to be honest I'm not sure that is the rule really. I think it's more like "who's code are you testing?". 

In the event that the situation is easy to setup, and obvious to verify, then it's probably our code. So even though Postgres is mega complicated, it's easy to configure it and easy to tell if it's doing what we want.

In most cases where we talk to a cloud service it's likely to be the opposite. We can't easily control (or even necessarily understand) what they are up to, so maybe the best we can say is we follow the interface. Google services in particular strike me as something who's behavior would shift without notice, so even if we could find a set of HTTP responses, I'm not sure we could rely on them.

via/app.py
jon-betts: They both come from Google. Service accounts are one type of credentials I think. So potentially we could one day have a different form of credentials and consume them differently.

via/views/google_drive.py
```diff
@@ -0,0 +1,39 @@
...
+        }
+    )
+    # Add an iterable to stream the content instead of holding it all in memory
+    response.app_iter = _reify_first(
+        request.find_service(GoogleDriveAPI).iter_file(request.matchdict["file_id"])
+    )
```
seanh, line 27: The [docs for `app_iter`](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/webob.html?highlight=app_iter#response) say that it's <q>also accessible as `response.body (bytes)`, `response.text` (a Unicode string, informed by `response.charset`), and `response.body_file` (a file-like object; ...</q>. If this is really true and `response.body` is the same thing as `response.app_iter` then it might be nicer to use `response.body` here, it's maybe a slightly more friendly name.

seanh:
I tested the error handling. Some issues:

* We do need to report unexpected errors from Google (non-404s) to Sentry from either the first or subsequent calls to `next_chunk()` (or any of the other calls to Google's libs for that matter). These aren't being reported currently
* We do need an alarm to go off if we start getting a large number of Google Drive proxying failures. Currently I think the use of 409's for these will evade our alarms.

Suggestions and testing notes:

### How to test Sentry reporting

You need to go and make yourself a free Sentry account, create a `via-dev` project in it, get the project's "DSN" from Sentry and set the envvar:

```shell-session
$ export SENTRY_DSN="https://d28...6ed@o3..66.ingest.sentry.io/59...59"
```

Then Sentry reporting to your personal account will just work.

### File that doesn't exist (404 from Google)

If I load <http://0.0.0.0:9082/google_drive/ID_THAT_DOES_NOT_EXIST> then:

:heavy_check_mark: I get a 404 from Via

:heavy_check_mark: I see a nice error page:

:heavy_check_mark: I don't see anything in Sentry (this is correct: we don't want crash reports for 404s).

Screenshot from 2021-08-31 17-51-40

### Unexpected exception (not a 404) from the first `next_chunk()` call to Google's API

To test this I hacked the code so that we no longer recognise 404's:

```diff
diff --git a/via/services/google_drive.py b/via/services/google_drive.py
index 4e7ba77..09fc659 100644
--- a/via/services/google_drive.py
+++ b/via/services/google_drive.py
@@ -92,7 +92,7 @@ class GoogleDriveAPI:
 
     @staticmethod
     def _raise_translated_error(http_error):
-        if http_error.status_code == 404:
+        if http_error.status_code == 987:
             try:
                 # Is this always the case? It seems like Google errors don't
                 # have a fixed format by looking at the code to generate them.
```

Then visited <http://0.0.0.0:9082/google_drive/ID_THAT_DOES_NOT_EXIST>

:x:  I got a 409 Conflict from Via which means that alarms in AWS and New Relic will not go off. I think we **do** want alarms to go off if lots of attempts to proxy Google Drive files are failing. Possible solutions:

1. Use a 5xx status code (and verify that Cloudflare doesn't replace our error pages with its own)
2. Continue to use a 4xx status code but also report a New Relic custom metric. We could have a "Google Drive Proxy" dashboard in New Relic and one of the graphs could be the failure rate, and we could also set alarms based on that failure rate.
3. Or instead of a custom metric can you just make a graph of 409's from this endpoint in New Relic and alarm on that?

:heavy_check_mark: I got a nice error page:

:x: Nothing got reported to Sentry. I think we do want these to go to Sentry. This is because the code is raising `UpstreamServiceError` which is a subclass of Pyramid's `HTTPConflict` and [Sentry doesn't report Pyramid `HTTPException` subclasses by default](https://github.com/getsentry/sentry-python/blob/832263bedca595be1e31a519d4f49f477bd77760/sentry_sdk/integrations/pyramid.py#L140-L141). I think we probably do want these to be reported to Sentry. If we still want the status code to be 409 and we do want to report them to Sentry then I think we probably have to call [`h-pyramid-sentry`'s `report_exception()`](https://github.com/hypothesis/h-pyramid-sentry/blob/fa3a0b560d0a077ec375dea94cc8900f4d47a96b/src/h_pyramid_sentry/__init__.py#L14-L23) manually.

Screenshot from 2021-08-31 17-57-28

### Exception from a subsequent `next_chunk()` call

To test this I hacked the code to make multiple `next_chunk()` calls (by making the `chunksize` really small) and also hacked the _third_ `next_chunk()` call to raise an unknown error:

```diff
diff --git a/via/services/google_drive.py b/via/services/google_drive.py
index 4e7ba77..761861d 100644
--- a/via/services/google_drive.py
+++ b/via/services/google_drive.py
@@ -79,11 +79,18 @@ class GoogleDriveAPI:
         request = self._service.files().get_media(fileId=file_id)
 
         handle = BytesIO()
-        downloader = MediaIoBaseDownload(handle, request)
+        downloader = MediaIoBaseDownload(handle, request, chunksize=1)
         done = False
+        count = 0
         while not done:
+            count += 1
             try:
                 _status, done = downloader.next_chunk()
+                if count == 3:
+                    import httplib2
+                    raise HttpError(
+                        httplib2.Response({"status": 437}), content=b""
+                    )
             except HttpError as exc:
                 self._raise_translated_error(exc)
```

:heavy_check_mark: I got a 200 OK response status. This is to be expected: we've already sent the response headers to the client by the time the exception happens so we can't prevent it from being a 200.

:heavy_check_mark: I see a "Failed to load PDF document" error from PDF.js. This is completely unhelpful to users or Hypothesis staff trying to debug. But that's to be expected: when an error occurs mid-chunking it's already too late for us to get a helpful error response to the client.

:x: Nothing got reported to Sentry. I think we **do** want to report these to Sentry. The reason nothing got reported is again that we raised `UpstreamServiceError` which is an `HTTPError` subclass which Sentry ignores. I think there are two possible solutions:

1. Call `report_exception()` manually, same as when the first `download_chunk()` call raises
2. Raise an exception that is not an `HTTPError` subclass. Note that in this case (when the exception happens in a subsequent `download_chunk()` call) the 200 OK has already been sent to the client so you could even raise `RuntimeError` and that's not going to cause the HTTP status to be 500.

Screenshot from 2021-08-31 18-37-02

seanh:
I think we may just want to return a 500 Internal Server Error if we get any kind of unexpected exception from any of the calls to Google. So probably just let the exception from Google fly, don't even catch/wrap it. And then let's try to test that and see whether Cloudflare interferes with it or not, and if so whether we can disable the interference.

Certain 5xx responses from us get replaced by Cloudflare but I think that might only be things like 502 and 504. We should verify.

I think we probably do want AWS and the like (New Relic) to consider the LMS app to be broken if the Google Drive proxying endpoint is failing a lot.

seanh (changes requested):

via/views/google_drive.py
jon-betts: The wording is a bit unclear here, but what this is saying is if you set the data via `app_iter` it's accessible through these options, not that they are aliases for it.

You can't set `request.body` to an iterator.

seanh (approved):

PR #565, opened 2021-08-27
Change the secure link decorator to be a service instead

jon-betts:
This is some prep work to make it easier for us to sign our own URLs.


via/services/secure_link.py
```diff
@@ -0,0 +1,39 @@
...
+def has_secure_url_token(view):
+    """Require the request to have a valid signature."""
+
+    def view_wrapper(context, request):
+        if not request.find_service(SecureLinkService).request_is_valid(request):
+            return HTTPUnauthorized()
+
+        return view(context, request)
+
+    return view_wrapper
```
seanh, line 15: :+1:

As a view decorator the previous `views/decorators.py` location seems like the "right" place for this: `services/` is supposed to be for "business logic"-type concerns and decoupled from Pyramid (as much as possible) whereas `views/` is supposed to be for views-like concerns and coupled to Pyramid, and a view decorator is very much a Pyramid/views concern: it exists only to serve Pyramid, to be used by Pyramid `@view_config`'s, the entire definition of what a view decorator _is_, how it gets called, with what arguments, what it returns, is determined by Pyramid.

That said, I don't think the "violation" here (of the definition of the services directory) is a big deal and it keeps `SecureLinkService`-related code together.

seanh: The code in this decorator seems to be simpler than [how it used to be](https://github.com/hypothesis/via/blob/b4839541d21ba63a05fccf8f6334ec7dd58da5ce/via/views/decorators.py#L7-L23). It's no longer checking the `signed_urls_required` setting and I can see that got moved into `SecureLinkService.request_is_valid()` below. It's also no longer pulling the signature out from `request.params.get(signature_param)` but I don't see where that went?

via/services/secure_link.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class SecureLinkService:
+    def __init__(self, secret, signed_urls_required):
+        self._secure_token = ViaSecureURL(secret)
```
seanh:
```suggestion
        self._via_secure_url = ViaSecureURL(secret)
```

The name change from "Via Secure URL" to "secure token" confused me when reading the `request_is_valid()` method below. I had to look back up here at the `__init__()` to see what `self._secure_token` was and to see that it's an instance of `ViaSecureURL`.

The previous version of the code was more readable by doing `ViaSecureURL(request.registry.settings["via_secret"])).verify(request.url)` every time. No looking around to see what instance variables are or where the secret comes from. But I can see that this version avoids creating a new `ViaSecureURL` every time

seanh (changes requested):
I started reviewing this but have to go for lunch. Posting my comments so far and I'll continue reviewing after lunch

via/services/secure_link.py
jon-betts: The param is extracted and checked inside the `ViaSecureURL`.

There is a change of behavior here. Previously this would allow you skip signature checking if there was no signature, _and_ no signature was required. Now we just skip checking if no signature is required. This is why it's gotten simpler and now clearer that us getting the param out wasn't really for much. Having a variable for the param name was misleading too, as it implied you could change the value, but the code didn't pass this through to the `ViaSecureURL` (never mind the other service calling us) as we should have and so you couldn't really.

I think this previous behavior was to allow us to deploy it and test it without it being enabled. So we _needed_ it to fail for testing purposes, not really security purposes.

via/services/secure_link.py
jon-betts: I've renamed the variable

via/views/route_by_content.py
```diff
@@ -5,7 +5,7 @@
 
 from via.get_url import get_url_details
 from via.services import ViaClientService
-from via.views.decorators import has_secure_url_token
+from via.services.secure_link import has_secure_url_token
```
seanh: We should hoist `has_secure_url_token` into `services/__init__.py` so that it's easier to import: `from via.services import has_secure_url_token`. Especially if importing other things from services (exceptions) at the same time

seanh (approved):
:+1: Looks good to me. I think some name-hoisting into `__init__.py` could be done, that's my only suggestion

PR #574, opened 2021-08-31
Google API service account imports - POC

jon-betts:
This is to trial a suggestion from review about using absolute style imports. I've done this as a separate PR as I'm not sure it's better really, and thought it might be worth discussing.

This is a modification of: https://github.com/hypothesis/via/pull/564


via/services/google_drive.py
```diff
@@ -4,10 +4,10 @@
 from json import JSONDecodeError
 from typing import ByteString, Iterator
 
-from google.oauth2.service_account import Credentials
-from googleapiclient.discovery import build
+import google.oauth2.service_account
+import googleapiclient.discovery
+import googleapiclient.http
```
jon-betts, line 9: If you don't fully qualify the imports then patching fails with "no such attribute blah on google" etc.

jon-betts:
My feeling on this is it's not really better. I think it partially achieves the aim of making it so that a reader who didn't look at any context could see which parts came from Google, but not completely as we still import the exception separately. So it's a partial win for one kind of readability in the service code.

But I'd say it's worse on the test side for a couple of reasons:

 * It makes fixture usage very long in combination with `return_value` etc.
 * It make the fixtures imported into methods unhelpful for knowing what the test is about

The first I'd say is quite obvious, but I'll go into a little more detail about the second in case it's not. If you have a test with a signature like this:

```python
def test_it_does_the_do(self, user, annotation)
```

You get a bit of information about the focus of the tests (it involves a user and annotation somehow?). But that context is lost when you have a signature like this instead:

```python
def test_it_does_the_do(self, models)
```

Now we can only tell that it does something with one or more models. I think we have a similar thing here where some signatures have changed like this:

Old:

```python
def test_it_builds_a_service_as_we_expect(self, Credentials, build):
def test_it_with_bad_credentials(self, Credentials):
def test_iter_file(self, MediaIoBaseDownload):
def test_iter_file_with_errors(self, MediaIoBaseDownload, code, expected_exception):
```

New:
```python
def test_it_builds_a_service_as_we_expect(self, google, googleapiclient):
def test_it_with_bad_credentials(self, google):
def test_iter_file(self, googleapiclient):
def test_iter_file_with_errors(self, googleapiclient, code, expected_exception):
```

So we've traded knowing what the thing was but not knowing where it came from in one situation, to only knowing where the things come from, but not what or how many there are in another.

Overall, I'd say it's probably reduced clarity, and specifically in our tests which I think could be worse as it's the majority of our code and maintenance.

What do you think?

jon-betts:
This is been superseded

PR #576, opened 2021-09-01
Separate components of `get_url` out so they can be re-used

jon-betts:
This makes some components like the error handling and header adding separate from the `get_url` method so they can be re-used. Specifically so we can re-use them in the Google Drive work.

This moves `via.get_url.details` to `via.get_url`. At this point it should almost certainly be a service as this would allow us to have a single HTTP session which would speed up requests to the same domain (as we'd skip SSL verification).

The rest of the error handling code and header things have moved to `via.requests_tools` which are generic(ish) helpers for working with `requests`.

### DIFF
--- a/tests/unit/via/get_url_test.py
+++ b/tests/unit/via/get_url_test.py
@@ -4,15 +4,10 @@ from unittest.mock import sentinel
 import pytest
 from h_matchers import Any
 from requests import Response
-from requests.exceptions import (
-    MissingSchema,
-    ProxyError,
-    SSLError,
-    UnrewindableBodyError,
-)
+from requests.exceptions import SSLError
 
-from via.exceptions import BadURL, UnhandledException, UpstreamServiceError
-from via.get_url.details import get_url_details
+from via.exceptions import BadURL, UpstreamServiceError
+from via.get_url import get_url_details
 
 
 class TestGetURLDetails:
@@ -51,20 +46,14 @@ class TestGetURLDetails:
         )
 
     @pytest.mark.usefixtures("response")
-    def test_it_cleans_and_passes_on_the_users_headers(self, requests, clean_headers):
-        get_url_details(url="http://example.com")
+    def test_it_modifies_headers(self, requests, clean_headers, add_request_headers):
+        get_url_details(url="http://example.com", headers={"X-Pre-Existing": 1})
 
         _args, kwargs = requests.get.call_args
 
-        assert kwargs["headers"] == clean_headers.return_value
-
-    @pytest.mark.usefixtures("response")
-    def test_it_adds_abuse_policy_headers(self, requests):
-        get_url_details(url="http://example.com", headers={})
-
-        headers = requests.get.call_args[1]["headers"]
-        assert headers["X-Abuse-Policy"] == "https://web.hypothes.is/abuse-policy/"
-        assert headers["X-Complaints-To"] == "https://web.hypothes.is/report-abuse/"
+        clean_headers.assert_called_once_with({"X-Pre-Existing": 1})
+        add_request_headers.assert_called_once_with(clean_headers.return_value)
+        assert kwargs["headers"] == add_request_headers.return_value
 
     def test_it_assumes_pdf_with_a_google_drive_url(self, requests):
         result = get_url_details(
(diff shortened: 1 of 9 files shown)


tests/unit/via/get_url/error_handling_test.py
```diff
@@ -0,0 +1,39 @@
...
+        (
+            (exceptions.MissingSchema, BadURL),
+            (exceptions.InvalidSchema, BadURL),
+            (exceptions.InvalidURL, BadURL),
+            (exceptions.URLRequired, BadURL),
+            (exceptions.ConnectionError, UpstreamServiceError),
+            (exceptions.Timeout, UpstreamServiceError),
+            (exceptions.TooManyRedirects, UpstreamServiceError),
+            (exceptions.SSLError, UpstreamServiceError),
+            (exceptions.UnrewindableBodyError, UnhandledException),
```
jon-betts, line 20: I decided to just include all of the different exceptions... why not

via/exceptions.py
```diff
@@ -1,24 +1,10 @@
...
-    exceptions.InvalidURL,
-    exceptions.URLRequired,
-)
-REQUESTS_UPSTREAM_SERVICE = (
-    exceptions.ConnectionError,
-    exceptions.Timeout,
-    exceptions.TooManyRedirects,
-    exceptions.SSLError,
-)
-
```
jon-betts, line 21: This keeps these exceptions lists private to where they are used. Not quite sure why these were here if the function wasn't...

via/exceptions.py
```diff
@@ -1,24 +1,10 @@
...
-    exceptions.InvalidSchema,
-    exceptions.InvalidURL,
-    exceptions.URLRequired,
-)
-REQUESTS_UPSTREAM_SERVICE = (
-    exceptions.ConnectionError,
-    exceptions.Timeout,
-    exceptions.TooManyRedirects,
-    exceptions.SSLError,
-)
```
seanh, line 20: It doesn't seem like this was the right place for these, to me. `exceptions.py` is where we keep our custom exception classes and exception views. These are just some constants for grouping together different types of `requests` exceptions. I think the constants probably should have always been defined where they were used (and they don't seem like they should be used in more than one place if we have the code design right)

via/request_tools/__init__.py
```diff
@@ -0,0 +1,2 @@
+from via.request_tools.error_handling import handle_errors
+from via.request_tools.headers import add_request_headers, clean_headers
```
seanh: Should this be `requests_tools` rather than `request_tools`? I think they're specifically tools for working with the `requests` library

via/request_tools/error_handling.py
```diff
@@ -0,0 +1,40 @@
...
+REQUESTS_UPSTREAM_SERVICE = (
+    exceptions.ConnectionError,
+    exceptions.Timeout,
+    exceptions.TooManyRedirects,
+    exceptions.SSLError,
+)
+
+
+def handle_errors(inner):
+    """Translate errors into our application errors."""
```
seanh:
```suggestion
    """Translate `requests` errors into custom ones."""
```

via/request_tools/error_handling.py
```diff
@@ -0,0 +1,40 @@
...
+            return inner(*args, **kwargs)
+
+        except REQUESTS_BAD_URL as err:
+            raise BadURL(err.args[0]) from None
+
+        except REQUESTS_UPSTREAM_SERVICE as err:
+            raise UpstreamServiceError(err.args[0]) from None
+
+        except RequestException as err:
+            raise UnhandledException(err.args[0]) from None
```
seanh: Is the `from None` existing behaviour here? That looks like it deliberately swallows the original error but I'm not sure why we'd want to do that? I'd expect `raise CustomErrorClass(...) from err`

seanh: Also not sure about the `err.args[0]`'s. Are those really necessary / why do we do that? It looks like it might crash if `requests` ever raises an exception where `err.args` is empty

seanh: Yeah, looks like both of these came from the existing code. We might want to consider changing it to `raise CustomErrorClass() from err` (with no arguments to `CustomErrorClass`)

tests/unit/via/get_url_test.py
```diff
@@ -80,21 +69,12 @@ def test_it_raises_BadURL_for_invalid_urls(self, bad_url):
...
-            (UnrewindableBodyError, UnhandledException),
-        ),
-    )
-    def test_it_catches_requests_exceptions(
-        self, requests, request_exception, expected_exception
-    ):
-        requests.get.side_effect = request_exception("Oh noe")
+    def test_it_catches_requests_exceptions(self, requests):
+        # We don't need to go crazy here, just enough to prove we have the
+        # decorator applied
```
seanh:
```suggestion
```

I think it's unclear what _go crazy_ or _the decorator_ refer to. Especially if the code changes in future so that the function has more than one decorator or so that `@handle_errors` is no longer a decorator, this comment is unlikely to get updated if that happens

seanh (approved):

via/exceptions.py
jon-betts: That's the situation we now have, which is better.

via/request_tools/__init__.py
jon-betts: Yeah, I'll change that.

tests/unit/via/get_url_test.py
jon-betts: I'll reword this

via/request_tools/error_handling.py
jon-betts: Requests exceptions happen to always have a message, so it's actually fine. It would be potentially dodgy in general.

I'll change it over. It's purely for us when debugging. We don't rely on the exact message at all when presenting the errors in the interface (just the class).

via/request_tools/__init__.py
jon-betts: In fact I think I meant it to already be this as the test directory shows... :facepalm:

PR #579, opened 2021-09-03
Update the Google Drive timeout to 10 seconds

jon-betts:
This matches the timeout we have for regular URLs as 1 second does not appear to be enough.

### DIFF
--- a/tests/unit/via/services/google_drive_test.py
+++ b/tests/unit/via/services/google_drive_test.py
@@ -47,7 +47,7 @@ class TestGoogleDriveAPI:
                 "X-Complaints-To": Any.string(),
             },
             stream=True,
-            timeout=1,
+            timeout=10,
         )
 
         api._session.get.return_value.raise_for_status.assert_called_once_with()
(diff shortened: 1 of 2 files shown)


tests/unit/via/services/google_drive_test.py
```diff
@@ -47,7 +47,7 @@ def test_iter_file(self, api, stream_bytes):
                 "X-Complaints-To": Any.string(),
             },
             stream=True,
-            timeout=1,
+            timeout=10,
```
indigobravo, line 50: Timeout increased from 1 to 10.

via/services/google_drive.py
```diff
@@ -84,7 +84,7 @@ def iter_file(self, file_id) -> Iterator[ByteString]:
             }
         )
 
-        response = self._session.get(url=url, headers=headers, stream=True, timeout=1)
+        response = self._session.get(url=url, headers=headers, stream=True, timeout=10)
```
indigobravo, line 87: Timeout increased from 1 to 10.

indigobravo (approved):
Looks good. Approved.

PR #580, opened 2021-09-03
Google Drive credentials in a file

jon-betts:
We're moving to a single mechanism for adding new things to the container which is to copy files from S3. This loads from a specific directory instead of an environment variable.

This also takes the opportunity to swap to a list. This should hopefully prevent any kind of cliff edge deployment if we need to add more credentials in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,10 @@ Environment variables:
 | `CHECKMATE_ALLOW_ALL` | Whether to bypass Checkmate's allow-list (and use only the blocklist) | `true`
 | `CHECKMATE_IGNORE_REASONS` | Comma-separated list of Checkmate block reasons to ignore | `publisher-blocked,high-io` |
 | `CLIENT_EMBED_URL` | The URL of the client's embed script | `https://hypothes.is/embed.js`
+| `DATA_DIRECTORY` | Directory for externally provided data | `/via-data`
 | `ENABLE_FRONT_PAGE` | Show a front page at the root URL | `true` |
 | `GOOGLE_API_KEY` | The API key to use to authenticate with the Google Drive API |
-| `GOOGLE_DRIVE_CREDENTIALS` | Service account credentials in JSON to authenticate with the Google Drive API |
-| `GOOGLE_DRIVE_IN_PYTHON` | Process Google Drive in Python (feature flag) (requires `GOOGLE_DRIVE_CREDENTIALS`) |
+| `GOOGLE_DRIVE_IN_PYTHON` | Process Google Drive in Python (feature flag) (requires `google_drive_credentials.json` specified below.) |
 | `NEW_RELIC_*` | Various New Relic settings. See New Relic's docs for details |
 | `NGINX_SECURE_LINK_SECRET` | The NGINX secure links signing secret. This is used by Via's Python endpoints to generate the signed URLs required by its NGINX-implemented `/proxy/static/` endpoint. All instances of Via must have this setting |
 | `NGINX_SERVER` | The URL of Via's NGINX server for proxying PDF files | `https://via.hypothes.is`
@@ -86,6 +86,12 @@ Environment variables:
 | `SIGNED_URLS_REQUIRED` | Require URLs to Via's Python endpoints to be signed so that Via can only be used by something that has the URL signing secret. Public instances of Via should _not_ enable this. Private instances of Via (e.g. the LMS app's instance of Via) _should_ enable this | `true`
 | `VIA_HTML_URL` | The URL of the Via HTML instance to redirect to for proxying HTML pages | `https://viahtml.hypothes.is/proxy`
 | `VIA_SECRET` | The secret that must be used to sign URLs to Via's Python endpoints if `SIGNED_URLS_REQUIRED` is on |
+
+Expected data:
+
+The following data is expected to be provided in the `DATA_DIRECTORY`:
+
+ * `google_drive_credentials.json` - A list of credential JSON objects provided by the Google API console if using `GOOGLE_DRIVE_IN_PYTHON`
 
 Updating the PDF viewer
 -----------------------
--- a/bin/devdata.py
+++ b/bin/devdata.py
@@ -15,10 +15,17 @@ def _get_devdata():
         check_call(["git", "clone", "git@github.com:hypothesis/devdata.git", git_dir])
 
         # Copy devdata env file into place.
-        copyfile(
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -75,10 +75,10 @@ Environment variables:
 | `CHECKMATE_ALLOW_ALL` | Whether to bypass Checkmate's allow-list (and use only the blocklist) | `true`
 | `CHECKMATE_IGNORE_REASONS` | Comma-separated list of Checkmate block reasons to ignore | `publisher-blocked,high-io` |
 | `CLIENT_EMBED_URL` | The URL of the client's embed script | `https://hypothes.is/embed.js`
+| `DATA_DIRECTORY` | Directory for externally provided data | `/via-data`
```
seanh, line 78: I guess if we want to re-use the same approach for another app in the future then the pattern will be `/{APP}-data`. Though it's configurable by an envvar anyway so it can be whatever we want. I wondered about something generic like `/data` but it might clash with an existing directory.

README.md
```diff
@@ -87,6 +87,12 @@ Environment variables:
 | `VIA_HTML_URL` | The URL of the Via HTML instance to redirect to for proxying HTML pages | `https://viahtml.hypothes.is/proxy`
 | `VIA_SECRET` | The secret that must be used to sign URLs to Via's Python endpoints if `SIGNED_URLS_REQUIRED` is on |
 
+Expected data:
+
+The following data is expected to be provided in the `DATA_DIRECTORY`:
+
+ * `google_drive_credentials.json` - A list of credential JSON objects provided by the Google API console if using `GOOGLE_DRIVE_IN_PYTHON`
+
```
seanh, line 95: I think you could just inline this in the table cell above. Would avoid having to jump up and down in the file to read things, and it's okay for the table cell to get bigger

conf/supervisord-dev.conf
```diff
@@ -4,6 +4,7 @@ silent = true
 
 [program:web]
 command=gunicorn -c conf/gunicorn/dev.conf.py --paste conf/development.ini
+environment=DATA_DIRECTORY=".devdata/"
```
seanh: Can we put this in `tox.ini` instead? That's where we put dev envvars:

https://github.com/hypothesis/via/blob/91847a432a58ea5e054b79ce27182f211ee42e72/tox.ini#L20-L34

conf/supervisord.conf
```diff
@@ -13,6 +13,7 @@ stderr_events_enabled=true
 
 [program:web]
 command=newrelic-admin run-program gunicorn via.app:create_app() -b unix:/tmp/gunicorn-web.sock
+environment=DATA_DIRECTORY="/via-data/"
```
seanh: I think this should be set in Elastic Beanstalk, that's where we set all our envvars for QA and prod.

I can see why you might just want to set it in this file: we always want `DATA_DIRECTORY` to be there and we never want it to change (e.g. between QA and prod, or between different deployments). You could probably say the same for a bunch of envvars that we currently set in Elastic Beanstalk.

But there's a confusion cost to having envvars set in multiple places as people would normally look in Elastic Beanstalk for an envvar (that's where they all are, and that's where our docs on Stack Overflow say to find them) and so they'd think this one wasn't set.

tox.ini
```diff
@@ -41,7 +41,7 @@ passenv =
     dev: NGINX_SERVER
     dev: CLIENT_EMBED_URL
     dev: GOOGLE_API_KEY
-    dev: GOOGLE_DRIVE_CREDENTIALS
+    dev: DATA_DIRECTORY
```
seanh: If you add this to `setenv` rather than `passenv` then you can both pass through an external value (if set) and specify a default if not. See the others:

https://github.com/hypothesis/via/blob/91847a432a58ea5e054b79ce27182f211ee42e72/tox.ini#L20-L34

via/services/google_drive.py
```diff
@@ -90,17 +91,41 @@ def iter_file(self, file_id) -> Iterator[ByteString]:
...
-            "The flag 'google_drive_in_python' is enabled but no credentials found"
-        )
-
-    try:
-        return GoogleDriveAPI(service_account_info=json.loads(credentials_json))
-    except JSONDecodeError as exc:
-        raise ConfigurationError("Invalid Google credentials file format") from exc
+    return GoogleDriveAPI(
+        credentials_list=load_injected_json(request, "google_drive_credentials.json")
+    )
```
seanh: I think we want the app to crash _on startup_ if this credentials file is missing or invalid so that it's not possible to deploy the app. Since pyramid_services calls service factories lazily it currently doesn't crash until you try to proxy a Google Drive file. So it'd be possible to deploy the app and the Google Drive endpoint would be broken.

Can we move the `load_injected_json()` code to somewhere that gets executed during app startup?

bin/devdata.py
```diff
@@ -15,10 +15,17 @@ def _get_devdata():
...
+            ("via/devdata.env", ".devdata.env"),
+            (
+                "via/devdata/google_drive_credentials.json",
+                ".devdata/google_drive_credentials.json",
+            ),
+        ):
+            copyfile(
+                os.path.join(git_dir, source),
+                os.path.join(Path(__file__).parent.parent, target),
+            )
```
seanh, line 28: This is crashing if the (empty) `.devdata` dir doesn't exist. Can it create the directory if it doesn't exist? Or just add a `.devdata/.gitkeep` file to git?

seanh (changes requested):
A few minor suggestions and I found two bigger issues:

1. `make devdata` is crashing
2. The app doesn't crash on startup if the credentials data file is missing so it will be possible to deploy the app without this file, but then it _will_ crash whenever the Google Drive endpoint receives a request

README.md
jon-betts: We're about to add another element in here next PR for the resource keys so I think it's better to separate it out. I doubt we'll spend much time looking at this on the average day.

bin/devdata.py
jon-betts: I thought I did that?

jon-betts: I did but it's `gitignored` of course :facepalm:

conf/supervisord.conf
jon-betts: Done.

I think we might want to have a think about "variables" which can't vary. Something about that doesn't seem right.

tox.ini
jon-betts: Done

seanh (approved):

PR #581, opened 2021-09-03
Fix the route URL to include query parameters

jon-betts:
At the moment if you put a URL into the index page which has URL parameters, the parameters are stripped from it. Amongst many other URLs, this means Google Drive URLs don't work.

This PR does two things:

 * Uses `request.path_qs` instead of `request.path` so the query params are included
 * Use `h-vialib` routines to strip off Via params from the URL

### DIFF
--- a/tests/unit/via/views/proxy_test.py
+++ b/tests/unit/via/views/proxy_test.py
@@ -6,20 +6,16 @@ from via.views.proxy import proxy
 
 
 class TestProxy:
-    def test_it(
-        self, pyramid_request, get_url_details, via_client_service, url_from_user_input
-    ):
-        pyramid_request.path = "/https://example.org"
+    def test_it(self, pyramid_request, get_url_details, via_client_service):
+        pyramid_request.path_qs = "/https://example.org?a=1&via.sec=HEX"
 
         result = proxy(pyramid_request)
 
-        url_from_user_input.assert_called_with("https://example.org")
-        pyramid_request.checkmate.raise_if_blocked.assert_called_once_with(
-            url_from_user_input.return_value
-        )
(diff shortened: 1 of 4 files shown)


tests/unit/via/views/proxy_test.py
```diff
@@ -6,20 +6,16 @@
 
 
 class TestProxy:
-    def test_it(
-        self, pyramid_request, get_url_details, via_client_service, url_from_user_input
-    ):
-        pyramid_request.path = "/https://example.org"
+    def test_it(self, pyramid_request, get_url_details, via_client_service):
+        pyramid_request.path_qs = "/https://example.org?a=1&via.sec=HEX"
```
jon-betts, line 10: I've changed this test to actually just show what we do with things instead of mocking out the URL manipulation routines. 

The is a more integrated test, but it actually proves it does something rather than just an assertion about what we call. All of these things are stateless.

seanh: I think this is probably fine given that `url_from_user_input()` is just a nearby helper

via/views/proxy.py
```diff
@@ -8,7 +9,8 @@
 @view_config(route_name="proxy", renderer="via:templates/proxy.html.jinja2")
 def proxy(request):
     # Strip leading '/' and normalize URL.
-    url = url_from_user_input(request.path[1:])
+    url = url_from_user_input(request.path_qs[1:])
+    url = Configuration.strip_from_url(url)
```
seanh: Hmm. I wonder if this call belongs in `url_from_user_input()`

seanh (approved):

via/views/proxy.py
jon-betts: Yep I've moved that in now

PR #582, opened 2021-09-03
Use Google resource key if available in the URL

jon-betts:
This also includes much fancier Google Drive URL parsing as we've seen the URLs have more variation in them than we previously thought.

Based on:

 * https://github.com/hypothesis/via/pull/583
 * https://github.com/hypothesis/via/pull/581
 * https://github.com/hypothesis/via/pull/580
 
Requires:

 * https://github.com/hypothesis/devdata/pull/31

## Testing notes

All of this should be done with caching diabled.

Either drop a debug point into `via/services/google_drive.py` or modify it to log when it adds a resource key:

```diff
--- a/via/services/google_drive.py
+++ b/via/services/google_drive.py
@@ -111,6 +111,9 @@ class GoogleDriveAPI:
             resource_key = self._resource_keys.get(file_id)
 
         if resource_key:
+            from logging import getLogger
+            getLogger(__name__).warning(f"Resource key {resource_key}")
+
             headers["X-Goog-Drive-Resource-Keys"] = f"{file_id}/{resource_key}"
 
         response = self._session.get(url=url, headers=headers, stream=True, timeout=10)
```

You should be able to see the resource key printed out when you visit:

 * http://0.0.0.0:9082/google_drive/0ByrDah1tVd5qMmVzeTF3a0dDOWc/AMADEUPRESOURCEKEY/proxied.pdf
 * The above demonstrates proxying the resource key
 * http://0.0.0.0:9082/google_drive/0ByrDah1tVd5qMmVzeTF3a0dDOWc/proxied.pdf
 * The above demonstrates looking up the value

We should extract the key from URLs. Try pasting these into the root:

 * https://drive.google.com/uc?id=0ByrDah1tVd5qMmVzeTF3a0dDOWc&export=download&resourceKey=AMADEUPRESOURCEKEY
* https://drive.google.com/uc?id=0ByrDah1tVd5qMmVzeTF3a0dDOWc&export=download&resourcekey=AMADEUPRESOURCEKEY

We also accept a wider range of URLs now:

 * https://drive.google.com/file/d/17zLSXuH87GhQJRqTXDxP-vGO--GkeaC8/view?usp=sharing
 * https://drive.google.com/uc?id=0BzxGAA_njUeBSE5wdHlxblJtck0

Try pasting these into the landing page and observe the mapped URL

### DIFF
--- a/via/views/view_pdf.py
+++ b/via/views/view_pdf.py
@@ -28,21 +28,12 @@ def view_pdf(context, request):
 
     google_drive_api = request.find_service(GoogleDriveAPI)
     if google_drive_api.is_available and (
-        file_id := google_drive_api.google_drive_id(url)
+        file_details := google_drive_api.parse_file_url(url)
     ):
-        proxy_pdf_url = request.find_service(SecureLinkService).sign_url(
-            request.route_url(
-                "proxy_google_drive_file",
-                file_id=file_id,
-                # Pass the original URL along so it will show up nicely in
-                # error messages. This isn't useful for users as they don't
-                # see this directly, but it's handy for us.
-                _query={"url": url},
-            )
-        )
+        proxy_pdf_url = ProxyURLBuilder.google_file_url(request, file_details, url)
 
     else:
-        proxy_pdf_url = _pdf_url(
+        proxy_pdf_url = ProxyURLBuilder.nginx_pdf_url(
             url,
             nginx_server=request.registry.settings["nginx_server"],
             secret=request.registry.settings["nginx_secure_link_secret"],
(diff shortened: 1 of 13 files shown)


seanh:
@jon-betts Could you add manual testing instructions to this please?

marcospri (approved):
Looks good :+1: 

For the real effect on google enforcing the resource key param we'll have to wait until the 13th but I think we should merge it now.

jon-betts:
Yes, unfortunately you can't really test it because Google Drive API seems to ignore the header for now. Or at least is indifferent to it being wrong.

via/views/view_pdf.py
```diff
@@ -62,34 +53,53 @@ def view_pdf(context, request):
...
-    sec = b64encode(sec)
-    sec = sec.replace(b"+", b"-")
-    sec = sec.replace(b"/", b"_")
-    sec = sec.replace(b"=", b"")
-    sec = sec.decode()
-
-    # Construct the URL, inserting sec and exp where our NGINX config file
-    # expects to find them.
-    return f"{nginx_server}/proxy/static/{sec}/{exp}/{url}"
+class ProxyURLBuilder:
```
seanh, line 56: :+1: Nice, this seems like a nice helper class

PR #591, opened 2021-09-07
More fine grained error info

jon-betts:
This PR:

 * Unifies the mapping of exceptions between the two handlers
 * Adds the option to format a nice log message from `RequestExceptions`

This should let us:

 * Get better feedback on what's going on
 * Add more cases to the mapping if we want to

### DIFF
--- a/via/requests_tools/error_handling.py
+++ b/via/requests_tools/error_handling.py
@@ -1,6 +1,7 @@
 """Helpers for capturing requests exceptions."""
 
 from functools import wraps
+from logging import getLogger
 
 from requests import exceptions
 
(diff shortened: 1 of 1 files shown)


via/requests_tools/error_handling.py
```diff
@@ -1,27 +1,71 @@
...
+    exception: exceptions.RequestException, error_message
+):
+    """Create a summary string from details from a `RequestsException`."""
+    url = f"'{exception.request.url}'" if exception.request else "unknown URL"
+    status = (
+        "no response"
+        if exception.response is None
+        else f"{exception.response.status_code}: '{exception.response.reason}'"
+    )
+    return f"Requests exception - {type(exception)} Got {status} from {url} ({error_message})"
```
jon-betts, line 39: This error message on the end is often going to be pretty much what we logged, but it means we are independent of the format of requests exceptions messages.

seanh (changes requested):
H'okay, I've fixed a few error reporting issues with Via:

1. It wasn't reporting _any_ exceptions at all to Sentry. Like, even if a straight up bug in our own code raised an uncaught `ValueError` that was not getting reported. There are long-standing bugs in Via that have been hidden as a result. https://github.com/hypothesis/via/pull/593
2. It wasn't logging any exceptions at all to Papertrail either. https://github.com/hypothesis/via/pull/600
   1. There was a second additional reason why `requests`-related exceptions were still not getting logged to Papertrail: https://github.com/hypothesis/via/pull/602
   2. There was a third additional reason why exception logging still wasn't working: https://github.com/hypothesis/via/pull/603
3. It was turning *all* exceptions raised from 500s into 417s: https://github.com/hypothesis/via/pull/601

So any errors in Via were being completely hidden from Sentry, the logs, and the status codes and the rest of our monitoring :)

Exceptions are now being reported to Sentry in production. Example: https://sentry.io/share/issue/0097c55675c9442ba9175e0ceeea0569/

All exceptions are also being logged to Papetrail automatically. Example: https://my.papertrailapp.com/groups/24774711/events?focus=1372174398429212681&selected=1372174398429212681

You don't want to call `log.error()` manually as this is unnecessary and will double-log the exception to Papertrail because the exception is already being logged automatically by `pyramid_exclog`. In addition `sentry_sdk` automatically picks up logged exceptions so it'd double-report it to Sentry as well.

If you want to customize the summary string for an exception the right thing to do is:

1. Catch all requests exceptions and wrap them in a custom exception class. Like LMS does:

   ```python
   try:
       response = self._session.request(...)
       response.raise_for_status()
   except RequestException as err:
       raise HTTPError(err) from err
   ```

2. Add a `__str__()` to your custom exception class. That's where your string summary code goes.

That will mean that you get the new string summary wherever the exception appears, including:

1. When it gets logged to Papertrail automatically by `pyramid_exclog`
2. In Sentry
3. In New Relic (Python exceptions and tracebacks are shown in the **Errors** tab)
4. In any other monitoring tools, present or future, that show exceptions

jon-betts:
Hi @seanh . Thanks for the summary of recent PRs. I'm not quite sure what's actually proposed to be changed here as this is more narative in nature.

Does this PR even apply any more, or would it be worth starting over from where the code is now? I'm guessing it's changed significantly.

seanh:
I think it depends whether you still want to investigate exceptions from Google Drive further? We have captured a couple of them in Sentry now: https://sentry.io/share/issue/a05c95a0553247f98f431fff06b77635/ And they'll be in Papertrail too. But if you want to try and gather more info than is reported by default I think you still need to do something like what you were doing in this PR, it's just that tools-compatible way that you'd do that is slightly different, as I described above. So far we still don't have much more actual info to go on than "read timed out", although I think we can say that they're very rare

jon-betts:
This has been superceded

PR #601, opened 2021-09-08
Don't turn crashes into 417's

### LINKED ISSUE #598: Don't turn crashes into 417's
Via currently turns all crashes (even ones caused by straight bugs in the code) into HTTP 417 responses.

When Via is crashing due to a bug in our code we want that to be a 500 Internal Server Error as it would normally be. We don't want to turn it into a 417. Turning all 500s into 417s will make monitoring systems think the app is healthy when it's broken which could allow bugs to be deployed when the deployment would otherwise fail, or could cause alarms to fail to trigger. It's also misleading to anyone observing and debugging Via.

Note that error/invalid/unexpected responses from third-party sites that Via is proxying are a different issue: if Via tries to make an external request and it fails we may not want to return an HTTP 500 in that case. This issue isn't about that. It's about straight bugs in our own code.

For example this diff will make the Google Drive endpoint crash:

```
index 80be077..ac2c549 100644
@@ -71,6 +71,8 @@ class GoogleDriveAPI:
         :raises HTTPNotFound: If the file id is not valid
         :raises UpstreamServiceError: For other errors
         """
+        raise RuntimeError("Foobar")
+
         # https://developers.google.com/drive/api/v3/reference/files/get
         url = f"https://www.googleapis.com/drive/v3/files/{file_id}?alt=media"
 
```

If you now launch [localhost (make devdata) Google Drive Assignment](https://hypothesis.instructure.com/courses/125/assignments/876) and look for the `/google_drive/` response in dev tools you'll see that it's a 417. This should be a 500 (as should all unexpected errors in general).

seanh:
Fixes https://github.com/hypothesis/via/issues/598

Via currently turns all crashes (even ones caused by straight bugs in the code) into HTTP 417 responses.

When Via is crashing due to a bug in our code we want that to be a 500 Internal Server Error as it would normally be. We don't want to turn it into a 417. Turning all 500s into 417s will make monitoring systems think the app is healthy when it's broken which could allow bugs to be deployed when the deployment would otherwise be automatically backed out, or could cause alarms to fail to trigger. It's also misleading to anyone observing and debugging Via.

Note that error/invalid/unexpected responses from third-party sites that Via is proxying are a different issue: if Via tries to make an external request and it fails we may not want to return an HTTP 500 in that case. This issue isn't about that. It's about straight bugs in our own code.

For example this diff will make the Google Drive endpoint crash:

```
index 80be077..ac2c549 100644
@@ -71,6 +71,8 @@ class GoogleDriveAPI:
         :raises HTTPNotFound: If the file id is not valid
         :raises UpstreamServiceError: For other errors
         """
+        raise RuntimeError("Foobar")
+
         # https://developers.google.com/drive/api/v3/reference/files/get
         url = f"https://www.googleapis.com/drive/v3/files/{file_id}?alt=media"

```

If you now launch [localhost (make devdata) Google Drive Assignment](https://hypothesis.instructure.com/courses/125/assignments/876) and look for the `/google_drive/` response in dev tools you'll see that it's a 417. This should be a 500 (as should all unexpected errors in general).

### DIFF
--- a/via/views/exceptions.py
+++ b/via/views/exceptions.py
@@ -1,12 +1,7 @@
 """Error views to handle when things go wrong in the app."""
 
 import h_pyramid_sentry
-from pyramid.httpexceptions import (
-    HTTPClientError,
-    HTTPError,
-    HTTPExpectationFailed,
-    HTTPNotFound,
-)
+from pyramid.httpexceptions import HTTPClientError, HTTPError, HTTPNotFound
 from pyramid.view import exception_view_config
 
 from via.exceptions import BadURL, UnhandledException, UpstreamServiceError
@@ -79,8 +74,7 @@ def all_exceptions(exc, request):
     try:
         status_code = exc.status_int
     except AttributeError:
-        # This is our 501, chosen to not scare Cloudflare.
-        status_code = HTTPExpectationFailed.code  # 417
+        status_code = 500
 
     request.response.status_int = status_code
 
--- a/tests/unit/via/views/exceptions_test.py
+++ b/tests/unit/via/views/exceptions_test.py
@@ -26,7 +26,7 @@ class TestErrorView:
                 HTTPUnsupportedMediaType.code,
                 id="Pyramid error",
             ),
-            param(BlockingIOError, 417, id="Unknown python error"),
+            param(BlockingIOError, 500, id="Unknown python error"),
         ),
     )
     def test_values_are_copied_from_the_exception(


marcospri (approved):

via/views/exceptions.py
```diff
@@ -79,8 +74,7 @@ def all_exceptions(exc, request):
     try:
         status_code = exc.status_int
     except AttributeError:
-        # This is our 501, chosen to not scare Cloudflare.
```
robertknight, line 82: What was the "don't scare Cloudflare" scenario that was a concern here? Is it still relevant in any way?

seanh: I don't think this comment was ever correct. 501 is 501 Not Implemented. The comment must have meant 502 Gateway Error. In the past people have sometimes thought that, when a request to a third-party fails, it would be semantically correct for us to return 502 Gateway Error. But this has been found to cause a problem in the past: Cloudflare replaced 502 responses from us with their own Gateway Error page. That must be what this comment referred to. But the `except` block that contains this comment is never entered when handling an exception related to an upstream request, so the comment makes no sense here.

PR #613, opened 2021-09-09
Make the resource keys mandatory for Google Drive

jon-betts:
For: https://github.com/hypothesis/lms/issues/3090

## Testing notes

 * `rm .devdata/google_drive_resource_keys.json`
 * `make dev`
 * Explode!
 * This would work in `master`
 * (`make devdata` to fix)


tests/conftest.py
```diff
@@ -17,7 +17,7 @@ def pyramid_settings():
         "checkmate_ignore_reasons": None,
         "checkmate_allow_all": False,
         "enable_front_page": True,
-        "data_directory": ".devdata",
+        "data_directory": "tests/data_directory",
```
jon-betts, line 20: This change is needed because this directory doesn't exist in CI, and so the functests fail to start. We don't actually need much to be in those files on startup, but we do need them to exist.

marcospri (approved):

Issue #612, opened 2021-09-09
Filter out exceptions from `route_by_content()` talking to third-parties


seanh:
When the `route_by_content()` view [sends requests to third-parties](https://github.com/hypothesis/via/blob/38f0c85e679f93b3d112a92437ca84aee74ff98e/via/views/route_by_content.py#L17) we don't want exceptions from those requests to get logged. They're just a lot of noise in our monitoring systems.

### Done when

- [ ] Exceptions from `route_by_content()` talking to third-parties don't get reported to Slack
- [ ] Exceptions from `route_by_content()` talking to third-parties don't get logged to Papertrail
- [ ] **All** exceptions from Via talking to Google Drive do still get logged to both Papertrail and Slack.

  `route_by_content()` [doesn't send requests to Google Drive](https://github.com/hypothesis/via/blob/38f0c85e679f93b3d112a92437ca84aee74ff98e/via/get_url.py#L27-L28) because it just assumes that drive.google.com URLs are PDFs.

  So we just need to make sure that all the requests [sent by `GoogleDriveAPI`](https://github.com/hypothesis/via/blob/38f0c85e679f93b3d112a92437ca84aee74ff98e/via/services/google_drive.py#L125-L128) are still getting reported.

### Implementation notes

We probably want to make it raise one exception class for `route_by_content()` errors and a different exception class for Google Drive proxying errors, as that's what's going to make filtering from both tools easiest.

`pyramid_exclog` is the thing that's logging these exceptions to Papertrail. It has an `exclog.ignore` setting for filtering out exceptions: <https://docs.pylonsproject.org/projects/pyramid_exclog/en/latest/#settings>.

`h-pyramid-sentry` is the thing that's logging these exceptions to Sentry. It has an `h_pyramid_sentry.filters` setting for filtering things out: <https://github.com/hypothesis/h-pyramid-sentry#filters>. Also see Sentry's [Filtering for Python](https://docs.sentry.io/platforms/python/configuration/filtering/) and [Inbound Filters](https://docs.sentry.io/product/data-management-settings/filtering/) docs for other ways.

jon-betts:
This is quite easily done with a different exception handler. 

Exceptions raised by 3rd party things are all requests exceptions. These are mapped to our three exception types. So if we don't call `exc_log` on them, then we don't need to filter them back out again.

seanh:
You mean if we don't call `report_exception()`?

jon-betts:
[...]

PR #614, opened 2021-09-09
Filter which exceptions get reported to Sentry

jon-betts:
For: https://github.com/hypothesis/via/issues/612

Currently we over report issues, which means for public Via in particular Sentry will get filled up with lots of things to do with 3rd parties going wrong.

This PR:

 * Adds a separate view for Google Drive
 * Doesn't report on errors we aren't interested in

## Testing notes

 * Get the Sentry DSN listed here: https://sentry.io/settings/hypothesis/projects/sentry-test-environment/keys/
 * `SENTRY_DSN=<dsn-here> make dev`
 * BREAK STUFF!
 * Specifically you can try: http://0.0.0.0:9082/google_drive/NOTATHING/proxied.pdf
 * You should see _non_ filtered exceptions show up here: https://sentry.io/organizations/hypothesis/projects/sentry-test-environment/?project=5952630

### DIFF
--- a/via/exceptions.py
+++ b/via/exceptions.py
@@ -10,7 +10,7 @@ class UpstreamServiceError(Exception):
     status_int = 409
 
 
-class UnhandledException(Exception):
+class UnhandledUpstreamException(Exception):
     """Something we did not plan for went wrong."""
 
     status_int = 417
--- a/via/views/exceptions.py
+++ b/via/views/exceptions.py
@@ -1,10 +1,10 @@
 """Error views to handle when things go wrong in the app."""
 
 import h_pyramid_sentry
-from pyramid.httpexceptions import HTTPClientError, HTTPError, HTTPNotFound
+from pyramid.httpexceptions import HTTPClientError, HTTPNotFound
 from pyramid.view import exception_view_config
 
-from via.exceptions import BadURL, UnhandledException, UpstreamServiceError
+from via.exceptions import BadURL, UnhandledUpstreamException, UpstreamServiceError
 
 EXCEPTION_MAP = {
     BadURL: {
@@ -24,7 +24,7 @@ EXCEPTION_MAP = {
         "stage": "upstream",
         "retryable": True,
     },
-    UnhandledException: {
+    UnhandledUpstreamException: {
         "title": "Something went wrong",
         "long_description": ["We experienced an unexpected error."],
         "stage": "via",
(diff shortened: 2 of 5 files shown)


via/views/exceptions.py
```diff
@@ -61,16 +61,44 @@ def _get_meta(exception):
...
+    Exception,
+    route_name="proxy_google_drive_file:resource_key",
+    renderer="via:templates/exception.html.jinja2",
+)
+def google_drive_exceptions(exc, request):
+    """Catch all errors for Google Drive and display an HTML page."""
+
+    # Don't log 404's as they aren't very interesting
+    if not isinstance(exc, HTTPNotFound):
+        h_pyramid_sentry.report_exception()
```
seanh: Can we continue logging 404s for now?

I think not logging those is slightly controversial as a 404 could mean "user tried to proxy a deleted file" or it could mean something else. Maybe we need to be a little bit more cautious and actually look in the response body and positively identify it as a "file does not exist". Even then we might be getting that error from Google due to a bug in our code (if something changed meaning our code for extracting IDs no longer works). Maybe make it a separate PR so we can discuss it separately.

via/views/exceptions.py
```diff
@@ -61,16 +61,44 @@ def _get_meta(exception):
...
 @exception_view_config(Exception, renderer="via:templates/exception.html.jinja2")
-@exception_view_config(HTTPError, renderer="via:templates/exception.html.jinja2")
 def all_exceptions(exc, request):
     """Catch all errors (Pyramid or Python) and display an HTML page."""
 
-    h_pyramid_sentry.report_exception()
+    # We don't want to log errors from upstream services or things which are
+    # the user goofing about making bad queries.
+    if not isinstance(
+        exc, (UpstreamServiceError, UnhandledUpstreamException, BadURL, HTTPClientError)
```
seanh, line 92: I think it would be a good idea for these four to share a common base class like `ExternalRequestError` or something. Then you wouldn't need to list them all here, and if a new type of exception gets added you won't need to remember to add it here (assuming it is a subclass).

seanh: `other_exceptions()` may be a more accurate name for `all_exceptions()` now

via/exceptions.py
```diff
@@ -10,7 +10,7 @@ class UpstreamServiceError(Exception):
     status_int = 409
 
 
-class UnhandledException(Exception):
+class UnhandledUpstreamException(Exception):
```
seanh, line 13: Is this class really only used for upstream-related errors, and nothing else?

via/views/exceptions.py
seanh: Also, does filtering out `HTTPNotFound` actually remove the 404s? Is that what gets raised?

seanh (changes requested):
:+1: I think the general approach of using `route_name` to separate two exception views is good.

I think you might want to introduce a shared base class into the exception hierarchy.

I think you should at least move the filtering of 404s into a separate PR as that's a separate issue and one that is controversial, so don't tangle it up with this.

seanh:
The downside of using `route_name` is it's more easy to break things: you could rename the route and forget to fix the `route_name` predicate and the code wouldn't crash but you'd have made an unintended change. It's on the developer to grep for the route name they've changed and find it in the exception view. Whereas if you use exception classes you can't rename a class and forget to rename it in the exception views 'cos it'd crash. So I think different-exception-classes is the best practice for exception views. But I don't think we rename routes often so I think the approach in this PR is acceptable. Especially since it might be practically expedient given the way that the request-sending code in Via is structured

via/views/exceptions.py
jon-betts: I was working on this separately: https://github.com/hypothesis/via/pull/617

via/views/exceptions.py
jon-betts: It's a little awkward here. The two upstream ones definitely could. `BadURL` is sometimes raised as part of an upstream thing and sometimes locally. `HTTPClientError` just isn't ours at all.

I don't think it's too bad to just list the things we want here. It has the virtue of being dumb and clear for now.

via/exceptions.py
jon-betts: Yep 100%. It's only raised as a result of `RequestException`'s. The name was overly broad before.

via/views/exceptions.py
jon-betts: Renamed to `other_exceptions`

marcospri (approved):
Nice, I think this is probably the most pyramid-y way of handling this :+1:

PR #615, opened 2021-09-09
Add longer and more timeouts to Google Drive

jon-betts:
This is to try and cut down on the number of documents that hang because Google is being slow.

This also possibly prevents us from other situations where Google might run out the clock and cause us to hit, for example, a Gunicorn exception.


seanh (commented):
:+1: I haven't tested this so let's get a proper review from @marcospri but the approach looks good to me

marcospri (approved):

PR #617, opened 2021-09-09
Add 404 handling to Google Drive

jon-betts:
This distinguishes between the 404's we know are to do with the file being gone and other potentially much worse ones. This is quite specific to the format of errors we see right now. If that changes we'll start seeing these as more serious errors.

On the whole it's probably better we err on the side of flipping out instead of potentially concealing errors.


marcospri (approved):
Tested with a non existing ID and just changing the URL we download the files from.

PR #618, opened 2021-09-09
Handle Google Drive throttling us (403)

jon-betts:
This catches the case we are getting throttled and handles it separately from other errors so we can tell it's happening.

We return a 429 in this case (Too many requests).

### Testing notes

Install `gevent` into a tox env and then run this in the root.

```python
import gevent.monkey
gevent.monkey.patch_all()

import json
import gevent

from via.exceptions import GoogleDriveServiceError
from via.services import GoogleDriveAPI


with open('.devdata/google_drive_credentials.json', encoding='utf-8') as handle:
    api = GoogleDriveAPI(
        credentials_list=json.load(handle),
        resource_keys={}
    )


def worker():
    try:
        list(api.iter_file("1dkuAdpPEHwWUGfDMwC4a53JvX8fnSxwL"))
    except GoogleDriveServiceError as err:
        print(type(err), err)
        print(err.error_json)
        print(err.status_int)
        exit()


threads = []
for i in range(20):
    threads.append(gevent.spawn(worker))

gevent.joinall(threads)
```

### DIFF
--- a/tests/unit/via/services/fixtures/google_403.json
+++ b/tests/unit/via/services/fixtures/google_403.json
@@ -1,0 +1,14 @@
+{
+    "error": {
+        "errors": [
+            {
+                "domain": "usageLimits",
+                "reason": "userRateLimitExceeded",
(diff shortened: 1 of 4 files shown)


marcospri (approved):
We'll see very clearly what's wrong if we have a similar issue again. Nothing to add :+1:

Issue #547, opened 2021-07-29
Fix error page to display correct URL when proxying


jon-betts:
Prior to Via3 becoming Via and the public site, all pages used URL params to convey which document was being proxied. We rely on this being true to display the URL in the error page.

The new proxy end-point however uses a URL in the URL path instead, which means the error details are wrong, and have a blank proxied URL.

## Related PRs:

 * https://github.com/hypothesis/via/pull/621
 * https://github.com/hypothesis/via/pull/622
 * https://github.com/hypothesis/via/pull/627

PR #621, opened 2021-09-13
Unify URL parsing and handling

jon-betts:
When the new proxying stuff was added to Via it added a parallel way of getting the URL without using the existing resource object.

This merges the two types of URL access into one. This is:

 * With the hopes it's easier to grok and maintain
 * Because I think it might help us with: https://github.com/hypothesis/via/issues/547

### Changes in behavior

This does result in some behavior changes, which I think are good:

 * We now perform URL normalisation on all end-points
 * We now raise exceptions when the URL is missing and we don't expect that

These aren't likely to be practically interesting for various reasons, but that's not the aim of the PR.

### Testing notes

I don't think the error handling makes a difference, as it's not currently logically possible to call the proxy end-point without a URL as that _is_ the index URL, and in the case of index we catch the error. So I don't think you can get in that situation, but some things to prove the roof is still on:

 * `make dev`
 * Start ViaHTML
* http://localhost:9083/http://example.com/ - A normal working webpage
* http://localhost:9083/pdf?url=https%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1heF_Yjs_ZZyms4QpdG0P06fTEXWkMdVW%26export%3Ddownload - A Google Drive PDF

### New behavior:

All of our PDF end-points and our routing now can handle a missing schema etc. This is also unlikely to particularly important as these tend to come from systems which are more likely to control the URL and ensure it's well formed.

* http://localhost:9083/pdf?url=drive.google.com%2Fuc%3Fid%3D1heF_Yjs_ZZyms4QpdG0P06fTEXWkMdVW%26export%3Ddownload - URL without scheme


marcospri (approved):
Nice, there was a lot of guesswork on what to use to get the URL from anytime you change anything on via :+1:

PR #622, opened 2021-09-13
URL normalisation improvements

jon-betts:
Discovered during: https://github.com/hypothesis/via/pull/621

There are quite a few cases where the naive approach here fails. I've added in the solution we have in Checkmate for this:

https://github.com/hypothesis/checkmate/blob/main/checkmate/url/canonicalize.py#L83

It's not super intuitive as `urlparse`'s behavior is a bit strange.

These aren't common cases, but if a user is quickly copy pasting URLs they could easily miss bits. We can't account for everything of course.

## Not that important?

It turns out some of this isn't hugely consequential, because we accidentally fix it else where in the code by re-parsing the URL. This converts things like:

 * `http:////example.com` to `http://example.com`

Even if we emit that it would probably work in most browsers. However it does seem nice that our code works on purpose rather than as a side effect of how we process the URL later on. If we skipped or changed that part it might stop working or behave strangely.

### Testing notes

It's easy enough to test the other changes though:

 * `make dev`
 * Goto the landing page: http://0.0.0.0:9082
 * Paste in`ftp://example.com`
 * In master we get an error page about not being able to retrieve `https://ftp://example.com`
 * In this branch you get a "BadURL" message saying your URL is wrong (which is better)

### DIFF
--- a/via/resources.py
+++ b/via/resources.py
@@ -1,6 +1,10 @@
 """Context objects for views."""
+from urllib.parse import urlparse
+
 from h_vialib import Configuration
 from pyramid.httpexceptions import HTTPBadRequest
+
+from via.exceptions import BadURL
 
 
 class URLResource:
@@ -13,7 +17,8 @@ class URLResource:
         """Get the 'url' parameter from the query.
 
         :return: The URL as a string
-        :raise HTTPBadRequest: If the url is missing or empty
+        :raise HTTPBadRequest: If the URL is missing or empty
+        :raise BadURL: If the URL is malformed
         """
         try:
             url = self._request.params["url"].strip()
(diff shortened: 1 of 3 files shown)


via/resources.py
```diff
@@ -45,7 +51,18 @@ def _normalise_url(cls, url):
...
+
+        try:
+            parsed = urlparse(url)
+        except ValueError as exc:
+            raise BadURL(url) from exc
+
+        if not parsed.scheme:
+            if not parsed.netloc:
+                # Without a scheme urlparse often assumes the domain is the
+                # path. To prevent this add a fake scheme and try again
```
marcospri, line 63: I'm 100% sure there's a very good explanation for this behaviour on urlparse but this always surprises me.

marcospri (approved):
Not much to add, much nicer to catch this explicitly that to let it buble all the way down :+1:

via/resources.py
jon-betts: I think it's because it's logically impossible to tell the difference between a bare host and a path, so they had to arbitrarily choose one behavior:

```
# Totally cool!
http://example.com/path/localhost
http://localhost/example.com/path
http://path/localhost/example.com

# Erm...
example.com/path/localhost
localhost/example.com/path
path/localhost/example.com

# Um...
path/localhost
example.com/path
localhost/path

# Gah...
path
example.com
localhost
```

PR #624, opened 2021-09-15
Add error handling for a 403 from Google that actually means forbidden

jon-betts:
Rather than the usual 403 which means rate limited. As we now know they re-use codes for different meanings I've also added descriptions to the fixture files to distinguish them.


via/services/google_drive.py
```diff
@@ -46,6 +46,16 @@ def translate_google_error(error):
             requests_err=error,
         )
 
+    if status_code == 403 and reason == "cannotDownloadFile":
+        # This seems to happen if a file is locked down in various ways
+        # rather than for general consumption
+        return GoogleDriveServiceError(
+            "We do not have permission to access the file through the "
```
marcospri: nitpicking here I think, maybe replace "access" by "download"?

marcospri (approved):

via/services/google_drive.py
jon-betts: Nitpicking is what we do

PR #626, opened 2021-09-15
Add details of our error codes

jon-betts:
This documents the codes we raise at the current time and why


indigobravo (approved):
Looks good. Approved.

PR #627, opened 2021-09-15
Get the correct original URL for the error page for all views

jon-betts:
For:

 * https://github.com/hypothesis/via/issues/547

Prior to this PR the error handling only displays the correct URL for views which expect a URL in the query parameter. This didn't work for proxying end-point.

This PR:

 * Uses resources to declare whether each view expects:
    * A URL in the path: `PathURLResource`
    * A URL in the query: `QueryURLResource` 
    * No URL at all: no resource
 * Uses the context type to retrieve the correct URL in the error handler

### DIFF
--- a/via/resources.py
+++ b/via/resources.py
@@ -7,42 +7,11 @@ from pyramid.httpexceptions import HTTPBadRequest
 from via.exceptions import BadURL
 
 
-class URLResource:
+class _URLResource:
     """Methods for routes which accept a 'url'."""
 
     def __init__(self, request):
         self._request = request
-
-    def url_from_query(self):
-        """Get the 'url' parameter from the query.
-
-        :return: The URL as a string
-        :raise HTTPBadRequest: If the URL is missing or empty
-        :raise BadURL: If the URL is malformed
-        """
-        try:
-            url = self._request.params["url"].strip()
-        except KeyError as err:
-            raise HTTPBadRequest("Required parameter 'url' missing") from err
-
-        if not url:
-            raise HTTPBadRequest("Required parameter 'url' is blank")
-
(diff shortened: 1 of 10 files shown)


via/resources.py
```diff
@@ -44,25 +54,41 @@ def url_from_path(self):
...
+
+def get_original_url(context):
+    """Get the original URL from a provided context object (if any)."""
+
+    if isinstance(context, QueryURLResource):
+        getter = context.url_from_query
+    elif isinstance(context, PathURLResource):
+        getter = context.url_from_path
+    else:
+        return None
```
jon-betts, line 87: There's fancier ways of doing this, like having both the contexts have a method called `url` and then call that. This has the benefit of having a very clear, easy to read and grep for name in the views that call these.

If you see `url_from_path` you know exactly what's going on.

via/resources.py
```diff
@@ -44,25 +54,41 @@ def url_from_path(self):
...
+
+    try:
+        return getter()
+    except BadURL as err:
+        return err.url
+    except HTTPBadRequest:
+        return None
```
jon-betts, line 94: We could try and facilitate this with a `dont_raise=True` or something in the methods, but we're unlikely to be able to write that more concisely than this.

Not sure if we should be catching `Exception` here instead though?

marcospri (approved):

via/resources.py
marcospri: Instead of `HTTPBadRequest`? I think we rather see them bubble up to sentry and check there if we find something interesting.

jon-betts: We should report them anyway in the error handling so this should only be if our serialisation fails? I guess we'll see!

PR #630, opened 2021-09-17
Handle request objects without contexts, which can happen apparently

jon-betts:
For: https://sentry.io/organizations/hypothesis/issues/2653839684/?project=5921964&query=is%3Aunresolved

Apparently we can get in the situation where the request object has no context attribute. I expected it to be `None` sometimes, but not absent.


marcospri (approved):
Looking at the error in sentry, which seems to be a malformed request, I guess that not all pyramid machinery it's at play yet at that moment. 

This will fix the error handling :+1:

PR #631, opened 2021-09-17
Add `pyramid-sanity` to Via to catch UTF-8 and other errors

jon-betts:
For many UTF-8 errors like this: https://sentry.io/organizations/hypothesis/issues/2635391672/?project=5921964&query=is%3Aunresolved

## Testing notes

I'm not going to list everything as it's covered what `pyramid-sanity` fixes here, but here's one which you can try:

 * `make dev`
 * http://localhost:9083/route?q=%FC
 * In this branch you should get a 400
 * In master you get a 500

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -153,12 +153,15 @@ pyramid==2.0
     #   pyramid-exclog
     #   pyramid-ipython
     #   pyramid-jinja2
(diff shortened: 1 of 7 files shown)


marcospri (approved):
Nice work removing noise from sentry :+1:

PR #632, opened 2021-09-17
Handle URLs with no scheme and a leading dot

jon-betts:
It seems reasonably common for users to type in a URL without a scheme but with a leading dot:

* https://sentry.io/organizations/hypothesis/issues/2648094577/?project=5921964&query=is%3Aunresolved

For example `.tumbler.com`. We add `https://` to the front and create an unparsable URL`https://.tumbler.com`

## Testing notes

* `make dev`
* Goto the landing page http://localhost:9083
* Type in a URL with no scheme and a leading dot
* We should now attempt to go to the URL with the dot stripped
* Previously we would crash

### DIFF
--- a/tests/unit/via/resources_test.py
+++ b/tests/unit/via/resources_test.py
@@ -14,6 +14,7 @@ NORMALISED_URLS = [
     ("http://example.com?a=1", "http://example.com?a=1"),
     # URLs without a protocol that should have `https://` prefixed
     ("example.com", "https://example.com"),
+    (".example.com", "https://example.com"),
     ("//example.com", "https://example.com"),
     # Leading and trailing whitespace that should be stripped
     (" http://example.com", "http://example.com"),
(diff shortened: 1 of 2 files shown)


tests/unit/via/resources_test.py
```diff
@@ -8,18 +8,19 @@
...
+    # ("http://example.com?a=1", "http://example.com?a=1"),
+    # # URLs without a protocol that should have `https://` prefixed
+    # ("example.com", "https://example.com"),
+    (".example.com", "https://example.com"),
+    # ("//example.com", "https://example.com"),
+    # # Leading and trailing whitespace that should be stripped
+    # (" http://example.com", "http://example.com"),
+    # ("http://example.com ", "http://example.com"),
+    # # Check we remove Via stuff
+    # ("http://example.com?a=1&via.sec=TOKEN", "http://example.com?a=1"),
```
marcospri: This was all commented probably testing locally

jon-betts: Errr yup!

jon-betts: This is now less comically broken.

marcospri (approved):

PR #633, opened 2021-09-17
Simplify PDF viewer initialization

robertknight:
This removes logic that is no longer needed with the current Hypothesis client, PDF.js version and supported browsers.

 - Remove references to IE 11. It hasn't been supported in over a year. The current baseline for the client is Chrome 70, Firefox 70, Safari 11.1.
 - Remove `PromiseRejectionEvent` reference. All of our target browsers support this API natively.
 - Remove `eventBusDispatchToDOM` option. It isn't needed with the current Hypothesis client
 - Use `PDFViewerApplication.initializedPromise` to simplify waiting for PDF.js to be ready.

See also https://github.com/hypothesis/browser-extension/pull/658


jon-betts (approved):
This is certainly a lot simpler! I've tested this locally and it seems to work. You might want to get a second opinion from someone on the front-end for a sanity check on the Javascript stuff?

Issue #635, opened 2021-09-24
Google Drive documents with zero length cause errors


jon-betts:
See: https://sentry.io/organizations/hypothesis/issues/2664484416/?project=1881822&query=is%3Aunresolved

When a document has zero length we encounter a `StopIteration` exception, but don't handle it correctly (_suspected cause, not proven_)

seanh:
I think this is done

PR #636, opened 2021-09-27
Handle zero length documents returned from Google Drive.

jon-betts:
With zero length documents the first chunk doesn't exist and we immediately get a `StopIteration` exception.

See: 

 * https://github.com/hypothesis/via/issues/635

## Testing notes

 * `make dev`
 * Try and view: http://localhost:9083/https://drive.google.com/uc?id=1d8JdZSYU-fgFuMQpS2Y9YoMtkQTBCAtD&export=download
 * You should now get an error: "Invalid or corrupted PDF file"
 * In master this is "Unexpected server response." (disable caching if you want to try)


marcospri (approved):
Looks good, I was looking for the error on pdfjs but on our own response before.

PR #637, opened 2021-09-28
Handle variations of URLs encoded with % encoding

jon-betts:
For: https://sentry.io/organizations/hypothesis/issues/2678991846/?project=5921964&query=is%3Aunresolved

This is a natural consequence of proxying a URL with unicode high chars in it from the interface. For example a URL with Cyrillic. But could also happen if someone were doing something nefarious.

This is _really_ starting to look like the logic from Google in their URL normalisation. I suspect we're slowly going to rediscover why they did everything. Our aims are slightly different through. They want to normalise for URL matching, we just want something which `requests` can call without crashing.

## What's going on?

When the user enters a URL with high valued unicode chars, these get URL encoded by something in the Gunicorn / Pyramid industrial complex. The result is that we receive a URL which has all components of the path URL encoded. This is a URL that Chrome would accept and display to you correctly, but requests can't parse and understand it.

The solution in this PR is to repeatedly decode the URL until the parsed net location doesn't change.

## Testing notes

 * Run ViaHTML
 * `make dev`
 * Visit http://localhost:9083/http://сказки-андерсена.рф/калоши-счастья
 * This should work in this branch
 * In master you should see a `LocationParseError` which originates in requests.

If you want you can also try encoding, double or triple encoding. We should also be able do all the same tricks we can with missing schemes etc.

### DIFF
--- a/tests/unit/via/resources_test.py
+++ b/tests/unit/via/resources_test.py
@@ -21,7 +21,36 @@ NORMALISED_URLS = [
     ("http://example.com ", "http://example.com"),
     # Check we remove Via stuff
     ("http://example.com?a=1&via.sec=TOKEN", "http://example.com?a=1"),
+    # If we have URLs with unicode in them, then by the time they get to us
+    # they can look like this. A browser will understand and decode this, but
+    # requests can't.%D0%BA%D0%B0%D0%BB%D0%BE%D1%88%D0%B8
+    (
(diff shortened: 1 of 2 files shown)


marcospri (approved):
Tested with the russian one and a few other unicode ones

Doesn't seem to work with emoji domains... but we could probably wait for the sentry report for those ones...  `http://localhost:9083/https://i%E2%9D%A4.ws`

PR #638, opened 2021-09-29
Return a 204 no content when Google Drive files are empty.

jon-betts:
For: 

 * https://github.com/hypothesis/via/issues/635

This means we serve a 204 No Content when the google files are empty. This should prevent Cloudflare from caching the response.

### Testing notes

 * `make dev`
 * Visit http://localhost:9082/google_drive/1d8JdZSYU-fgFuMQpS2Y9YoMtkQTBCAtD/proxied.pdf?url=https%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1d8JdZSYU-fgFuMQpS2Y9YoMtkQTBCAtD%26export%3Ddownload
 * It's hard to see as nothing happens, but look at the dev console and you should see a 204

### DIFF
--- a/tests/unit/via/views/google_drive_test.py
+++ b/tests/unit/via/views/google_drive_test.py
@@ -1,17 +1,19 @@
 from unittest.mock import sentinel
 
 import pytest
+from pyramid.httpexceptions import HTTPNoContent
 
 from via.views.google_drive import proxy_google_drive_file
 
 
 @pytest.mark.usefixtures("secure_link_service", "google_drive_api")
 class TestGetFileContent:
-    def test_it_adds_headers(
+    def test_status_and_headers(
         self, pyramid_request, secure_link_service, google_drive_api
     ):
         response = proxy_google_drive_file(pyramid_request)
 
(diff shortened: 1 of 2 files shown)


tests/unit/via/views/google_drive_test.py
```diff
@@ -12,6 +13,7 @@ def test_it_adds_headers(
     ):
         response = proxy_google_drive_file(pyramid_request)
 
+        assert response.status_code == 200
```
jon-betts, line 16: Added a test here to make sure we don't return 204's all the time

seanh: The test name `test_it_adds_headers()` might be inaccurate now. I don't think the status code is technically a header. Could be as simple as `test_status_and_headers()`

seanh (approved):

tests/unit/via/views/google_drive_test.py
jon-betts: Yep. Good catch. It's definitely not a header

PR #640, opened 2021-10-04
Don't crash on unexpected JSON formats from Google

seanh:
Don't crash if Google Drive sends us an error response with a JSON body whose `"errors"` list is not a list or whose `"errors"[0]` dict is not a dict.

FYI: this is why we use Marshmallow to parse responses in LMS. Python code like this that parses JSON responses gets complex (`translate_google_error()` is now tripping a PyLint `too-many-return-statements` and is close to tripping PyLint's `too-complex`) and error-prone (an attempt was clearly made to parse the response robustly but the code nonetheless contained at least two crashers). LMS once had exactly the same problem: in lots of places it had Python code to parse Canvas responses and there were dozens of bugs in this code.

The solution in LMS was to use Marshmallow to parse responses before using them. For example [`OAuthTokenResponseSchema`](https://github.com/hypothesis/lms/blob/2e2113e425fd4b5f9d46fe655bf0aad917eb8545/lms/validation/authentication/_oauth.py#L136-L146) parses OAuth token responses and then `OAuthHTTPService` simply [uses the schema to validate a response](https://github.com/hypothesis/lms/blob/eb27a4e5e562b558579e8adf069662ab2a06e01c/lms/services/oauth_http.py#L118-L126) and then assumes that `validated_data` is a dict containing the expected keys, value types, etc. `OAuthHTTPService` doesn't even catch the `ValidationError` that `OAuthTokenResponseSchema` might throw: there's an exception view for those.

(Part of the reason for the choice of Marshmallow was that it can parse any part of any type of request or response. E.g. query params, XML, form submissions, etc. Not just JSON bodies.)

### DIFF
--- a/via/services/google_drive.py
+++ b/via/services/google_drive.py
@@ -16,27 +16,36 @@ from via.requests_tools.error_handling import iter_handle_errors
 LOG = getLogger(__name__)
 
 
-def translate_google_error(error):
+def translate_google_error(
+    error,
+):  # pylint:disable=too-many-return-statements
     """Get a specific error instance from the provided error or None."""
 
     # This isn't a requests exception we can get meaningful data from
     if not isinstance(error, HTTPError) or error.response is None:
         return None
 
+    status_code = error.response.status_code
+
     # Try and parse out the Google details in the format we've seen
     try:
         google_error = error.response.json()["error"]["errors"][0]
-    except (JSONDecodeError, KeyError):
+    except (JSONDecodeError, KeyError, TypeError):
         return None
 
-    status_code, reason = error.response.status_code, google_error.get("reason")
+    try:
+        google_reason, google_message = google_error.get("reason"), google_error.get(
+            "message"
+        )
+    except AttributeError:
+        return None
 
     # Check carefully to see that this is Google telling us the file isn't
     # found rather than this being us going to the wrong end-point
-    if status_code == 404 and reason == "notFound":
-        return HTTPNotFound(google_error.get("message", "File id not found"))
+    if status_code == 404 and google_reason == "notFound":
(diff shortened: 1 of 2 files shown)


via/services/google_drive.py
```diff
@@ -16,27 +16,36 @@
...
     if not isinstance(error, HTTPError) or error.response is None:
         return None
 
+    status_code = error.response.status_code
+
     # Try and parse out the Google details in the format we've seen
     try:
         google_error = error.response.json()["error"]["errors"][0]
-    except (JSONDecodeError, KeyError):
+    except (JSONDecodeError, KeyError, TypeError):
```
seanh, line 33: The `[0]` raises `TypeError` if `error.response.json()["error"]["errors"]` is not a list (for example maybe Google sometimes returns `null` here)

via/services/google_drive.py
```diff
@@ -16,27 +16,36 @@
...
         return None
 
-    status_code, reason = error.response.status_code, google_error.get("reason")
+    try:
+        google_reason, google_message = google_error.get("reason"), google_error.get(
+            "message"
+        )
+    except AttributeError:
+        return None
```
seanh, line 41: The `.get()`'s raise `AttributError` if `google_error` is not a dict (for example maybe Google sometimes returns a string here)

jon-betts (approved):
Yep, looks good.

It could be worth trying it with marshmallow? If it's more compact it could be good.

seanh:
Here's a draft of a Marshmallow version: https://github.com/hypothesis/via/pull/643 ([Slack thread](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1633622693177600))

PR #643, opened 2021-10-07
Use Marshmallow to validate Google Drive error bodies

seanh:
Use Marshmallow instead of plain Python code to validate the JSON bodies of error responses from Google Drive.

The downside of this is that it's more code: the Marshmallow schema is more lines of code than it takes to parse this manually in Python.

It also requires knowledge of Marshmallow to understand the schema.

The benefits include:

* It separates validating the JSON body from taking different actions based on the contents of the valid JSON body. A single more complex function is split into two simpler ones.

* The Marshmallow schema has its own direct unit tests. This just allows us to be more thorough about testing the validation and makes bugs less likely.

* The `translate_google_error()` function can now just call the validation schema and then go ahead and assume that the data is as expected without worrying about crashing

You could get the same benefits without Marshmallow by just writing a separate plain Python function to do the validation. But I think Marshmallow is worthwhile here. It does require you to learn Marshmallow but you already have to learn that anyway because we use it elsewhere. And once you know Marshmallow I think it makes it easier to write the validation code. It's actually pretty tricky to write this kind of validation code in plain Python and not make any mistakes. To illustrate, here's a plain Python function that passes all the same tests as the Marshmallow schema does. It's actually about the same number of lines of code as the Marshmallow schema. But I think it's much trickier to write:

```python
def validate_google_error(error):
    try:
        json_data = error.response.json()
    except JSONDecodeError as err:
        raise ValueError() from err

    if not isinstance(json_data.get("error"), dict):
        raise ValueError()

    if not isinstance(json_data["error"].get("errors"), list):
        raise ValueError()

    if not json_data["error"]["errors"]:
        raise ValueError()

    for key in ("message", "reason"):
        try:
            value = json_data["error"]["errors"][0].get(key)
        except AttributeError as err:
            raise ValueError() from err

        if not (isinstance(value, str) or value is None):
            raise ValueError()

    return json_data
```

^ I think there would be ways to get the above Python function shorter.

### DIFF
--- a/tests/unit/via/services/google_drive_test.py
+++ b/tests/unit/via/services/google_drive_test.py
@@ -4,6 +4,7 @@ from unittest.mock import sentinel
 import importlib_resources
 import pytest
 from h_matchers import Any
+from marshmallow import ValidationError
 from pyramid.httpexceptions import HTTPNotFound
 from pytest import param
 from requests import TooManyRedirects
@@ -16,7 +17,7 @@ from via.exceptions import (
     UnhandledUpstreamException,
     UpstreamServiceError,
 )
-from via.services.google_drive import GoogleDriveAPI
+from via.services.google_drive import GoogleDriveAPI, GoogleDriveErrorSchema
 
 
 def load_fixture(filename):
@@ -25,6 +26,47 @@ def load_fixture(filename):
     )
     with ref.open(encoding="utf-8") as handle:
         return json.load(handle)
+
+
+class TestGoogleErrorBodySchema:
+    def test_it_returns_the_validated_data(self):
+        json_data = {
+            "error": {
+                "errors": [
+                    {
+                        "reason": "notFound",
+                        "message": "test_message",
+                        "other": "foo",
+                    },
+                    {"another": "error"},
+                ],
+                "other": "foo",
+            },
+            "other": "foo",
+        }
+
+        assert GoogleDriveErrorSchema().load(json_data) == json_data
+
+    @pytest.mark.parametrize(
+        "json_data",
+        [
+            # The first error in the "errors" list not a dict.
+            {"error": {"errors": [None]}},
+            # No errors in the "errors" list.
+            {"error": {"errors": []}},
+            # The "errors" list not a list.
+            {"error": {"errors": None}},
+            # The "errors" list missing.
+            {"error": {}},
+            # The top-level "error" dict not a dict.
+            {"error": 23},
+            # The top-level "error" dict missing.
+            {},
+        ],
+    )
+    def test_it_raises_ValidationError_if_the_json_data_is_invalid(self, json_data):
+        with pytest.raises(ValidationError):
+            GoogleDriveErrorSchema().load(json_data)
 
 
 class TestGoogleDriveAPI:
@@ -159,9 +201,6 @@ class TestGoogleDriveAPI:
             param({"error_class": InvalidJSONError}, id="unexpected class"),
             param({"status_code": 503}, id="unexpected status code"),
             param({"json_data": None}, id="no json"),
-            param({"json_data": {"hello": 1}}, id="unexpected json"),
-            param({"json_data": {"error": {"errors": 23}}}, id="unexpected json"),
-            param({"json_data": {"error": {"errors": [23]}}}, id="unexpected json"),
             param({"raw_data": "{... broken}"}, id="malformed json"),
         ),
     )
--- a/via/services/google_drive.py
+++ b/via/services/google_drive.py
@@ -6,6 +6,7 @@ from urllib.parse import parse_qs, urlparse
 
 from google.auth.transport.requests import AuthorizedSession
 from google.oauth2.service_account import Credentials
+from marshmallow import INCLUDE, Schema, ValidationError, fields, validate
 from pyramid.httpexceptions import HTTPNotFound
 from requests import HTTPError
 
@@ -16,34 +17,40 @@ from via.requests_tools.error_handling import iter_handle_errors
 LOG = getLogger(__name__)
 
 
-def translate_google_error(
-    error,
-):  # pylint:disable=too-many-return-statements
+class GoogleDriveErrorSchema(Schema):
+    """Schema for the JSON bodies of Google Drive error responses."""
+
+    class Meta:
+        unknown = INCLUDE
+
+    class Errors(Schema):
+        errors = fields.List(
+            fields.Dict(), required=True, validate=validate.Length(min=1)
+        )
+
+    error = fields.Nested(Errors(unknown=INCLUDE), required=True)
+
+
+def translate_google_error(error):
     """Get a specific error instance from the provided error or None."""
 
-    # This isn't a requests exception we can get meaningful data from
     if not isinstance(error, HTTPError) or error.response is None:
         return None
 
     status_code = error.response.status_code
 
-    # Try and parse out the Google details in the format we've seen
     try:
-        google_error = error.response.json()["error"]["errors"][0]
-    except (JSONDecodeError, KeyError, TypeError):
+        validated_data = GoogleDriveErrorSchema().load(error.response.json())
+    except (JSONDecodeError, ValidationError):
         return None
 
-    try:
-        google_reason, google_message = google_error.get("reason"), google_error.get(
-            "message"
-        )
-    except AttributeError:
-        return None
+    google_message = validated_data["error"]["errors"][0].get("message")
+    google_reason = validated_data["error"]["errors"][0].get("reason")
 
     # Check carefully to see that this is Google telling us the file isn't
     # found rather than this being us going to the wrong end-point
     if status_code == 404 and google_reason == "notFound":
-        return HTTPNotFound(google_message or "File id not found")
+        return HTTPNotFound(str(google_message) or "File id not found")
 
     if status_code == 403 and google_reason == "userRateLimitExceeded":
         return GoogleDriveServiceError(
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -96,6 +96,8 @@ markupsafe==1.1.1
     #   -r requirements/requirements.txt
     #   jinja2
     #   pyramid-jinja2
+marshmallow==3.13.0
+    # via -r requirements/requirements.txt
 newrelic==7.2.1.168
     # via -r requirements/requirements.txt
 oauthlib==3.1.1
--- a/requirements/functests.txt
+++ b/requirements/functests.txt
@@ -69,6 +69,8 @@ markupsafe==1.1.1
     #   -r requirements/requirements.txt
     #   jinja2
     #   pyramid-jinja2
+marshmallow==3.13.0
+    # via -r requirements/requirements.txt
 newrelic==7.2.1.168
     # via -r requirements/requirements.txt
 oauthlib==3.1.1
--- a/requirements/lint.txt
+++ b/requirements/lint.txt
@@ -108,6 +108,10 @@ markupsafe==1.1.1
     #   -r requirements/tests.txt
     #   jinja2
     #   pyramid-jinja2
+marshmallow==3.13.0
+    # via
+    #   -r requirements/requirements.txt
+    #   -r requirements/tests.txt
 mccabe==0.6.1
     # via pylint
 newrelic==7.2.1.168
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -12,3 +12,4 @@ pyramid-services
 requests
 whitenoise
 google-auth-oauthlib
+marshmallow
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -42,6 +42,8 @@ markupsafe==1.1.1
     # via
     #   jinja2
     #   pyramid-jinja2
+marshmallow==3.13.0
+    # via -r requirements/requirements.in
 newrelic==7.2.1.168
     # via -r requirements/requirements.in
 oauthlib==3.1.1
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -75,6 +75,8 @@ markupsafe==1.1.1
     #   -r requirements/requirements.txt
     #   jinja2
     #   pyramid-jinja2
+marshmallow==3.13.0
+    # via -r requirements/requirements.txt
 newrelic==7.2.1.168
     # via -r requirements/requirements.txt
 oauthlib==3.1.1


via/services/google_drive.py
```diff
@@ -16,29 +17,48 @@
...
+        )
+
+    @pre_load
+    def pre_load(self, data, **_kwargs):  # pylint:disable=no-self-use
+        try:
+            return data.response.json()
+        except JSONDecodeError as err:
+            raise ValidationError("Invalid JSON") from err
+
+    error = fields.Nested(Errors(unknown=INCLUDE), required=True)
```
jon-betts, line 31: We talked about this on slack, but I thought it might be good to capture it on the ticket.

I think it's simpler and better to keep the execution flow into the schemas as a sort of dead end, rather than through road for the data. We use Marshmallow `load()` to "transform" the data, but in truth the output is almost always the same as the input. So our transmformation is really just the identity transformation. I think we might be better off calling `validate()` instead as it advertises to the reader that the schema is side-effectless and can be ignored as long as it passes.

So:

```
data = MySchema().load(request)
... # use data
```

To:

```
data = request.json
MySchema().validate(data)
... # use data
```

via/services/google_drive.py
```diff
@@ -16,29 +17,48 @@
...
 
     status_code = error.response.status_code
 
-    # Try and parse out the Google details in the format we've seen
     try:
-        google_error = error.response.json()["error"]["errors"][0]
-    except (JSONDecodeError, KeyError, TypeError):
+        validated_data = GoogleDriveErrorSchema().load(error)
+    except ValidationError:
         return None
```
jon-betts, line 45: A companion of the above, but I think it's better to keep getting the data outside of the schema. This keeps us less tightly bound to Marshmallow. Also I don't think Marshmallow is actually is offering any additional functionality. You _can_ attach this to the schema, but you can also just not do that quite trivially, and I think it's simpler:

From:

```python
    @pre_load
    def pre_load(self, data, **_kwargs):  # pylint:disable=no-self-use
        try:
            return data.response.json()
        except JSONDecodeError as err:
            raise ValidationError("Invalid JSON") from err     
    
    # meanwhile... in the view code...
    
    try
      validated_data = GoogleDriveErrorSchema().load(error)
    except ValidationError:
        return None
```

To:

```python
try:
    data = request.json()
    GoogleDriveErrorSchema().validate(data)
except ValidationError, JSONDecodeError:
   return None
```

You could split the above up into separate try's if you preferred.

I think there are some other advantages to this:

* The bonus to readability (where does this data come from?) mentioned above
* It's actually a lot shorter
* The schema can trivially be applied to validating things which aren't from a request
* Testing becomes easier:
   * You don't have to worry `validate()`'s return value
   * You don't have to construct a request object, you can just provide a dict
   
There was some discussion of re-use around the `try ... catch JSONDecodeError` which I have a few thoughts on:

* I'm not sure something this basic is actually worth re-using. Repetition might be better
* In most views, why catch it at all? This could be dealt with by exception views
* If this is something worth re-using it doesn't necessarily imply it's anything to do with the schema. This looks like it could be a natural fit for a request method?

tests/unit/via/services/google_drive_test.py
```diff
@@ -27,6 +28,73 @@ def load_fixture(filename):
...
+            {},
+        ],
+    )
+    def test_it_raises_ValidationError_if_the_json_data_is_invalid(self, json_data):
+        error = make_requests_exception(
+            error_class=HTTPError, status_code=400, json_data=json_data
+        )
+
+        with pytest.raises(ValidationError):
+            GoogleDriveErrorSchema().load(error)
```
jon-betts: I feel like there should be a lighter touch way of testing this. Negative schema tests have some difficult properties about them, which is that the number of things which _don't_ match is effectively infinite. So you can write a real load of tests and also not really cover the space. For example the following should also fail:

```python
# Root object is not a list
[],
# Root object is not an int
123,
# Root object is not a float
123.56,
# Root object is not None
None,
# Root object is not a bool
True,
False
# Root object is not a string
"test",
...
```

I'm not sure any of those need testing, but I'm also not sure what's different between any of these cases and the ones above? Which makes me wonder how much of the above needs testing either? I think the idea is to have one negative test for each aspect of the schema and leave it at that maybe?

If this is the correct level of testing for a schema of this complexity (which is very simple) I find the amount of testing we'd need for anything more complex a bit daunting.

jon-betts (commented):
I think this is actually proving to be a good petri dish for thinking about Marshmallow usage in general, and I think there are some ideas we've generated which could be exported.

For the problem stated here, I think it's definitely worth a round of simplification in this PR to see if we can get the line count down a bit and simplify the logical flow by moving to the schema validating the data rather than acting as a pipeline for it and maybe some other ideas. Then probably take another look at how it looks.

I think we might be slightly over specifying on the contents of final dict. We specify that there are two keys with string values, but we don't require them to be present. So the only enforcement here is actually that the values are strings. I don't think this is a property we rely on in the code. We test for equality with strings. So there's potential to say "it's a dict" and leave it there.

If we came to the conclusion that that was sufficient, I think it might be worth revisiting the addition of marshmallow, as the exiting code would be functionally identical with the addition of:

```
assert isinstance(data, dict)
```

Or something.

via/services/google_drive.py
jon-betts: I tried this out, and Marshmallow's interface isn't as I expected:

* https://github.com/hypothesis/via/pull/644

The validation call returns an error object, so you have to wrap it with an if, rather than expecting it to raise.

seanh: I think there's a few suggestions here. One easy suggestion is to move the `response.json()` call and `JSONDecodeError`  catching outside of the schema. I've tried that in https://github.com/hypothesis/via/pull/643/commits/09ad5bd129f1e80e61377313a95293e8b37de247 and it seems pretty incontrovertible. It removes six lines of code from the schema and the only cost is that in the service instead of:

```python
try:
    validated_data = GoogleDriveErrorSchema().load(error)
except ValidationError:
    return None
```

You now have to do:

```python
try:
    validated_data = GoogleDriveErrorSchema().load(error.response.json())
except (JSONDecodeError, ValidationError):
    return None
```

More complicated for sure but the `response.json()` call and the `except JSONDecodeError` have just moved from one location to slightly further down in the same file and we've removed six lines of code. The schema unit tests get simpler too.

I think there is something to be said for the schema parsing the JSON **if the schema is going to be re-used** (or is a base class for multiple schema classes to inherit from) because that allows you to write the  `response.json()` call and the `except JSONDecodeError` once and then re-use that multiple times. But that's not what's happening here in Via (at least not yet anyway).

seanh: Another suggestion is to not use `GoogleDriveErrorSchema().load()`'s return value since we know it just returns the given data unmodified. This arguably makes it more explicit in the service code that we're just using `GoogleDriveErrorSchema` to validate the data and not to mutate/deserialize it. This is independent of changing from `load()` to `validate()`:

You can't do this:

```python
try:
    json_data = error.response.json()
    GoogleDriveErrorSchema().load(json_data)
except (JSONDecodeError, ValidationError):
    return None
```

because it will trigger PyLint's `try clause contains 2 statements, expected at most 1 (too-many-try-statements)`.

So instead you have to do this:

```python
try:
    json_data = error.response.json()
except JSONDecodeError:
    return None

try:
    GoogleDriveErrorSchema().load(json_data)
except ValidationError:
    return None
```

But this trigger's PyLint's `Too many return statements (7/6) (too-many-return-statements)`. It's also quite a few more lines of code than the current.

Changing from `load()` to `validate()` actually saves a couple of trivial lines of code by replacing a `try` / `except` with an `if`:

```python
try:
    json_data = error.response.json()
except JSONDecodeError:
    return None

if GoogleDriveErrorSchema().validate(json_data):
    return None
```

But it's still longer and still triggering PyLint's `too-many-return-statements`.

(https://github.com/hypothesis/via/pull/644/ also has the `too-many-return-statements`.)

tests/unit/via/services/google_drive_test.py
seanh: You just have to make a judgement about which test cases to have and which not, just pick several :man_shrugging: Try to pick a minimal set that covers all the possible bugs. I'm not sure there's any significant definable difference between the cases I chose to include and others that I didn't. I've just made a judgement about what constitutes a "good enough" set of cases to help find any bugs, within reason. I'm not sure there's any escaping from this.

> I think the idea is to have one negative test for each aspect of the schema and leave it at that maybe?

Something like that, yes. So I might test what happens if the `errors` dict is not a dict (and at random I may choose to make it a list instead). But I'm not gonna then add tests for what if it's a string, float, boolean, etc, 'cos those are effectively the same test.

seanh:
> I think we might be slightly over specifying on the contents of final dict. We specify that there are two keys with string values, but we don't require them to be present. So the only enforcement here is actually that the values are strings. I don't think this is a property we rely on in the code. We test for equality with strings. So there's potential to say "it's a dict" and leave it there.

I think this is right. We do two things with `google_message` and `google_reason`:

1. We compare `google_reason` to strings. E.g. `if ... and google_reason == "notFound"`. This doesn't actually require `google_reason` to be a string
2. We pass `google_message` as the `details` argument to `HTTPNotFound`: `return HTTPNotFound(google_message or "File id not found")`. `details` actually is supposed to be a string but we could ensure that by calling `str()` on it (and we know that any type that came from JSON is going to be nicely `str()`-able)

This allows us to remove a bunch of lines of code from the schema and from the tests as well: https://github.com/hypothesis/via/pull/643/commits/0dc48b035eb4d10e79dbd0f8617b79ecb448fb18

seanh:
> If we came to the conclusion that that was sufficient, I think it might be worth revisiting the addition of marshmallow, as the exiting code would be functionally identical with the addition of:
>
> ```python
> assert isinstance(data, dict)
> ```
>
> Or something.

I think the sweet spot may be to continue using Marshmallow but to keep the schemas simpler and not over-specify.

I think I'd vote to keep Marshmallow for a few reasons:

I think the presence of Marshmallow enables us to have a very visible "best practice" that data from requests and responses should be validated. It encourages us to actually write validation code and think about invalid data and therefore avoid bugs of the kind this code contained before. My past experience of people writing validation code in plain Python is that it quickly leads to dozens of crashes.

I think Marshmallow is also likely to lead to more consistent code and tests by providing a standard way of writing and testing validation code across apps.

Finally I think it's easier to write Marshmallow schemas that actually catch all the bugs than to write plain Python. The schema is doing a few more things for us than it may seem:

1. It ensures that `response.json()` is a dict
2. It ensures that the dict contain a top-level `"error"` key whose value is also a dict
3. It ensures that the `"error"` dict contains an `"errors"` key whose value is a list
4. It ensures that the `"errors"` list contains at least one item
5. It ensures that all the items in the `"errors"` list are dicts

So basically it's validating that the data looks like this: `{"error": {"errors": [{ <ANY_DICT_CONTENTS> }]}`.

I think the plain Python would be something like:

```python
def validate_google_drive_error_schema(data):
    try:
        error = data["error"]["errors"][0]
    except (KeyError, TypeError, IndexError) as err:
        raise ValueError() from err

    if not isinstance(error, dict):
        raise ValueError()

    return data
```

That's 10 lines of Python. The Marshmallow schema to achieve the same is half the size:

```python
class GoogleDriveErrorSchema(MyBaseSchema):
    class Errors(Schema):
        errors = fields.List(fields.Dict(), required=True, validate=Length(min=1))

    error = fields.Nested(Errors(unknown=INCLUDE), required=True)
```

I've cheated slightly by removing the `class Meta` from the Marshmallow schema and assuming that would be in a custom schema base class that we inherit from. That's what we do in LMS and I think you'd do that in Via too if you had more than one schema.

But I think the fewer lines of code doesn't quite get to the real point, which is that Marshmallow schemas are easier to write correctly and more likely to catch all the bugs. If writing plain Python I think you're more likely to fail to catch all the exceptions or do all the `isinstance` checks, as we saw in the initial plain Python version of this code.

via/services/google_drive.py
seanh: This is covered in https://github.com/hypothesis/via/pull/643#discussion_r726058550 below. I tried it but I ended up with more lines of code and triggering PyLint warnings to boot :man_shrugging: You'd need to do more work on `translate_google_error()` to enable it. That function is too long and complex. Maybe something for a separate PR

seanh:
To sum up the changes in this PR I think we've gone from:

```python
try:
    google_error = error.response.json()["error"]["errors"][0]
except (JSONDecodeError, KeyError, TypeError):
    return None

try:
    google_reason, google_message = google_error.get("reason"), google_error.get(
        "message"
    )
except AttributeError:
    return None
```

(which is still not actually quite right: I think it also needs to catch `IndexError` as well as `JSONDecodeError`, `KeyError` and `TypeError`)

To:

```python
class GoogleDriveErrorSchema(Schema):
    class Errors(Schema):
        errors = fields.List(fields.Dict(), required=True, validate=Length(min=1))

    error = fields.Nested(Errors(), required=True)

try:
    validated_data = GoogleDriveErrorSchema().load(error.response.json())
except (JSONDecodeError, ValidationError):
    return None

google_message = validated_data["error"]["errors"][0].get("message")
google_reason = validated_data["error"]["errors"][0].get("reason")
```

(Omitting the `class Meta` noise again).

We've also fixed at least one bug (it no longer crashes on an empty `"errors"` list) and made the validation tests more thorough. And we managed to remove a `# pylint:disable=too-many-return-statements`.

The Marshmallow version is no shorter but I think it's easier to write correctly. We've taken a few passes at the pre-Marshmallow version of this code already and there are still bugs in it.

via/services/google_drive.py
```diff
@@ -16,34 +17,40 @@
...
-    except AttributeError:
-        return None
+    google_message = validated_data["error"]["errors"][0].get("message")
+    google_reason = validated_data["error"]["errors"][0].get("reason")
 
     # Check carefully to see that this is Google telling us the file isn't
     # found rather than this being us going to the wrong end-point
     if status_code == 404 and google_reason == "notFound":
-        return HTTPNotFound(google_message or "File id not found")
+        return HTTPNotFound(str(google_message) or "File id not found")
```
jon-betts, line 53: ~This shouldn't be necessary by the schema? It's either a string or missing?~

I see this has been dropped from the schema

jon-betts (commented):
If the time for discussion has passed let me know, but I think there are some claims about Marshmallow made here which are potentially applicable to the wider case, or aren't unique to Marshmallow schema / only really land in more complex use cases. So this comment is a discussion about this piece of code, not a higher principal or the usage of Marshmallow in general. I don't believe what we decide is right for this code has to imply to anything else.

There's an example at the bottom of the ticket description which is pitched as a pure Pythong alternative to the Marshmallow schema. I assume it was written before the JSON deserialisation was removed from the schema (otherwise it's double counting a bit), but I think it's kind of written in a way to show it's all the same rather than how you might actually write it.

One of the questions you asked me recently to start asking myself is: "What is the simplest thing that could work here?". For example I think the following code both passes all the tests _and_ doesn't reject valid cases like the existing code would (like `[{}, 1]`):

```python
    def is_valid(self, data):
        try:
            return isinstance(data['error']['errors'][0], dict)
        except (TypeError, KeyError, IndexError, AssertionError):
            return False
```

To paraphrase the advantages at the top of the file, introducing a function like this would:

- Separate validating the JSON body from taking different actions based on the contents of the valid JSON body. A single more complex function is split into two simpler ones.
- Could have its own direct unit tests. This just allows us to be more thorough about testing the validation and makes bugs less likely.
- The translate_google_error() function can now just call the function and then go ahead and assume that the data is as expected without worrying about crashing

### So?

* Many of the points raised about using a schema library are good points in the general case, especially when the data is more complicated
* I think the value in this PR is the extra testing
* It's good and catches many potential error cases
* Splitting this functionality out from the main function has some advantages
* I think there's are simpler solutions to getting values from the nested data structure we have than a schema

### Is this even a schema enforcement task?

I think we have a number of distinct problem types when it comes to injesting data from unknown sources:

 * Sometimes we really want to enforce a particular shape (when reaching our APIs for example) and a schema is perfect / the only sane way to do it
* Sometimes a schema is a means to an end, and we just want to make sure our code won't fall over later when we try and use the data structure
* One specific case of that (this one) is trying to pluck a few values out of a larger structure without crashing

In this last case _we really don't care_ about the structure of the data save the fact we can get what we need out of it. You can see that reflected in our schema. We have to go to some  lengths to get it to _not_ apply rules we don't care about. 
So if we view this as a problem of data extraction rather than schema enforcement, then there might be other solutions available.

In this case something like JSON path might be more appropriate. For example there are python bindings to `jq`:

```python
reason = jq.first(".error.errors[0].reason", json_data)
message = jq.first(".error.errors[0].message", json_data)
```

_(This looks super neat, but is a bit of a cheat, as it can still raise `ValueError` in some cases, but you could deal with that with a very small wrapper)_

I picked `jq` at random, but proper JSON path actually has good support for recursive operations, which means we could be _very_ tolerant of inputs if we wanted to. For example `$...reason` will find any dict anywhere in the data structure which includes the key `reason`. This means we'd be very tolerant to the input structure.

seanh:
> One of the questions you asked me recently to start asking myself is: "What is the simplest thing that could work here?". For example I think the following code both passes all the tests _and_ doesn't reject valid cases like the existing code would (like `[{}, 1]`):
> 
> ```python
>     def is_valid(self, data):
>         try:
>             return isinstance(data['error']['errors'][0], dict)
>         except (TypeError, KeyError, IndexError, AssertionError):
>             return False
> ```
> 
> To paraphrase the advantages at the top of the file, introducing a function like this would:
> 
> * Separate validating the JSON body from taking different actions based on the contents of the valid JSON body. A single more complex function is split into two simpler ones.
> * Could have its own direct unit tests. This just allows us to be more thorough about testing the validation and makes bugs less likely.
> * The `translate_google_error()` function can now just call the function and then go ahead and assume that the data is as expected without worrying about crashing

I think this is true - that actually does pass all the same tests and has the desired properties. And putting the `data['error']['errors'][0]` inside the `isinstance()` is a neat trick for getting it all into a single `try`. The only thing that's giving me pause about this is that, while the end result is very small and simple, I think it might actually be quite hard to write things like this without making mistakes:

```python
    return isinstance(data['error']['errors'][0], dict)
except (TypeError, KeyError, IndexError, AssertionError):
```

It's hard not to forget to catch one of the exception classes or forget the `isinstance`.

Having said that if we were good about putting this kind of code into separate and directly unit tested validation functions then unit test cases may rescue us from not getting the code quite right.

Changing `is_valid()` into a boolean function (rather than one that raises `ValidationError`) causes the `too-many-return-statements` warning to pop up again for `translate_google_error()`.

> # Is this even a schema enforcement task?
> 
> ...I think we have a number of distinct problem types when it comes to injesting data from unknown sources:
> 
> Sometimes a schema is a means to an end, and we just want to make sure our code won't fall over later when we try and use the data structure

I think this is true. What `translate_google_error()` really wants is just a method for getting a `(message, reason)` tuple from `error.response` (where `message` and/or `reason` can potentially be `None`). I think it's actually deserialization not validation:

```python
def load_google_drive_error(error):
    try:
        error_dict = error.response.json()["error"]["errors"][0]
    except (JSONDecodeError, TypeError, KeyError, IndexError):
        return (None, None)

    try:
        return (error_dict.get("message"), error_dict.get("reason"))
    except AttributeError:
        return (None, None)

def translate_google_error(error):
    ...

    google_message, google_reason = load_google_drive_error(error)

    ...
```

This also has the benefit of keeping knowledge of the shape of Google Drive responses in one place in `load_google_drive_error()` rather than having it split between the schema or `is_valid()` function and the `translate_google_error()` function.

I'm not sure what to call that function though or how best to write the code.

> I picked `jq` at random, but proper JSON path actually has good support for recursive operations, which means we could be very tolerant of inputs if we wanted to. For example `$...reason` will find any dict anywhere in the data structure which includes the key `reason`. This means we'd be very tolerant to the input structure.

That sounds possibly a bit _too_ tolerant

seanh:
Squashed and rebased...

seanh:
I think I'm just gonna stick with the Marshmallow schema in order to draw a line under this.

The difficulties I'm having with this plain Python version:

```python
def is_valid(self, data):
    try:
        return isinstance(data['error']['errors'][0], dict)
    except (TypeError, KeyError, IndexError, AssertionError):
        return False
```

are:

1. While it's fewer lines of code and avoids the Marshmallow dependency I think this kind of Python code is quite hard to write: having to remember to catch every different type of exception and also the `isinstance()`. I think it may be error-prone as an approach.

2. Because it's a boolean function rather than one that raises `ValidationError` it ends up having to add more lines of code in `translate_google_error()`. Instead of:

   ```python
   try:
       validated_data = GoogleDriveErrorSchema().load(error.response.json())
   except (JSONDecodeError, ValidationError):
       return None
   ```

   You end up having to do:

   ```python
   try:
       data = error.response.json()
   except JSONDecodeError:
      return None

   if not is_valid(data):
      return None
   ```

   So it doesn't end up being an overall saving in lines of code compared to the Marshmallow version.

3. You also end up triggering a `too-many-return-statements` PyLint warning in `translate_google_error()`. This is more `translate_google_error()`'s fault but nonetheless I don't want to refactor that function right now.

You can try to rewrite the `is_valid()` function to have the same kind of interface as Marshmallow so that `translate_google_error()` can use it in the same way. `is_valid()` would have to raise on invalid, and it would have to return the validated data. But this starts to result in more awkward and longer code, it ends up becoming longer than the Marshmallow schema and I'd say even more error-prone to write or to review:

```python
def validate_google_drive_error(self, data):
    try:
        error_dict = data['error']['errors'][0]
    except (TypeError, KeyError, IndexError) as err:
        raise ValueError() from err

    if not isinstance(error_dict, dict):
        raise ValueError()

    return data

def translate_google_error(error):
    ...

    try:
        validated_data = validate_google_drive_error(error.response.json())
    except (JSONDecodeError, ValueError):
        return None

    google_message = validated_data["error"]["errors"][0].get("message")
    google_reason = validated_data["error"]["errors"][0].get("reason")

    ...
```

I do appreciate the simplicity of doing it in plain Python. And perhaps with enough effort a concise plain Python version that avoids the above problems is possible. But I feel we may be trying too hard to avoid a Marshmallow schema that's actually pretty simple and does the job.

I've got a feeling that Marshmallow does quite a lot for free and that writing Marshmallow schemas might be a more foolproof way of doing this sort of thing (especially given appropriate helpers like a custom base class to inherit from). It requires you to know Marshmallow but it's actually simpler code that's harder to mess up:

```python
class GoogleDriveErrorSchema(Schema):
    class Errors(Schema):
        errors = List(Dict(), required=True, validate=Length(min=1))

    error = Nested(Errors(), required=True)
```

In any case, for time and practicality's sake I'm gonna leave the Marshmallow schema in for now.

jon-betts (approved):

PR #648, opened 2021-10-18
Proxy MS OneDrive PDFs on python

marcospri:
Part of: https://github.com/hypothesis/lms/issues/3256

# Testing notes

URLs: 
- one drive `https://api.onedrive.com/v1.0/shares/u!aHR0cHM6Ly8xZHJ2Lm1zL2IvcyFBaU9PUWVsRWNZS3diZnVvb3JRdmNYYnR2cjQ/root/content`

- sharepoint `https://listhypothesis-my.sharepoint.com/:b:/g/personal/eng_list_hypothes_is/EQS1fwBTvJRJkgiKu_qz-oIBce4F4t2zIWdSVWSMF2MEOg?e=TLbqHg&download=1`

Try to proxy both on http://localhost:9083/

On `master` one drive url gets proxied, sharepoint one fails. On this branch both are successful.


via/services/__init__.py
```diff
@@ -22,39 +22,3 @@ def includeme(config):  # pragma: no cover
     config.register_service_factory(
         "via.services.secure_link.factory", iface=SecureLinkService
     )
-
-
-def create_google_api(settings):
```
marcospri, line 27: Replaced with `google_drive.factory`.

via/services/__init__.py
```diff
@@ -22,39 +22,3 @@ def includeme(config):  # pragma: no cover
...
-    """Create from Pyramid settings."""
-
-    return GoogleDriveAPI(
-        credentials_list=load_injected_json(settings, "google_drive_credentials.json"),
-        resource_keys=load_injected_json(settings, "google_drive_resource_keys.json"),
-    )
-
-
-def load_injected_json(settings, file_name):
```
marcospri, line 36: Moved to `google_drive.py` and made private there. If this needs to be public I think it belongs to app/config or similar.

via/services/google_drive.py
```diff
@@ -183,17 +179,55 @@ def iter_file(self, file_id, resource_key=None) -> Iterator[ByteString]:
                     resource_key,
                 )
 
+        headers = {}
         if resource_key:
             headers["X-Goog-Drive-Resource-Keys"] = f"{file_id}/{resource_key}"
 
         response = self._session.get(
```
marcospri, line 186: I tried to use the new service's iter_url here but the google part:
- Uses a session which looks like a request.session but is not. Taking a session on the new service works but I think it's confusing.
- Uses `@iter_handle_errors(translate_google_error)` for error handling. Using something similar in the new service will shallow the exceptions so the translation here wound't work. I could refactor the error handling on both to make it work but I think what I tried was too complicated at the end.

via/services/proxy_pdf.py
```diff
@@ -0,0 +1,79 @@
...
+                "Accept-Encoding": "gzip, deflate",
+                "User-Agent": "(gzip)",
+            }
+        )
+        if headers:
+            request_headers.update(headers)
+        return request_headers
+
+    @staticmethod
+    def reify_first(iterable):
```
marcospri, line 66: Is this the right place for `reify_first`? 

I think you could argue it belongs to some generic `utils` package but I think this is as good as `via/views/google_drive.py`

tests/unit/via/services/google_drive_test.py
```diff
@@ -69,6 +74,69 @@ def test_it_raises_ValidationError_if_the_json_data_is_invalid(self, json_data):
             GoogleDriveErrorSchema().load(json_data)
 
 
+class TestLoadInjectedJSON:
```
marcospri, line 77: Moved from `tests/unit/via/services/__init___test.py`

tests/unit/via/services/google_drive_test.py
```diff
@@ -69,6 +74,69 @@ def test_it_raises_ValidationError_if_the_json_data_is_invalid(self, json_data):
...
+    def test_it_with_missing_data(self, settings):
+        with pytest.raises(ConfigurationError):
+            _load_injected_json(settings, "missing.json")
+
+    @pytest.fixture
+    def settings(self, tmpdir):
+        return {"data_directory": tmpdir}
+
+
+class TestFactory:
```
marcospri, line 102: Moved from tests/unit/via/services/__init___test.py

via/services/ms_one_drive.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class MSOneDriveService:
+    URL_REGEXPS = (
+        re.compile(r"^https://.*\.sharepoint.com/.*download=1"),
+        re.compile(r"^https://api.onedrive.com/v1.0/.*/root/content"),
+    )
+
+    @classmethod
+    def is_one_drive_url(cls, url):
```
marcospri, line 11: This doesn't really need to be a service + class + method but for consistency when for the service.

via/views/view_pdf.py
```diff
@@ -51,6 +55,28 @@ def view_pdf(context, request):
     }
 
 
+@view.view_config(
+    route_name="python_proxy_pdf",
+    decorator=(has_secure_url_token,),
+)
+def python_proxy_pdf(context, request):
```
marcospri: This is `python_proxy_pdf`, it doesn't do anything related to OneDrive.

If we eventually have to write some OneDrive specific view it can go to its own module.

If we ever need to proxy any other PDFs, eg all pdfs coming from "example.com" we can just match the URL and take it to this view.

PR #654, opened 2021-10-21
Add HTTPService

marcospri:
Just the HTTPService part of https://github.com/hypothesis/via/pull/656/files on it's own PR

# Testing 

On http://localhost:9083/ and with caching disabled

- Proxy a Google pdfs https://drive.google.com/file/d/1VSp920krfYJ17txtJb-WOKeynNC_t3jE/view?usp=sharing
- Same but alter the code to produce a 400
```
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: via/services/google_drive.py
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ via/services/google_drive.py:192 @ class GoogleDriveAPI:

        yield from self._http_service.stream(
            url=url,
            url=url + "SOMETHING ELSE",
            headers=headers,
            timeout=self.TIMEOUT,
            max_allowed_time=self.TIMEOUT,
```
should produce a  `via.exceptions.UnhandledUpstreamException`

- Compare the headers from `master` on the proxied PDF
```
HTTP/1.1 200 OK
Server: nginx/1.17.6
Date: Tue, 26 Oct 2021 15:36:22 GMT
Content-Type: application/pdf
Transfer-Encoding: chunked
Connection: keep-alive
Content-Disposition: inline
Cache-Control: public, max-age=43200, stale-while-revalidate=86400
X-Via: compute
```

- Proxy/launch an assignment with `resource_key` https://hypothesis.instructure.com/courses/125/assignments/1912

- Proxy a big PDF in Google drive (2GB) https://drive.google.com/file/d/1nv0L6oHKUqI2-shVQGhdros1uv2eeR4R/view?usp=sharing. Memory usage shouldn't suddenly  increase to 2GB (CPU  will get hot though)

- Proxy a non-google pdf https://h.readthedocs.io/_/downloads/en/latest/pdf/


via/services/google_drive.py
```diff
@@ -196,4 +197,4 @@ def iter_file(self, file_id, resource_key=None) -> Iterator[ByteString]:
 
         response.raise_for_status()
 
-        yield from stream_bytes(response)
+        yield from HTTPService.stream_bytes(response)
```
marcospri: I think if I'm not going to completely switch GoogleService to HTTPService using the functions this way it's cleaner than instantiating a service.

via/services/http.py
```diff
@@ -0,0 +1,134 @@
+from itertools import chain
+
+import requests
+from requests import RequestException
+
+from via.exceptions import UnhandledUpstreamException
+from via.requests_tools.error_handling import DEFAULT_ERROR_MAP
+
+
+class HTTPService:
```
marcospri, line 24: This is mostly copied from LMS. I don't think it would make sense to make a lib out of it.

via/services/http.py
```diff
@@ -0,0 +1,134 @@
+from itertools import chain
+
+import requests
+from requests import RequestException
+
+from via.exceptions import UnhandledUpstreamException
+from via.requests_tools.error_handling import DEFAULT_ERROR_MAP
```
marcospri: I think a dict is the best solution for what we are trying to do here. Importing the existing one.

via/services/http.py
```diff
@@ -0,0 +1,134 @@
...
+                url,
+                timeout=timeout,
+                **kwargs,
+            )
+            response.raise_for_status()
+        except RequestException as err:
+            TargetClass = DEFAULT_ERROR_MAP.get(  # pylint:disable=invalid-name
+                err.__class__, UnhandledUpstreamException
+            )
+            raise TargetClass(
```
marcospri: I think `target_class(...)` is worse that the pylint:disable

via/services/http.py
```diff
@@ -0,0 +1,134 @@
...
+            response = self._session.request(
+                method,
+                url,
+                timeout=timeout,
+                **kwargs,
+            )
+            response.raise_for_status()
+        except RequestException as err:
+            TargetClass = DEFAULT_ERROR_MAP.get(  # pylint:disable=invalid-name
+                err.__class__, UnhandledUpstreamException
```
marcospri: Default to `UnhandledUpstreamException`.

Using a dict seems right here, eventually that dict could be local in this file but as it's shared with the google part I didn't want to just copy paste it.

via/services/http.py
```diff
@@ -0,0 +1,134 @@
...
+                message=err.args[0] if err.args else None,
+                requests_err=err,
+            ) from err
+
+        return response
+
+    def stream(self, url, method="GET", **kwargs):
+        response = self.request(method=method, url=url, stream=True, **kwargs)
+
+        yield from self.reify_first(self.stream_bytes(response))
```
marcospri: I don't think we would want to stream something without doing these two transformations so doing them here directly.

via/services/http.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @staticmethod
+    def reify_first(iterable):
```
marcospri: These two are `staticmethod` as they are used on the google side. I don't think that's wrong, if the google side started using `.stream()` I would only change them to be private.

via/services/proxy_pdf.py
```diff
@@ -0,0 +1,55 @@
+from via.requests_tools import add_request_headers
```
marcospri: A toolbox approach for this it's not that bad, they are request headers we want to be used everywhere, I think the alternative would be a `Proxy` service with just that but just a functions seems better.

via/services/proxy_pdf.py
```diff
@@ -0,0 +1,55 @@
...
+    TIMEOUT = (30, 30)
+
+    def __init__(self, http_service):
+        self.http_service = http_service
+
+    def proxy(self, url, headers=None, timeout=TIMEOUT):
+        yield from self.http_service.stream(url, headers=headers, timeout=timeout)
+
+    @staticmethod
+    def response_headers(headers=None):
```
marcospri: Again, these two are not self-methods as they will be used on the google side.

via/get_url.py
```diff
@@ -29,12 +26,10 @@ def get_url_details(url, headers=None):
 
     headers = add_request_headers(clean_headers(headers))
 
-    with requests.get(
+    with http_service.stream(
```
marcospri: ~Switch this over to `.stream()`~

I left it as get as it both:
- Uses the context manager syntax with the response which won't work with the iterator coming from `.stream()`
- Doesn't consume the request's body which would be the ideal way to consume the iterator.

This points to make it part of the service itself might be the best answer instead of dealing with this requests details elsewhere.

I'm not 100% sure of the effect of `allow_redirects` on a GET request so I left it here for now instead of making it the default.

via/services/google_drive.py
```diff
@@ -115,7 +115,10 @@ def __init__(self, credentials_list, resource_keys):
                 "The Google Drive service account information is invalid"
             ) from exc
 
-        self._session = AuthorizedSession(credentials, refresh_timeout=self.TIMEOUT)
+        self._http_service = HTTPService(
```
marcospri, line 118: `http_service = request.find_service(name"http").configure(session=..., ...)
`

Something like that could be useful if the interface becomes more complex but I think now they are pretty equivalent, if the API of `configure` vs the API of `__init__` change, will have to most likely go change every occurrence in a very similar manner.

via/services/google_drive.py
```diff
@@ -186,14 +188,9 @@ def iter_file(self, file_id, resource_key=None) -> Iterator[ByteString]:
         if resource_key:
             headers["X-Goog-Drive-Resource-Keys"] = f"{file_id}/{resource_key}"
 
-        response = self._session.get(
+        yield from self._http_service.stream(
             url=url,
             headers=headers,
-            stream=True,
             timeout=self.TIMEOUT,
             max_allowed_time=self.TIMEOUT,
```
marcospri, line 195: Note that max_allowed_time is not a request argument, it only works on Google's Authorized session so I think it's better to be explicit about both values here together.

via/services/http.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def __init__(self, session=None, error_translator=None):
+        # A requests session is used so that cookies are persisted across
+        # requests and urllib3 connection pooling is used (which means that
+        # underlying TCP connections are re-used when making multiple requests
+        # to the same host, e.g. pagination).
+        #
+        # See https://docs.python-requests.org/en/latest/user/advanced/#session-objects
+        self._session = session or requests.Session()
+        self._error_translator = error_translator
```
marcospri, line 35: Renamed to translator vs mapper.

via/services/http.py
```diff
@@ -0,0 +1,138 @@
...
+        try:  # pylint:disable=too-many-try-statements
+            response = self._session.request(
+                method,
+                url,
+                timeout=timeout,
+                **kwargs,
+            )
+            response.raise_for_status()
+        except Exception as err:
+            if self._error_translator and (mapped := self._error_translator(err)):
```
marcospri: @seanh I think this will most likely fit within exceptions views and bit of code (or maybe this same code moved over) but I'd leave it for a feature PR.

via/services/http.py
```diff
@@ -0,0 +1,138 @@
...
+
+        return response
+
+    def stream(self, url, method="GET", **kwargs):
+        response = self.request(method=method, url=url, stream=True, **kwargs)
+
+        yield from self._stream_bytes(response)
+
+    @staticmethod
+    def _stream_bytes(response, min_chunk_size=64000):
```
marcospri, line 121: Kept it as a private method to not pollute the API but also kept the existing test which test this behavior nicely.

via/get_url.py
```diff
@@ -2,17 +2,14 @@
...
-import requests
-
-from via.requests_tools.error_handling import handle_errors
 from via.requests_tools.headers import add_request_headers, clean_headers
 from via.services.google_drive import GoogleDriveAPI
 
 
-@handle_errors
-def get_url_details(url, headers=None):
+def get_url_details(http_service, url, headers=None):
```
marcospri, line 9: Yes, this whole file make sense as part of HTTP service, `via/get_url.py` is a bit in no man's land.

via/services/http.py
```diff
@@ -0,0 +1,139 @@
...
+
+        try:  # pylint:disable=too-many-try-statements
+            response = self._session.request(
+                method,
+                url,
+                timeout=timeout,
+                **kwargs,
+            )
+            response.raise_for_status()
+        except Exception as err:
```
marcospri, line 89: `_error_translator` Might be interested on some error not related to RequestException so we have to broad here.

tests/.pylintrc
```diff
@@ -22,6 +22,7 @@ disable=bad-continuation,
         no-value-for-parameter,
         redefined-outer-name,
         too-few-public-methods,
+        too-many-public-methods,
```
seanh, line 25: :+1: Yeah, I don't think we really want this warning for the tests. I can certainly see an argument for disabling it anyway

seanh (approved):

tests/.pylintrc
marcospri: Forgot to comment on that line, yeah, that's disabled on the test code in other apps (eg LMS)

via/services/http.py
```diff
@@ -0,0 +1,139 @@
...
+                message=err.args[0] if err.args else None,
+                requests_err=err if hasattr(err, "request") else None,
+            ) from err
+
+        return response
+
+    def stream(self, url, method="GET", **kwargs):
+        response = self.request(method=method, url=url, stream=True, **kwargs)
+
+        yield from self._stream_bytes(response)
```
jon-betts: I don't think this is quite right here. As we return a generator, no exceptions are raised until we start evaluating it. So many potential errors during connection and all errors during iteration will be missed.

I think you need to wrap the error handling around the whole call in `stream`.

You can see in the previous version that it's important that some form of evaluation happens inside the error handler. You don't actually need to reify it there, but that's what the `yield from` was doing:

Not sure I'm explaining this very well. A pseudocode version:

```
# Current situation - broken I think
with error handling:
    do request
yield from request

# Also broken I think
with error handling:
    do request
    return request
    
# This works - as long as it's externally reified before the Pyramid starts the response
with error handling:
    do request
    yield from request
```

PR #655, opened 2021-10-25
An example of the outcome of some rounds of refactoring

jon-betts:
This is the result of a demo run to show some refactoring for practice. This is based on the code from:

 * https://github.com/hypothesis/via/pull/648

### The big changes from the parent PR

 * Consolidate PDF views into one file
 * Identifying common code in these views and inlining it as a result
 * Extract a PDF URL service from the views (good name TBC)
 * Inline the `MSOneDrive` service into that service
 * Big reduction of the responsibilities of the existing `PDFService` which disentangles if from many places

### Results

 * I think this results in smaller and more obvious view code
 * This drops the code size a fair bit
 * I suspect that will really kick in once the tests are rewritten to take advantage of this (common bits for the views and interface boundaries should help)

### Food for thought

 * Do we have a OneDrive service or a generic service?
 * Might it be nice if all the PDF services were consolidated into a directory? 
 * If that happened are some private to a public interface?
 * This might make the contents more mobile and refactorable, but it might just add unnecessary layers


jon-betts:
I think this has served it's purpose. It was just for demonstration.

PR #656, opened 2021-10-25
Include a HTTPService that wraps requets

marcospri:


via/get_url.py
```diff
@@ -29,7 +28,7 @@ def get_url_details(url, headers=None):
 
     headers = add_request_headers(clean_headers(headers))
 
-    with requests.get(
+    with http_service.get(
```
seanh, line 31: :+1: Nice to see `HTTPService` being used here as well.

I think you could now omit the `timeout` argument? As `get_url_details()` can just go with `HTTPService`'s default (and more correct) `timeout`.

I also wonder if the `allow_redirects=True` could be moved into `HTTPService` as a default?

No need to address this now but I noticed that we're also passing `stream=True` here which makes me wonder whether this should be using `http_service.stream()`?

via/services/google_drive.py
```diff
@@ -117,7 +117,10 @@ def __init__(self, credentials_list, resource_keys):
                 "The Google Drive service account information is invalid"
             ) from exc
 
-        self._session = AuthorizedSession(credentials, refresh_timeout=self.TIMEOUT)
+        self._http_service = HTTPService(
+            session=AuthorizedSession(credentials, refresh_timeout=self.TIMEOUT),
+            error_mapping=translate_google_error,
+        )
```
seanh, line 123: :+1: I think this is fair enough for now.

I'm not sure what the ideal solution is more generally for when we have a service (like `HTTPService`) but we want to construct different instances of it in different places. Service factories have the advantage of hiding how a service is constructed from the users of that service, which has all sorts of benefits (service users don't need to martial loads of arguments to construct a service; if you change how a service is constructed you only need to change the service factory; getting mock services in tests is easier). These benefits are bigger when a service is complex to construct because it needs a lot of arguments. In some cases I think this pattern might be better:

```python
http_service = request.find_service(name"http").configure(session=..., ...)
```

In other words `HTTPService` would have a `configure()` method that returns another instance of itself.

In this case though I think just calling `HTTPService()` directly is fine :+1:

via/services/google_drive.py
```diff
@@ -177,25 +179,22 @@ def iter_file(self, file_id, resource_key=None) -> Iterator[ByteString]:
...
+                "Accept": "*/*",
+                "Accept-Encoding": "gzip, deflate",
+                "User-Agent": "(gzip)",
+            }
+        )
         if resource_key:
             headers["X-Goog-Drive-Resource-Keys"] = f"{file_id}/{resource_key}"
 
-        response = self._session.get(
+        yield from self._http_service.stream(
```
seanh, line 192: I wonder if the `timeout` and `max_allowed_time` arguments here should be defaults in `HTTPService.stream()`? (And streaming should use a different default timeout than `get()` etc).

Also wonder whether `get()` and the others should have a `max_allowed_time` as well as a `timeout`?

via/services/http.py
```diff
@@ -0,0 +1,131 @@
...
+            response = self._session.request(
+                method,
+                url,
+                timeout=timeout,
+                **kwargs,
+            )
+            response.raise_for_status()
+        except RequestException as err:
+            if self._error_mapping and (mapped := self._error_mapping(err)):
+                raise mapped from err
```
seanh, line 87: I think we might want to rename `self._error_mapping` to something else. "Mapping" made me think it was a dict, but it's actually a _callable_ that takes an exception object and returns another exception object or `None`. Maybe `self._translate_error`? Or `_error_translator`?

I bet we could get rid of this by using exception views though. An exception view's job is to catch exceptions and turn them into different error responses based on the type and values of the exceptions. And you'd just register different exception views for different routes (like the Google route). Not something I'm asking you to do now though, to be clear

via/services/http.py
```diff
@@ -0,0 +1,131 @@
...
+            translated_class = UnhandledUpstreamException
+            for error_class, target_class in DEFAULT_ERROR_MAP.items():
+                if isinstance(err, error_class):
+                    translated_class = target_class
+                    break
+
+            raise translated_class(
+                message=err.args[0] if err.args else None,
+                requests_err=err if hasattr(err, "request") else None,
+            ) from err
```
seanh, line 98: It's not for this PR but perhaps for one day in the future: I feel there must be a simpler and more obvious way to handle this, without all this code about "error mappings" and "translated classes". I suspect this might be code that belongs in the exception views rather than in the service (which I think might also remove the need to configure the service with different error mappings)

seanh: Already said in another comment, sorry

via/services/http.py
```diff
@@ -0,0 +1,131 @@
...
+
+        return response
+
+    def stream(self, url, method="GET", **kwargs):
+        response = self.request(method=method, url=url, stream=True, **kwargs)
+
+        yield from self.stream_bytes(response)
+
+    @staticmethod
+    def stream_bytes(response, min_chunk_size=64000):
```
seanh, line 108: I think this could probably either be a private method or be inlined into `stream()` above, as it's only called by `HTTPService.stream()`. Hopefully it'd be possible to unit test `stream_bytes()` by calling `stream()` (if `stream_bytes()` had either been inlined into `stream()` or had been made private)?

via/views/proxy.py
```diff
@@ -1,15 +1,15 @@
...
+from via.services import HTTPService, ViaClientService
 
 
 @view_config(route_name="proxy", renderer="via:templates/proxy.html.jinja2")
 def proxy(context, request):
     url = context.url_from_path()
     request.checkmate.raise_if_blocked(url)
 
-    mime_type, _status_code = get_url_details(url)
+    mime_type, _status_code = get_url_details(request.find_service(HTTPService), url)
```
seanh, line 12: :+1: I think this is fine for now but could be refactored in future. It looks like `get_url_details()` probably belongs in `HTTPService` so it'd be: `mime_type, _status_code = request.find_service(HTTPService).get_details(url)`. That'd also get rid of the weird `via/get_url.py` module.

via/views/view_pdf.py
```diff
@@ -66,7 +65,13 @@ def proxy_google_drive_file(request):
 
 
 def _iter_pdf_response(response, content_iterable):
+
     try:
+        # Get an iterable where the first item only has already been reified.
+        # Return None if there is no content to return.
+        # This takes a potentially lazy generator, and ensures the first item is
+        # called now. This is so any errors or problems that come from starting the
+        # process happen immediately, rather than whenever the iterable is evaluated.
```
seanh, line 74: I wonder whether `iter_pdf_response()` belongs in `HTTPService.stream()`? It looks like every time you call `HTTPService.stream()` you then have to call `_iter_pdf_response()` on the result.

I think we might not want `HTTPService` to have to have the Pyramid response object or to be coupled to Pyramid. But at least the `content_iterable = chain((next(content_iterable),), content_iterable)` part of `_iter_pdf_response()` looks like it may belong in `HTTPService.stream()`. So then the view code could be something more like:

```python
url = ...
response.app_iter = http_service.stream(url, ...)
response.headers.update({ <PDF streaming headers> })
return response
```

seanh: Again not for this PR but the PDF streaming headers could be moved into a Pyramid renderer too. Then you wouldn't need some sort of `_pdf_streaming_headers()` helper that every PDF proxying view needs to import and call. Instead a view could just do:

```python
@view_config(..., renderer="proxy_pdf")
def proxy_ms_onedrive_pdf(request):
    url = ...
    request.response.app_iter = http_service.stream(url, ...)
```

via/services/google_drive.py
marcospri: Note that `max_allowed_time` is not a request argument, it only works on Google's Authorized session so I think it's better to be explicit about both values here together.

via/views/proxy.py
marcospri: Yeah, getting this into HTTPService itself would be nicer :+1:

PR #657, opened 2021-10-26
New PDFURLBuilder service

marcospri:
Same testing as for https://github.com/hypothesis/via/pull/654, nothing should have changed from the proxing PoV.


via/views/view_pdf.py
```diff
@@ -23,81 +23,61 @@ def view_pdf(context, request):
...
+
+def _iter_pdf_response(response, content_iterable):
+
+    try:
+        # Get an iterable where the first item only has already been reified.
+        # Return None if there is no content to return.
+        # This takes a potentially lazy generator, and ensures the first item is
+        # called now. This is so any errors or problems that come from starting the
+        # process happen immediately, rather than whenever the iterable is evaluated.
+        content_iterable = chain((next(content_iterable),), content_iterable)
```
marcospri, line 63: > @seanh   wonder whether iter_pdf_response() belongs in HTTPService.stream()? It looks like every time you call HTTPService.stream() you then have to call _iter_pdf_response() on the result.

Not 100% sure, moving it to `stream()` is trivial but as  this is dealing with iterators I think "reifying" them should happen as late as possible, here we do that just before sending back to the client.

tests/unit/via/services/pdf_url_test.py
```diff
@@ -0,0 +1,136 @@
...
+            return_value=datetime(
+                year=2020,
+                month=2,
+                day=8,
+                hour=17,
+                minute=30,
+                second=21,
+                tzinfo=timezone.utc,
+            ),
+        )
```
jon-betts, line 46: This is code as you found it, so keeping it is absolutely fine. I don't think you have to make it any better just because it's related to something you're working on.

I had a few separate suggestions to make this shorter, but due to the hashing you can't really do that and have it work. In any case... 

```suggestion
    def test_it(self, quantized_expiry):
        service = _NGINXSigner(nginx_server="http://nginx", secret="not_a_secret")

        signed_url = service.sign_url("https://other/?q=s", "/path/")

        quantized_expiry.assert_called_once_with(max_age=timedelta(hours=25))
        assert (
            signed_url
            == "http://nginx/path/cYa67Tuf4V3j6KuQsAuBNw/1581120000/https://other/?q=s"
        )

    @pytest.fixture(autouse=True)
    def quantized_expiry(self, patch):
        return patch(
            "via.services.pdf_url.quantized_expiry",
            return_value=datetime(year=2020, month=2, day=8, tzinfo=timezone.utc),
        )
```

tests/unit/via/services/pdf_url_test.py
```diff
@@ -0,0 +1,136 @@
...
+        assert svc == PDFURLBuilder.return_value
+
+    @pytest.fixture
+    def PDFURLBuilder(self, patch):
+        return patch("via.services.pdf_url.PDFURLBuilder")
+
+    @pytest.fixture
+    def _NGINXSigner(self, patch):
+        return patch("via.services.pdf_url._NGINXSigner")
```
jon-betts: I always feel so dumb writing these types of tests. I can't see a better way to do it though?

tests/unit/services.py
```diff
@@ -39,3 +41,13 @@ def secure_link_service(mock_service):
 @pytest.fixture
 def http_service(mock_service):
     return mock_service(HTTPService)
+
+
+@pytest.fixture
+def nginx_signer_service(mock_service):
+    return mock_service(_NGINXSigner)
```
jon-betts: I don't think this is a service as we never register it to the services registry in the app code. Would this be better inside `via/tests/unit/via/services/pdf_url_test.py:TestPDFURLBuilder`?

tests/unit/via/views/view_pdf_test.py
```diff
@@ -100,17 +73,62 @@ def google_drive_api(self, google_drive_api):
...
-            "via.views.view_pdf.quantized_expiry",
-            return_value=datetime(
-                year=2020,
-                month=2,
-                day=8,
-                hour=17,
-                minute=30,
-                second=21,
-                tzinfo=timezone.utc,
-            ),
```
jon-betts, line 115: Nice to see the view testing code getting a bit more manageable.

via/services/pdf_url.py
```diff
@@ -0,0 +1,90 @@
...
+from typing import Callable
+
+from h_vialib.secure import quantized_expiry
+
+from via.services.google_drive import GoogleDriveAPI
+from via.services.secure_link import SecureLinkService
+
+
+@dataclass
+class _NGINXSigner:
```
jon-betts, line 14: Good call on the private here. This shouldn't really leak outside this file (and it's tests).

via/services/pdf_url.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@dataclass
+class PDFURLBuilder:
+    google_drive_api: GoogleDriveAPI
+    secure_link_service: SecureLinkService
+    route_url: Callable
+    nginx_signer: _NGINXSigner
+
+    def get_pdf_url(self, url):
```
jon-betts, line 58: A doc string here might help. My suggestion of `get_pdf_url` is a bit vague out of context.

via/views/view_pdf.py
jon-betts: Depends on how puritanical you want to be. Technically it's a view concern as it's only for the needs of Pyramid's error handling and return values that we need to do this. The services can return a pristine generator and you can reify it or not as you see fit.

I think if we have a unified `HTTPSession` which both regular and Google Drive calls go through it's much more of a stylistic question. (At the moment you'd still need the helper either way).

As an aside, it _seems_ like you'd want to make this optional if you look at `via/get_url.py`. As we make a streaming call, and then _purposefully_ don't iterate the content at all. But I think all this means is you'd use the non-streaming call and pass `stream=True`. And presumably add a big comment explaining why you aren't mad.

via/views/view_pdf.py
```diff
@@ -23,81 +23,61 @@ def view_pdf(context, request):
...
+        file_id=request.matchdict["file_id"],
+        resource_key=request.matchdict.get("resource_key"),
+    )
+
+    return _iter_pdf_response(request.response, content_iterable)
+
+
+def _iter_pdf_response(response, content_iterable):
+
+    try:
```
jon-betts:
```suggestion
    try:
```

Saving the whales one new line at a time. :whale:

via/views/view_pdf.py
```diff
@@ -23,81 +23,61 @@ def view_pdf(context, request):
...
+        # Get an iterable where the first item only has already been reified.
+        # Return None if there is no content to return.
+        # This takes a potentially lazy generator, and ensures the first item is
+        # called now. This is so any errors or problems that come from starting the
+        # process happen immediately, rather than whenever the iterable is evaluated.
```
jon-betts:
```suggestion
        # This takes a potentially lazy generator, and ensures the first item is
        # called now. This is so any errors or problems that come from starting the
        # process happen immediately, rather than whenever the iterable is evaluated.
```

I think your comment is better, and the old one is out of date now?

tests/unit/via/services/pdf_url_test.py
```diff
@@ -0,0 +1,136 @@
...
+        secure_link_service,
+        pyramid_request,
+        nginx_signer_service,
+    ):
+        return PDFURLBuilder(
+            google_drive_api,
+            secure_link_service,
+            pyramid_request.route_url,
+            nginx_signer_service,
+        )
```
jon-betts, line 103: I mentioned it above, but belt and braces and all that. I  think the `_NGINXSigner` is a private helper rather than service. We don't register and access it using `pyramid-services` so I think a regular fixture here would be fine.

jon-betts (changes requested):
I feel like I've really done you dirty with this whole process and given you a fun new set of problems to work on.

This PR is marked as `Request changes` but there's actually only one change I think you should make (the rest are all just ideas). Some of it is just a result of reviewing old code now it's been moved about.

The one thing is I don't think `_NGINXSigner` is a service, as we don't register it with Pyramid services in the app. So I think it would be enough to have the unit tests define a local fixture for it. This also means the tests would more closely mirror how this is instantiated in the code.

tests/unit/services.py
marcospri: Yup, no need for it to be a service

via/views/view_pdf.py
marcospri: I think we should move most of what `via/get_url.py` does inside HTTPService and maybe a better name for `.stream()` would be `stream_body` or similar.

tests/unit/via/services/pdf_url_test.py
```diff
@@ -0,0 +1,137 @@
...
+        google_drive_api,
+        secure_link_service,
+        pyramid_request,
+        NGINXSigner,
+    ):
+        return PDFURLBuilder(
+            google_drive_api,
+            secure_link_service,
+            pyramid_request.route_url,
+            NGINXSigner,
```
jon-betts:
```suggestion
            NGINXSigner.return_value,
```
I'm confused why this doesn't cause a problem. Shouldn't this pass an instance instead of the mocked class?

tests/unit/via/services/pdf_url_test.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+        assert svc == PDFURLBuilder.return_value
+
+    @pytest.fixture
+    def PDFURLBuilder(self, patch):
+        return patch("via.services.pdf_url.PDFURLBuilder")
+
+
+@pytest.fixture
+def NGINXSigner(patch):
```
jon-betts:
```suggestion
def _NGINXSigner(patch):
```

I think our style is to exactly match the class name.

jon-betts (commented):
This is all looking good, but there's one bit in the tests I don't understand. 

The `PDFURLBuilder` accepts an `_NGINXSigner` instance, but we are passing in the mocked class (not an instance). I'd expect this to fail?

tests/unit/via/services/pdf_url_test.py
marcospri: `_NGINXSigner` clashes with the import of the real clash, not sure which one to alias or if having `_NGINXSigner` as a class fixture in two classes (instead of a global one) is better.

marcospri: Used a different name to avoid the name clash, but to something less subtle than removing the underscore `PatchedNGINXSigner`

tests/unit/via/services/pdf_url_test.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+        assert svc == PDFURLBuilder.return_value
+
+    @pytest.fixture
+    def PDFURLBuilder(self, patch):
+        return patch("via.services.pdf_url.PDFURLBuilder")
+
+
+@pytest.fixture
+def PatchedNGINXSigner(patch):
```
jon-betts, line 136: Yep. :shrug: What can you do?

jon-betts (approved):
It's just a consequence of how we normally name things that it's prone to name space collisions. I think this is fine.

PR #658, opened 2021-10-26
Proxy MS OneDrive PDFs in a python view

marcospri:
Part of: https://github.com/hypothesis/lms/issues/3256

# Testing notes

URLs: 
- one drive `https://api.onedrive.com/v1.0/shares/u!aHR0cHM6Ly8xZHJ2Lm1zL2IvcyFBaU9PUWVsRWNZS3diZnVvb3JRdmNYYnR2cjQ/root/content`

- sharepoint `https://listhypothesis-my.sharepoint.com/:b:/g/personal/eng_list_hypothes_is/EQS1fwBTvJRJkgiKu_qz-oIBce4F4t2zIWdSVWSMF2MEOg?e=TLbqHg&download=1`

Try to proxy both on http://localhost:9083/

On `master` one drive url gets proxied, sharepoint one fails. On this branch both are successful.

### DIFF
--- a/tests/unit/via/services/pdf_url_test.py
+++ b/tests/unit/via/services/pdf_url_test.py
@@ -77,6 +77,30 @@ class TestPDFURLBuilder:
         secure_link_service.sign_url.assert_called_once_with(url)
         assert pdf_url == secure_link_service.sign_url.return_value
 
+    @pytest.mark.parametrize(
+        "url,endpoint_url",
+        (
+            # pylint:disable=line-too-long
+            (
+                "https://my-sharepoint.sharepoint.com/FILE_ID/?download=1",
+                "http://example.com/onedrive/proxied.pdf?url=https%3A%2F%2Fmy-sharepoint.sharepoint.com%2FFILE_ID%2F%3Fdownload%3D1",
+            ),
+            (
+                "https://api.onedrive.com/v1.0/FILE_ID/root/content",
+                "http://example.com/onedrive/proxied.pdf?url=https%3A%2F%2Fapi.onedrive.com%2Fv1.0%2FFILE_ID%2Froot%2Fcontent",
+            ),
+        ),
+    )
+    def test_onedrive_url(
+        self, svc, secure_link_service, google_drive_api, url, endpoint_url
+    ):
+        google_drive_api.parse_file_url.return_value = None
+
+        pdf_url = svc.get_pdf_url(url)
+
+        secure_link_service.sign_url.assert_called_once_with(endpoint_url)
+        assert pdf_url == secure_link_service.sign_url.return_value
+
     def test_nginx_file_url(self, google_drive_api, svc):
         google_drive_api.parse_file_url.return_value = None
 
--- a/tests/unit/via/views/view_pdf_test.py
+++ b/tests/unit/via/views/view_pdf_test.py
@@ -5,7 +5,7 @@ from h_matchers import Any
 from pyramid.httpexceptions import HTTPNoContent
 
 from via.resources import QueryURLResource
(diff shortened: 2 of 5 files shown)


tests/unit/via/views/view_pdf_test.py
```diff
@@ -42,21 +42,6 @@ def test_it_builds_the_url_if_not_google(
             "https://example.com/foo/bar.pdf?q=s"
         )
 
-    def test_it_builds_the_url_if_google(
```
marcospri, line 45: This was re-testing the PDFURLBuilder service. Keept just one of the use cases and test every view independently.

tests/unit/via/services/pdf_url_test.py
```diff
@@ -77,6 +77,31 @@ def test_google_file_url(
...
+            (
+                "https://my-sharepoint.sharepoint.com/FILE_ID/?download=1",
+                "http://example.com/one_drive/proxied.pdf"
+                "?url=https%3A%2F%2Fmy-sharepoint.sharepoint.com%2FFILE_ID%2F%3Fdownload%3D1",
+            ),
+            (
+                "https://api.onedrive.com/v1.0/FILE_ID/root/content",
+                "http://example.com/one_drive/proxied.pdf"
+                "?url=https%3A%2F%2Fapi.onedrive.com%2Fv1.0%2FFILE_ID%2Froot%2Fcontent",
+            ),
```
seanh: It might actually be better to let the long lines just be long lines here. It looked to me like each tuple actually had three separate strings in it, which was confusing, until I noticed

tests/unit/via/views/view_pdf_test.py
seanh: :+1:

seanh: Isn't the service mocked in view tests? It should be

via/views/__init__.py
```diff
@@ -20,6 +20,12 @@ def add_routes(config):  # pragma: no cover
         "/google_drive/{file_id}/{resource_key}/proxied.pdf",
         factory=QueryURLResource,
     )
+    config.add_route(
+        "proxy_one_drive_pdf",
+        "/one_drive/proxied.pdf",
```
seanh: It's actually "OneDrive" (all one word) :)

```suggestion
        "proxy_onedrive_pdf",
        "/onedrive/proxied.pdf",
```

I'd recommend this throughout, e.g. `_MS_ONEDRIVE_URL` etc.

Also because there are actually three logical parts to `"proxy_foo_pdf"` not four, so even if the `foo` in the middle really was two separate words you'd run them together. This is how `_`'s are used in names in the standard library and most big third-party libraries: `bytecode_helper` (not `byte_code_helper`), `itertools` not `iter_tools`, etc. Basically you insert the `_`'s when they're necessary for readability, not between every word

seanh (approved):

via/views/view_pdf.py
```diff
@@ -39,6 +40,17 @@ def view_pdf(context, request):
...
 
+@view_config(route_name="proxy_one_drive_pdf", decorator=(has_secure_url_token,))
+def proxy_one_drive_pdf(context, request):
+    url = context.url_from_query()
+
+    content_iterable = request.find_service(HTTPService).stream(
+        url, headers=add_request_headers({})
+    )
+
+    return _iter_pdf_response(request.response, content_iterable)
```
seanh: Nice and simple view :+1:

via/views/__init__.py
marcospri: Updated to "ONEDRIVE" all around.

PR #660, opened 2021-10-27
Handle errors during the streaming of the response

marcospri:
Correctly handle errors that happen during the streaming on the response (ie happen while  exhausting of the iterator as opposed to when calling  calling. `requests.requests `)

### DIFF
--- a/via/services/http.py
+++ b/via/services/http.py
@@ -83,34 +83,37 @@ class HTTPService:
                 **kwargs,
             )
             response.raise_for_status()
+
         except Exception as err:
-            if self._error_translator and (mapped := self._error_translator(err)):
-                raise mapped from err
+            if mapped_err := self._translate_exception(err):
+                raise mapped_err from err  # pylint: disable=raising-bad-type
 
-            if not isinstance(err, RequestException):
-                # If the error_translator didn't handle a non request exception, raise it directly
-                raise err
-
-            translated_class = UnhandledUpstreamException
-            for (  # pragma: no branch
-                error_class,
-                target_class,
-            ) in DEFAULT_ERROR_MAP.items():
-                if isinstance(err, error_class):
-                    translated_class = target_class
-                    break
-
-            raise translated_class(
-                message=err.args[0] if err.args else None,
-                requests_err=err if hasattr(err, "request") else None,
-            ) from err
+            raise
 
         return response
 
     def stream(self, url, method="GET", **kwargs):
         response = self.request(method=method, url=url, stream=True, **kwargs)
+        try:
+            yield from self._stream_bytes(response)
+        except Exception as err:
+            if mapped_err := self._translate_exception(err):
+                raise mapped_err from err  # pylint: disable=raising-bad-type
 
-        yield from self._stream_bytes(response)
+            raise
+
+    def _translate_exception(self, err):
+        if self._error_translator and (mapped := self._error_translator(err)):
+            return mapped
+
+        for (error_class, target_class) in DEFAULT_ERROR_MAP.items():
+            if isinstance(err, error_class):
+                return target_class(
+                    message=err.args[0] if err.args else None,
+                    requests_err=err if hasattr(err, "request") else None,
+                )
+
+        return None
 
     @staticmethod
     def _stream_bytes(response, min_chunk_size=64000):
--- a/tests/unit/via/services/http_test.py
+++ b/tests/unit/via/services/http_test.py
@@ -128,7 +128,7 @@ class TestHTTPService:
         )
 
     @pytest.mark.parametrize("request_exception,expected_exception", EXCEPTION_MAP)
-    def test_with_error_translator_defaults(
+    def test_request_with_error_translator_defaults(
         self,
         error_translator,
         session,
@@ -143,6 +143,26 @@ class TestHTTPService:
         with pytest.raises(expected_exception):
             svc.request("GET", url)
 
+    @pytest.mark.parametrize("generator_exception,expected_exception", EXCEPTION_MAP)
+    def test_stream_with_error_translator_defaults(
+        self,
+        error_translator,
+        session,
+        generator_exception,
+        expected_exception,
+        url,
+    ):
+        svc = HTTPService(session, error_translator)
+
+        def explode():
+            yield b"response-part-1"
+            raise generator_exception
+
+        session.request.return_value.iter_content.return_value = explode()
+
+        with pytest.raises(expected_exception):
+            list(svc.stream("GET", url))
+
     def test_with_custom_exceptions(self, error_translator, session, url):
         svc = HTTPService(session, error_translator)
 
@@ -151,13 +171,21 @@ class TestHTTPService:
         with pytest.raises(NotADirectoryError):
             svc.request("GET", url)
 
-    def test_with_regular_exceptions(self, error_translator, session, url):
+    def test_request_with_regular_exceptions(self, error_translator, session, url):
         svc = HTTPService(session, error_translator)
 
         session.request.side_effect = ValueError
 
         with pytest.raises(ValueError):
             svc.request("GET", url)
+
+    def test_stream_with_regular_exceptions(self, error_translator, session, url):
+        svc = HTTPService(session, error_translator)
+
+        session.request.return_value.iter_content.side_effect = ValueError
+
+        with pytest.raises(ValueError):
+            list(svc.stream("GET", url))
 
     @pytest.mark.parametrize(
         "input_bytes,output",


via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
-            raise translated_class(
-                message=err.args[0] if err.args else None,
-                requests_err=err if hasattr(err, "request") else None,
-            ) from err
+            raise self._translate_exception(err) from err
 
         return response
 
     def stream(self, url, method="GET", **kwargs):
         response = self.request(method=method, url=url, stream=True, **kwargs)
```
marcospri, line 96: Exceptions occurring here are raised before consuming the generator (eg BadUrl, ConnectionError)

via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
-
-        yield from self._stream_bytes(response)
+        try:
+            yield from self._stream_bytes(response)
+        except Exception as err:
+            raise self._translate_exception(err) from err
+
+    def _translate_exception(self, err):
```
marcospri, line 105: This is not super pretty but having a function handles the differences in function vs generator.

via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
+        if self._error_translator and (mapped := self._error_translator(err)):
+            return mapped
+
+        if not isinstance(err, RequestException):
+            # If the error_translator didn't handle a non request exception, raise it directly
+            return err
+
+        translated_class = UnhandledUpstreamException
```
jon-betts: I think all of this is handled by the contents of `DEFAULT_ERROR_MAP`. It only maps RequestExceptions and it has a default which maps to UnhandledUpstreamException. So I think this can be removed.

via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
+        for (  # pragma: no branch
+            error_class,
+            target_class,
```
jon-betts: I think you might be getting branch coverage issues here because of the situation described above. It's not possible to fall out of this without a value.

via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
+        ) in DEFAULT_ERROR_MAP.items():
+            if isinstance(err, error_class):
+                translated_class = target_class
+                break
+
+        return translated_class(
+            message=err.args[0] if err.args else None,
+            requests_err=err if hasattr(err, "request") else None,
+        )
```
jon-betts: What was the idea around removing the early return here?

via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
         return response
 
     def stream(self, url, method="GET", **kwargs):
         response = self.request(method=method, url=url, stream=True, **kwargs)
-
-        yield from self._stream_bytes(response)
+        try:
+            yield from self._stream_bytes(response)
+        except Exception as err:
+            raise self._translate_exception(err) from err
```
jon-betts: There's a slight change of behavior here from the original method. It's possible for `err` to not end up being mapped. In which case we end up doing `raise err from err` which will duplicate the error in all the reporting (and possibly be a bit confusing).

In the original formulation it was something like:


```suggestion
            if mapped_err := self._translate_exception(err):
                raise self._translate_exception(err) from err

            raise
```

You'd want to do this in the one above as well.

jon-betts (commented):
I think this will work. There are a few features about it which I had thoughts about.

There's double error handling happening here. We have one set around the call to `request` and then another around the streaming portion. I think this is actually fine. If you dig into `requests` it looks like the requesting portion happens synchronously it just doesn't pull the content. So even if we don't evaluate the generator produced until later, as soon as we do, we'll enter our `request()` method syncrhonously and the first set of error handling should kick in. Then the second lot should happen in the `yield from`. So I _think_ it all handles it well.

If you wanted to avoid this, you could have an abstract internal `_request` method that both `request` and `stream` call to do the common portion and then have 100% of the code inside the error handling. This might be a bit easier to be sure is ok. Even though I think this is fine too.

I think there's a coverage problem because there's a quite hard to spot unreachable code path. The `DEFAULT_ERROR_MAP` actually enforces both the fact that all errors must be children of `RequestException` by it's choice of mapped errors, and it also has a default of `UnhandledUpstreamException` added by the last item in the mapping. I think means some of the tests don't need to happen before we try and map.

There's also a slight change of behavior from before, in that it's now possible to end up in a situation where you raise an exception `from` itself. Which could be confusing, or at least lead to log bloat.

via/services/http.py
marcospri: Yes I needed the `# pragma: no branch` there.

via/services/http.py
marcospri: For lines 103-105, I think it's neeed, there's this test https://github.com/hypothesis/via/blob/2aa3a43b63aca853f12ef22f80c6aeb36f94923b/tests/unit/via/services/http_test.py#L174

which I understood as trying not to mask non-request exceptions we might be ignoring otherwise (ie a bug produces a `ValueError` exception but we just report it as a problem upstream)

For `translated_class = UnhandledUpstreamException`, yeah, that's not necessary. We either use the DEFAULT_MAP or a function so this was superfluous.

via/services/http.py
```diff
@@ -83,34 +83,42 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
+
+        for (  # pragma: no branch
+            error_class,
+            target_class,
+        ) in DEFAULT_ERROR_MAP.items():
+            if isinstance(err, error_class):
+                return target_class(
+                    message=err.args[0] if err.args else None,
+                    requests_err=err if hasattr(err, "request") else None,
+                )
```
marcospri, line 114: No `return None` here as exhausting the loop is not possible. Also still needed the `pragma: no branch`.

I think that's better that a return None a `#no-cover` there? :shrug:

via/services/http.py
```diff
@@ -83,34 +83,42 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
+
+            raise
+
+    def _translate_exception(self, err):
+        if self._error_translator and (mapped := self._error_translator(err)):
+            return mapped
+
+        if not isinstance(err, RequestException):
+            # If the error_translator didn't handle a non request exception, the caller can raise it direclty
+            return None
```
marcospri: Changed this to return None. This is the case that could cause `err from err` to happen on the calling code.

jon-betts: I don't think you need this statement at all. If the error isn't an instance of `RequestException` it won't match in the for loop below.

If you remove this and have a `return None` at the bottom, you can remove both of:

```python 
# pylint: disable=inconsistent-return-statements
...
# pragma: no branch
```

It gets a little shorter as a bonus too.

via/services/http.py
```diff
@@ -83,34 +83,46 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
+        if not isinstance(err, RequestException):
+            # If the error_translator didn't translate a non request exception
+            # the caller musht raise it directly
+            return None
+
+        for (  # pragma: no branch
+            error_class,
+            target_class,
+        ) in DEFAULT_ERROR_MAP.items():
+            # DEFAULT_ERROR_MAP contains RequestException , it will always find a match
```
jon-betts: If we did change this about as I suggest I don't think the comment would be accurate anymore

jon-betts (commented):
I think there's still some stuff we can do to fix the lint / coverage warnings that we're getting here. I think they are actually both valid and trying to tell us something.

At the moment we have something like:

```
if not RequestException:
    return None
loop that only matches RequestException
# implicit return None < unreachable
```

I think the implicit return None is the cause of both of the warnings. We can fix this instead by allowing the loop to run and then using the fact it doesn't map `RequestException`s instead:

```
loop that only matches RequestException
# Implicitly not a `RequestException`
return None 
```

This is slighly less efficient (as we go through the loop), but in the case when your app is crashing, you can take your time. This won't be in a tight loop ever.

via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
 
     def stream(self, url, method="GET", **kwargs):
         response = self.request(method=method, url=url, stream=True, **kwargs)
-
-        yield from self._stream_bytes(response)
+        try:
+            yield from self._stream_bytes(response)
+        except Exception as err:
+            if mapped_err := self._translate_exception(err):
+                raise mapped_err from err  # pylint: disable=raising-bad-type
```
marcospri, line 101: pylint doesn't understand this if's guard: https://github.com/PyCQA/pylint/issues/2529

via/services/http.py
```diff
@@ -83,34 +83,40 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
...
+            raise
+
+    def _translate_exception(self, err):
+        if self._error_translator and (mapped := self._error_translator(err)):
+            return mapped
+
+        for (
+            error_class,
+            target_class,
+        ) in DEFAULT_ERROR_MAP.items():
```
jon-betts: Now there's no pylint comment this can all be one one line I think

jon-betts (approved):
This is looking quite shiny now!

PR #661, opened 2021-10-29
Make raise_for_status optional in HTTPService

marcospri:
`get_url_details` is ok with 4xx and 5xx responses but HTTPService.request did raise_for_status by default.

Testing

- On the `http-service` branch, proxy https://h.readthedocs.io/_/downloads/en/latest/pdf/
from http://localhost:9083/

   There will be a lot of noise around a favicon.ico failed (404) request which will be translated to an underhanded exception.

- On this branch, testing the same thing logs just this: 

   ```2021-10-29 10:03:18,667 DEBUG [urllib3.connectionpool:452][MainThread] https://h.readthedocs.io:443 "GET /_/downloads/en/latest/pdf/favicon.ico HTTP/1.1" 404 None```

### DIFF
--- a/tests/unit/via/services/http_test.py
+++ b/tests/unit/via/services/http_test.py
@@ -127,6 +127,18 @@ class TestHTTPService:
             {"status_code": status}
         )
 
+    @pytest.mark.parametrize("status", [400, 401, 403, 404, 500])
+    def test_it_doesnt_raise_if_the_response_is_an_error_when_disabled(
+        self, svc, url, status
+    ):
+        httpretty.register_uri("GET", url, status=status)
+
+        response = svc.request("GET", url, raise_for_status=False)
+
+        assert response == Any.instance_of(requests.Response).with_attrs(
+            {"status_code": status}
+        )
+
     @pytest.mark.parametrize("request_exception,expected_exception", EXCEPTION_MAP)
     def test_request_with_error_translator_defaults(
         self,
--- a/via/services/http.py
+++ b/via/services/http.py
@@ -49,7 +49,7 @@ class HTTPService:
     def delete(self, *args, **kwargs):
         return self.request("DELETE", *args, **kwargs)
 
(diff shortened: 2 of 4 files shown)


tests/unit/via/services/http_test.py
```diff
@@ -127,6 +127,18 @@ def test_it_raises_if_the_response_is_an_error(self, svc, url, status):
             {"status_code": status}
         )
 
+    @pytest.mark.parametrize("status", [400, 401, 403, 404, 500])
```
marcospri, line 130: I had a parameterized version of the previous tests `test_it_raises_if_the_response_is_an_error`, but I think a separate test is clearer even if they are very similar.

via/services/http.py
```diff
@@ -82,7 +83,8 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
                 timeout=timeout,
                 **kwargs,
             )
-            response.raise_for_status()
+            if raise_for_status:
```
marcospri, line 86: There some risk of dumping here lots of conditionals and special cases here but I don't think this one is a bad one.

via/services/http.py
```diff
@@ -65,6 +65,7 @@ def request(self, method, url, timeout=(10, 10), **kwargs):
             Note that the read_timeout is *not* a time limit on the entire
             response download. It's a time limit on how long to wait *between
             bytes from the server*. The entire download can take much longer.
+        :param raise_for_status: Optionally raise for 4xx & 3xx response statuses.
```
jon-betts:
```suggestion
        :param raise_for_status: Optionally raise for 4xx & 5xx response statuses.
```

jon-betts (approved):
Yep looks good.

We should keep an eye on this though. At present we have three different consumers of these calls I think, and they all use them in distinct ways. There is a tipping point at which we're forcing everyone through the same gate and then having to cater for each separately on the other side.

PR #666, opened 2021-11-04
Diagnostic "fix" for errors proxying single slash URLs

jon-betts:
This is an attempt to rule in or out the fact we are receiving URLs with single slashes after the scheme. If we are, then this might also help them work again while we track down what's going on.

This has two commits:

 * The first should be functionally identical and everything should work the same
 * The second should allow single slash URLs


seanh (approved):
Let's give it a try

PR #668, opened 2021-11-04
Schemaless nginx signing

jon-betts:


jon-betts:
This was an attempt to help diagnose / fix a live issue. I don't think we need it.

PR #687, opened 2022-01-06
Ignore the deprecation warnings from Marshmallow for now.

jon-betts:
There's no public fix for this yet. As mentioned in https://github.com/marshmallow-code/marshmallow/pull/1903, a change has resulted in Marshmallow raising a deprecation warning. As we're pretty strict about warnings this has bombed out our build.

## Testing notes

For this to show up you need to remove your `.tox/` directory. Then run `make sure` and you should see failures / it working depending on whether you are in master or this PR.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,13 @@
 [pytest]
 filterwarnings =
+    # See: https://docs.python.org/3/library/warnings.html#the-warnings-filter
+
     error
+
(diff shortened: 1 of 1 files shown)


lyzadanger (approved):
Approved by this person who also saw the same approach used on ~~LMS~~ checkmate and nothing seems to be heavily on fire.

PR #701, opened 2022-03-08
JSTOR pdf proxing based on a new jstor:// schema

marcospri:
Here we rely on a `jstor://` URL coming from LMS

## Testing

- `export JSTOR_PDF_URL=<>`
- Configure a new URL assignment with `jstor://10.1086/508573` and launch it.

Tested with both: 

https://www.jstor.org/stable/1234 
https://www.jstor.org/stable/10.1086/508573

### When not configured

Remove any configured PDF_URL

```unset JSTOR_PDF_URL```

- Launch the assignment, you'll get a 401 in the `http://localhost:9083/jstor/proxied.pdf?url=jstor%3A%2F%2F10.1086%2F508573\u0026jstor.ip=3.141.59.26` endpoint.

### DIFF
--- a/tests/unit/via/views/view_pdf_test.py
+++ b/tests/unit/via/views/view_pdf_test.py
@@ -2,10 +2,15 @@ from unittest.mock import sentinel
 
 import pytest
 from h_matchers import Any
-from pyramid.httpexceptions import HTTPNoContent
+from pyramid.httpexceptions import HTTPNoContent, HTTPUnauthorized
 
 from via.resources import QueryURLResource
-from via.views.view_pdf import proxy_google_drive_file, proxy_onedrive_pdf, view_pdf
+from via.views.view_pdf import (
+    proxy_google_drive_file,
+    proxy_jstor_pdf,
+    proxy_onedrive_pdf,
+    view_pdf,
+)
 
 
 @pytest.mark.usefixtures(
@@ -138,6 +143,34 @@ class TestProxyOneDrivePDF:
         response = call_view("https://one-drive.com", view=proxy_onedrive_pdf)
 
         assert isinstance(response, HTTPNoContent)
+
+
+@pytest.mark.usefixtures("secure_link_service", "pdf_url_builder_service")
+class TestJSTORPDF:
+    @pytest.mark.usefixtures("http_service")
+    def test_it(self, call_view, jstor_api, pyramid_request):
+        response = call_view(
+            "jstor://DOI", view=proxy_jstor_pdf, params={"jstor.ip": "1.1.1.1"}
+        )
+
+        assert response.status_code == 200
+        assert response.headers["Content-Disposition"] == "inline"
+        assert response.headers["Content-Type"] == "application/pdf"
+        assert (
+            response.headers["Cache-Control"]
+            == "public, max-age=43200, stale-while-revalidate=86400"
+        )
+        jstor_api.jstor_pdf_stream.assert_called_once_with(
+            "jstor://DOI", pyramid_request.params["jstor.ip"]
+        )
+
+    def test_when_not_enabled(self, call_view, jstor_api):
+        jstor_api.enabled = False
+
+        with pytest.raises(HTTPUnauthorized):
+            call_view(
+                "jstor://DOI", view=proxy_jstor_pdf, params={"jstor.ip": "1.1.1.1"}
+            )
 
 
 @pytest.fixture
--- a/tox.ini
+++ b/tox.ini
@@ -48,6 +48,7 @@ passenv =
     dev: NGINX_SERVER
     dev: CLIENT_EMBED_URL
     dev: SIGNED_URLS_REQUIRED
+    dev: JSTOR_PDF_URL
 deps =
     dev: -r requirements/dev.txt
     tests: -r requirements/tests.txt
--- a/via/services/jstor.py
+++ b/via/services/jstor.py
@@ -1,0 +1,32 @@
+from via.requests_tools.headers import add_request_headers
+from via.services import HTTPService
+
+
+class JSTORAPI:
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

**Comment by robertknight** [comment]
This PR worked as expected after updating environment variables. I did get a pretty confusing error though before I updated the name of the `JSTOR_PDF_URL` env var:

```
requests.exceptions.MissingSchema: Invalid URL 'None/10.1086/508573?ip=...': No scheme supplied.
```

**Comment by marcospri** [comment]
> This PR worked as expected after updating environment variables. I did get a pretty confusing error though before I updated the name of the `JSTOR_PDF_URL` env var:

I think, both to avoid this error and to use it as a feature toggle (eg not having jstor proxing in the public via) I'll check for this value before attempting to use the API.

File: via/services/jstor.py [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, http_service: HTTPService, url):
+        self._jstor_pdf_url = url
+        self._http = http_service
+
+    @property
+    def enabled(self):
+        return bool(self._jstor_pdf_url)
+
+    def jstor_pdf_stream(self, jstor_url: str, jstor_ip: str):
+        doi = jstor_url.replace("jstor://", "")
```
marcospri, line 18: This is not very sophisticated parsing but I think is enough for now.

via/views/view_pdf.py
```diff
@@ -67,6 +67,18 @@ def proxy_google_drive_file(request):
     return _iter_pdf_response(request.response, content_iterable)
 
 
+@view_config(route_name="proxy_jstor_pdf", decorator=(has_secure_url_token,))
+def proxy_jstor_pdf(context, request):
+    jstor_api = request.find_service(JSTORAPI)
+    if not jstor_api.enabled:
+        raise HTTPUnauthorized()
```
marcospri: The UI for this is pretty bad. The request to the pdf fails but pdf.js doesn't show too much indication of what's going on. 

I think for us, for debugging etc, it's enough.

robertknight:
The URLs that are currently associated with annotations on JSTOR assignments are the `jstor://{article_id}` custom URL and the PDF fingerprint. This will be OK for an initial demo but we need to give this some more thought before deploying to users. We might want to make this: 1) The JSTOR article URL (`https://www.jstor.com/stable/{article_id}`) or the URL at which JSTOR normally serves the PDF (`https://www.jstor.com/stable/pdf/{article_id}`). For articles that have a real DOI, it would be good to also associate that with the article. We don't currently have a way of telling if that is the case though.

Via JSTOR URL

tests/unit/via/views/view_pdf_test.py
```diff
@@ -140,6 +145,34 @@ def test_it_can_stream_an_empty_iterator(self, http_service, call_view):
...
+        response = call_view(
+            "jstor://DOI", view=proxy_jstor_pdf, params={"jstor.ip": "1.1.1.1"}
+        )
+
+        assert response.status_code == 200
+        assert response.headers["Content-Disposition"] == "inline"
+        assert response.headers["Content-Type"] == "application/pdf"
+        assert (
+            response.headers["Cache-Control"]
+            == "public, max-age=43200, stale-while-revalidate=86400"
```
robertknight, line 161: I guess these are the default caching headers that we use for all PDFs? I do think some caching will be beneficial here, as PDFs can be large and it takes time for the JSTOR API to serve them. This caching will have some impact on eg. any analytics that JSTOR does, which we'll need to account for.

tox.ini
```diff
@@ -48,6 +48,7 @@ passenv =
     dev: NGINX_SERVER
     dev: CLIENT_EMBED_URL
     dev: SIGNED_URLS_REQUIRED
+    dev: JSTOR_PDF_URL
```
robertknight, line 51: It would be a good idea to re-sort these variables, unless there is a reason they need to be in this particular order?

via/services/jstor.py
```diff
@@ -0,0 +1,31 @@
+from via.requests_tools.headers import add_request_headers
+from via.services import HTTPService
+
+
+class JSTORAPI:
+    DEFAULT_DOI_PREFIX = "10.2307"
```
robertknight, line 6: This could do with a comment to indicate that is JSTOR's main DOI prefix, and that it is used both for real DOIs and pseudo-DOIs which have the format `10.2307/{jstor_article_id}` but are not actually registered DOIs.

via/services/jstor.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def jstor_pdf_stream(self, jstor_url: str, jstor_ip: str):
+        doi = jstor_url.replace("jstor://", "")
+        if not "/" in doi:
+            doi = f"{self.DEFAULT_DOI_PREFIX}/{doi}"
+
+        url = f"{self._jstor_pdf_url}/{doi}?ip={jstor_ip}"
+        return self._http.stream(
+            url, headers=add_request_headers({"Accept": "application/pdf"})
+        )
```
robertknight, line 25: This will do for the demo but it would be wise to follow-up by adding some validation of the DOI and IP formats, so we don't end up calling JSTOR's API with some invalid user-controlled path.

via/views/view_pdf.py
robertknight: Is there a way of adding an error message here that will be exposed at least in the HTTP response if the developer is looking at that?

robertknight (approved):
LGTM with some minor comments. This is good to ship but adding some more robust validation of the JSTOR URL would be wise once we've got the initial demo working.

tests/unit/via/views/view_pdf_test.py
marcospri: Yes this are the defaults for every PDF.

PR #707, opened 2022-03-14
Configure client to show JSTOR content banner

robertknight:
When serving content from JSTOR using their APIs, configure the client
to show a banner indicating where the content is from and providing a
link to JSTOR.

See also https://github.com/hypothesis/client/pull/4308 and https://github.com/hypothesis/private-issues/issues/99.


lyzadanger (approved):
This looks very straightforward.

PR #720, opened 2022-04-06
Update PDF.js to v2.14.137 and improve automation/documentation of process 

robertknight:
Update PDF.js to v2.14.137, matching https://github.com/hypothesis/browser-extension/pull/799 (almost, this update is a few minor updates newer).

The `bin/update-pdfjs` script has also been updated to improve the documentation on how to run it and auto-commit the results. In particular the comments in the script now note that it should be run via `make update-pdfjs` rather than being run directly as a user would do in other repositories. This is necessary to set up the Python environment for the `bin/create_pdf_template.py` script. Auto-committing the result ensures that the PDF.js version number is included in the commit message and the correct directories are committed. A similar change was recently made for the browser extension in https://github.com/hypothesis/browser-extension/pull/799.


bin/update-pdfjs
```diff
@@ -9,9 +9,10 @@ set -eu
 #
 # To update PDF.js to the latest version:
 #
-# 1. Create a new branch and run this script.
-# 2. Verify that PDFs work as expected in py-server.
-# 3. Commit the changes to the `vendor/` and `templates/` directories.
+# 1. Create a new branch and run `make update-pdfjs`. This will download the
+#    newest release of PDF.js, update the HTML template for the viewer and
+#    commit the changes.
```
lyzadanger, line 14: I'm a little concerned that this script looks so similar to that in other projects that devs might "read past" this different instruction. What "happens" if this script is run directly instead of with `make update-pdfjs`. Is it possible to build in a safety check?

robertknight: Yes. I have pushed a commit that adds a check, using the presence of the `TOX_ENV_NAME` var to test if the script is being run directly or via tox.

lyzadanger (approved):
Thanks for safeguarding that script!

PR #729, opened 2022-04-21
Catch and re-raise Checkmate `BadURL`s to trigger our nice UI

jon-betts:
Needs:

 * https://github.com/hypothesis/checkmatelib/pull/21

This will cause BadURL messages to be presented to the user rather than buried.

## Testing notes

At the moment it's quite hard to get `checkmatelib` to raise, so I'd suggest manually bodging `via/checkmate.py` to raise an error:

```python
    def raise_if_blocked(self, url):
        """Raise a redirect to Checkmate if the URL is blocked.

        This will sensibly apply all ignore reasons and other configuration for
        Checkmate.

        :param url: The URL to check
        :raises HTTPTemporaryRedirect: If the URL is blocked
        :raises BadURL: For malformed or private URLs
        """
        try:
            raise CheckmateBadURL("TERRIBLE NEWS!")

            blocked = self.check_url(
                url,
                allow_all=self._request.registry.settings["checkmate_allow_all"],
                blocked_for=self._request.params.get("via.blocked_for"),
                ...
```

Then you can run `make dev` visit http://localhost:9083/ and attempt to visit any site.

* You should see the error page blaming you for the mistake
* In the details box you should see the string from your error

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -54,7 +54,7 @@ google-auth-oauthlib==0.5.1
     # via -r requirements/requirements.txt
 gunicorn==20.1.0
     # via -r requirements/requirements.txt
-h-checkmatelib==1.0.11
+h-checkmatelib==1.0.12
     # via -r requirements/requirements.txt
 h-pyramid-sentry==1.2.3
     # via -r requirements/requirements.txt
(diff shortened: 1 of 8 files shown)


robertknight:
> At the moment it's quite hard to get checkmatelib to raise, so I'd suggest manually bodging via/checkmate.py to raise an error:

Do you mean that it's hard because the recent change to what constitutes a "bad URL" hasn't been integrated in checkmatelib yet?

robertknight (approved):
The code changes look good. I haven't done the manual testing part, but I'll leave that to you.

Issue #734, opened 2022-05-03
Accept a JSTOR id from LMS and send it to JSTOR


jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/110
 
Depends on:

 * https://github.com/hypothesis/lms/issues/3906

## What we need to do

LMS will be storing and passing us an ID. We need to accept it and pass it on to JSTOR. We need to:

*  Accept this value
 * Integrate with JSTOR to pass it on (looking like JWT at the moment?)

Details of the auth method can be found here: https://hypothes-is.slack.com/archives/C02T75RKBTK/p1651073373706719

## Tasks

 * https://github.com/hypothesis/devdata/pull/48
 * https://github.com/hypothesis/devdata/pull/50
 * https://github.com/hypothesis/via/pull/737
 * https://github.com/hypothesis/via/pull/738
 * https://github.com/hypothesis/via/pull/739
 * https://github.com/hypothesis/via/pull/735
 * https://github.com/hypothesis/via/pull/736
 * https://github.com/hypothesis/devdata/pull/51
 * [x] ... and AWS

jon-betts:
Done

PR #737, opened 2022-05-05
Add downloading of the `devdata.env` file for secret env vars

jon-betts:
For:

 * https://github.com/hypothesis/via/issues/734

This is being done specifically for JSTOR, but it's a general mechanism.

## Testing notes

 * `make devdata`
 * Notice the `.devdata.env` file is created

### DIFF
--- a/bin/devdata.py
+++ b/bin/devdata.py
@@ -1,4 +1,4 @@
-"""Download .devdata from github.com:hypothesis/devdata.git."""
+"""Download dev data from github.com:hypothesis/devdata.git."""
 
 import os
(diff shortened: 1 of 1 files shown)


marcospri (approved):
We could look into doing some glob magic to import all .json/.env that are in devdata/{project_name}...

We don't change this that often so maybe not worth it. Definitively not worth doing it right now.

jon-betts:
Yeah. I was wondering if we could have some general manifest mechanism that explains where each file should go, and is kept in dev data.

PR #738, opened 2022-05-05
Don't check JSTOR URLs against Checkmate

jon-betts:
For:

 * https://github.com/hypothesis/via/issues/734

This really makes me wonder if we aren't creating a pile of pain for ourselves here by encoding these as `jstor://` URLs in the first place? I wonder if we should have left them as the original types which would have some level of graceful degradation for people without JSTOR setup.

What we could have done is include another parameter, or expand the type hints we pass to Via.

In any case, this takes a fairly blunt way of achieving this.

## Testing notes

 * `make devdata` In both LMS and Via
 * Enable JSTOR as described here: https://github.com/hypothesis/lms/pull/3918 (minus the site code)
 * Create a JSTOR assignment with:  https://www.jstor.org/stable/resrep25332.5
 * On master you should see a domain warning
 * In this branch you don't (it still fails though)

### DIFF
--- a/tests/unit/via/checkmate_test.py
+++ b/tests/unit/via/checkmate_test.py
@@ -32,6 +32,13 @@ class TestViaCheckmateClient:
             ignore_reasons=sentinel.checkmate_ignore_reasons,
         )
 
+    def test_we_do_not_check_jstor_urls(self, client, check_url, JSTORAPI):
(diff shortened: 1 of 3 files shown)


marcospri (approved):
Yeah, this works :+1: 

As we talked before let's decide about the `jstor://` schema once everything is in place :+1:

PR #739, opened 2022-05-05
Add streaming Base64 decoding in the JSTOR service

jon-betts:
Currently JSTOR returns us Base64 encoded PDFs, whether we like it or not. This isn't intentional and they are going to try and fix it, but in the mean time we need to deal with it.

One option is to download the entire PDF, decode it, and then send it to the user. This is bad for a number of reasons:

 * The time to first byte for the user is worse than streaming
 * We have to hold the entire document in memory to decode it
 * Our system is heavily setup to expect a stream, not a string (for the reasons above)

This PR adds streaming base 64 decoding of the PDF content.

## Testing notes

 * `make devdata` In both LMS and Via
 * Enable JSTOR as described here: https://github.com/hypothesis/lms/pull/3918 (minus the site code)
 * Create a JSTOR assignment with any value: e.g. https://www.jstor.org/stable/resrep25332.5
 * On master you should see a broken PDF with malformed content
 * In this branch it should work

### DIFF
--- a/via/services/jstor.py
+++ b/via/services/jstor.py
@@ -1,3 +1,5 @@
+import base64
+
 from via.requests_tools.headers import add_request_headers
 from via.services import HTTPService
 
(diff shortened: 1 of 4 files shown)


via/services/jstor.py
```diff
@@ -36,10 +38,51 @@ def stream_pdf(self, url, jstor_ip):
...
 
+        # Currently, JSTOR is sending us a stream of base 64 encoded data. This
+        # isn't intentional and may change if they can fix it. For the time
+        # being we need to decode this on the fly. This is both because it
+        # gives the user a better time to first byte, but also because the
+        # machinery that consumes this expects a stream, not a string.
+        return decode_base64_stream(stream)
+
+
+def decode_base64_stream(b64_stream):
```
marcospri, line 53: Nice, much better than reading the whole thing just to decode it (and change all the machinery around streaming for pdfs).

We might have to delete it if JSTOR changes back to honor the pdf header.

marcospri (approved):

via/services/jstor.py
jon-betts: Yes. I have a horrible feeling we might have to try and see if it's Base64 on the fly as well.

We might be able to do something pretty cheesy for this though, as the first few bytes of a PDF are probably always the same.

PR #741, opened 2022-05-10
Remove all JSTOR code as LMS now provides a URL

jon-betts:
For:

 * https://github.com/hypothesis/private-issues/issues/110

I've asked JSTOR and they are happy for these short lived URLs to be exposed publicly.

### Testing notes

 * Get the latest dev data (after https://github.com/hypothesis/devdata/pull/52)
 * All the usual `make dev` magic to stand everything up
 * Make sure your dev tools are open with the cache disabled
 * Visit: https://hypothesis.instructure.com/courses/125/assignments/2968
 * You should see a document, nothing externally should be any different
 * If you inspect the proxied URL you will see the S3 temp URL passed from LMS


jon-betts:
This is probably getting superceded by https://github.com/hypothesis/lms/pull/3941

I might convert this into a full removal

marcospri (approved):
Good, nice to have all of this all in LMS instead.

There's some rebase needed here and we probably should only merge this until we confirm the test assignment works in prod with the LMS changes.

PR #770, opened 2022-07-13
Detect files blocked by Google Drive as malicous

jon-betts:
See: 

 * https://hypothes-is.slack.com/archives/C4K6M7P5E/p1657657579981149

When Google denies access to a particular file because it thinks it's malicious, we currently emit a 417 indicating a general failure. This catches this special case and allocates a new HTTP code for it.

### Review notes

There are two commits:

 * The first implements this directly and is should be easy to see
 * The second refactors the layout of the function to avoid a lint message

### DIFF
--- a/via/services/google_drive.py
+++ b/via/services/google_drive.py
@@ -72,6 +72,17 @@ def translate_google_error(error):
             requests_err=error,
         )
 
+    if status_code == 403 and google_reason == "cannotDownloadAbusiveFile":
+        # This file is blocked because Google thinks it is malware of some kind
+        return GoogleDriveServiceError(
+            "Google has identified this file as malicious and has denied "
+            "access to it",
+            # 423 - Locked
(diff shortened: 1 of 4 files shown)


via/services/google_drive.py
```diff
@@ -31,6 +31,31 @@ class Errors(Schema):
...
+    },
+    (403, "cannotDownloadFile"): {
+        # This seems to happen if a file is locked down in various ways
+        # rather than for general consumption
+        "message": "We do not have permission to download the file through the"
+        " Google Drive API",
+        "status_int": 403,
+    },
+    (403, "cannotDownloadAbusiveFile"): {
+        "message": "Google has identified this file as malicious and has "
```
marcospri, line 50: I thought pylint didn't love implicit string concatenation.

marcospri (approved):

PR #815, opened 2022-10-25
Use the new `deploy.yml` workflow

seanh:
I think we'll have to merge this to see if it actually works: use the new `deploy.yml` workflow from https://github.com/hypothesis/workflows/pull/2, specify a different GitHub environment name for each Elastic Beanstalk environment, and also specify a URL for each environment.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,50 +22,56 @@ jobs:
       Application: ${{ github.event.repository.name }}
     secrets: inherit
 
-  qa-via:
+  QA:
     needs: create-image
-    name: ${{ github.event.repository.name }}
-    uses: hypothesis/workflows/.github/workflows/eb-update.yml@main
+    uses: hypothesis/workflows/.github/workflows/deploy.yml@main
     with:
-      Application: ${{ github.event.repository.name }}
-      Environment: qa
-      Region: us-west-1
-      Operation: deploy
-      Version: latest
+      operation: deploy
+      github_environment_name: QA
+      github_environment_url: https://qa-via.hypothes.is/
+      aws_region: us-west-1
+      elasticbeanstalk_application: via
+      elasticbeanstalk_environment: qa
+      docker_tag: ${{ needs.create-image.outputs.docker_tag }}
     secrets: inherit
 
-  qa-lms-via:
+  QA_Edu:
+    name: QA (Edu)
     needs: create-image
-    name: lms-via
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -22,50 +22,54 @@ jobs:
...
+      github_environment_name: QA
+      github_environment_url: https://qa-via.hypothes.is/
+      aws_region: us-west-1
+      elasticbeanstalk_application: via
+      elasticbeanstalk_environment: qa
+      version: ${{ needs.create-image.outputs.docker_tag }}
     secrets: inherit
 
-  qa-lms-via:
+  QA (Edu):
```
seanh: Not sure if the space and parens in the labels will work or not. If it doesn't work I might need to change the job label to `QA_Edu` and add a `name: QA (Edu)` field below.

.github/workflows/deploy.yml
```diff
@@ -22,50 +22,54 @@ jobs:
...
-    uses: hypothesis/workflows/.github/workflows/eb-update.yml@main
+    uses: hypothesis/workflows/.github/workflows/deploy.yml@main
     with:
-      Application: lms-via
-      Environment: qa
-      Region: us-west-1
-      Operation: deploy
-      Version: latest
+      operation: deploy
+      github_environment_name: QA (Edu)
```
seanh, line 44: I'm gonna call these `QA (Edu)` and `Production (Edu)` instead of `QA (LMS)` and `Production (LMS)`. It's confusing to see the name of one app (`LMS`) in the name of an environment belonging to another app (`Via`). When you see something like `Via / Production (LMS)` it's unclear whether you're looking at the name of a Via environment or an LMS environment since both app names Via and LMS are in the environment name. We've already committed this mistake in the Elastic Beanstalk application and environment names and domain names: is https://lms-via.hypothes.is/ is an instance of the LMS app or of Via? But there's no need to continue the confusion in the GitHub environment labels

indigobravo (approved):
I can see where you are going with the `EDU` naming. I think it will provide a better experience in GitHub, and will be easier for new starters to get their heads around.

It has left me wondering if we should spin up the `edu-via-*` Elastic Beanstalk environments. The confusing `lms-via-*` environments could then be retired. I'm happy to schedule the work for that.

seanh:
I'm not sure what I did but somehow I seem to have gotten this PR into a confused state where GitHub seems to think that it has unresolved conflicts even though I have resolved them locally and force-pushed. Closing in favour of https://github.com/hypothesis/via/pull/818

PR #818, opened 2022-10-27
Use the new deploy.yml workflow

seanh:
Use the new `deploy.yml` workflow from
hypothesis/workflows#2, specify a different
GitHub environment name for each Elastic Beanstalk environment, and also
specify a URL for each environment.


seanh:
This is the same change that was already approved in https://github.com/hypothesis/via/pull/815#pullrequestreview-1157747796

PR #823, opened 2022-11-03
Use better GitHub environment URLs

seanh:
Replace the QA and production environment URLs with direct links to
proxy PDFs in Via, and replace the Via (Edu) QA and production
environment URLs with links to PDF assignments in our test Canvas
instance (you can't use Via (Edu) directly, you have to go through the
LMS app).


seanh:
This is kind of trivial so I'm just gonna go ahead and merge it

PR #829, opened 2022-11-22
Apply cookiecutter to deployment workflows

seanh:
Apply the `deploy.yml` and `redeploy.yml` templates from https://github.com/hypothesis/cookiecutters to the `deploy.yml` and `redeploy.yml` files in this repo.

As expected this doesn't make any significant changes to the workflow files since this repo already contains hand-created `deploy.yml` and `redeploy.yml` workflows, just some minor formatting changes as a result of applying the template.

This PR was generated by first manually creating the `.cookiecutter/includes/.github/workflows/environments.json` that's included in this PR, and then:

```console
$ # Run the cookiecutter over Via.
$ cookiecutter gh:hypothesis/cookiecutters --directory pyramid-app --output-dir .. --overwrite-if-exists --no-input slug=via
$ # For some reason template.txt isn't generated.
$ make requirements/template.txt
$ # Run `make template` to generate deploy.yml and redeploy.yml.
$ make template
$ # Undo the git commit the cookiecutter makes.
$ git reset HEAD^
$ # Undo all the unwanted changes the cookiecutter made.
$ git checkout .github/dependabot.yml .gitignore .python-version Dockerfile LICENSE Makefile README.md conf/ docker-compose.yml package.json pyproject.toml requirements/ tox.ini
$ rm -rf .docker.env bin/make_python bin/make_template my_pyramid_app requirements/prod.in requirements/prod.txt requirements/template.in requirements/template.txt setup.cfg tests/functional/app_test.py tests/pyproject.toml tests/unit/my_pyramid_app/ .cookiecutter/cookiecutter.json
$ # We should now be left with just the deployment workflow changes that we want.
$ git status --untracked-files=all
On branch apply-cookiecutter-to-deployment-workflows
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   .github/workflows/deploy.yml
	modified:   .github/workflows/redeploy.yml

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.cookiecutter/includes/.github/workflows/environments.json

no changes added to commit (use "git add" and/or "git commit -a")
```

I then renamed `continuous-integration.yml` to `ci.yml` since that's what the cookiecutter (and this PR's `deploy.yml` that was generated by the cookiecutter) calls it.


.github/workflows/deploy.yml
```diff
@@ -1,30 +1,24 @@
 name: Deploy
-
 concurrency:
-  group: ${{ github.event.repository.name }}-deploy
+  group: deploy
```
jon-betts, line 3: Are groups presumably inherently local to the repo, so putting the repo in is a waste of time?

.cookiecutter/includes/.github/workflows/environments.json
```diff
@@ -0,0 +1,30 @@
+{
+  "QA": {
+    "github_environment_url": "https://qa-via.hypothes.is/https://en.wikipedia.org/api/rest_v1/page/pdf/Comet_Kohoutek",
+    "aws_region": "us-west-1",
+    "elasticbeanstalk_application": "via",
+    "elasticbeanstalk_environment": "qa"
+  },
+  "QA_Edu": {
```
jon-betts, line 9: Casing looks a bit of an odd duck here. As this is only an ID and the name take precedence in the UI would this be better as `qa_edu`, or `QAEdu`?

Now might be a good time to revisit the base file, as I see we have `CI` in there as opposed to `ci` or `continuous-integration`?

.github/workflows/deploy.yml
```diff
@@ -1,30 +1,24 @@
...
     branches:
       - main
-
 jobs:
-  continuous-integration:
-    uses: ./.github/workflows/continuous-integration.yml
-    name: continuous integration
-
+  CI:
+    uses: ./.github/workflows/ci.yml
```
jon-betts, line 13: Losing the name here seems like a step backwards? ~I presume this is coming from the base template.~ Yep, I've put some notes on the base ticket.

jon-betts (approved):
This seems to work generate the file. Looking at the file I can't see any issues, but with CI stuff you never really know until it runs.

I think the choice of camel case with underscores is unusual and makes the names and ids harder to spot the difference between visually. Now would seem a good time to change that if we wanted to, before this pattern is rolled out widely.

seanh:
Just updated this with the final changes from https://github.com/hypothesis/cookiecutters/pull/95

.github/workflows/deploy.yml
seanh: Yep, changed this in the cookiecutter

.github/workflows/deploy.yml
seanh: I think so yes, that seems clear from the docs. If I'm wrong I guess we'll find out!

PR #831, opened 2022-12-06
Temporarily switch Via to be a Debian Slim-based container

### LINKED ISSUE #1409: LMS app fails to load U of Montana-Western OneDrive files
## Bug report form

We're not sure if this is a bug with our proxy server, or a security configuration in the school's OneDrive account. We were able to meet with the instructional designer to get a demo of the error. We **are** able to access files from their OneDrive account using the "Anyone with the link" shared link, so this is not a simple user error for the file permissions. They replicated the problem across multiple courses and Microsoft accounts.

### Steps to reproduce

1. Open a new Hypothesis assignment in Moodle
2. Use the "Select PDF from OneDrive" option
3. Create the assignment
4. Attempt to view the assignment

### Expected behaviour

The LMS app loads the OneDrive file with the proxy server, alongside the client. 

### Actual behaviour

Error Message: Could Not Get Web Page

```
UpstreamServiceError: HTTPSConnectionPool(host='universityofmontanawestern-my.sharepoint.com', port=443): Max retries exceeded with url: /:b:/g/personal/leslie_johnson_umwestern_edu/Eb_FY9OuE2ZDoy1AxXvTqMABfBKgdbgmiXgkx3gzBkCPwA?download=1 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f36f09dfc70>: Failed to establish a new connection: [Errno -2] Name does not resolve'))

URL: https://lms-via.hypothes.is/route?via.client.ignoreOtherConfiguration=1&via.client.openSidebar=1&via.external_link_mode=new-tab&via.client.requestConfigFromFrame.origin=https%3A%2F%2Flms.hypothes.is&via.client.requestConfigFromFrame.ancestorLevel=2&via.blocked_for=lms&url=https%3A%2F%2Funiversityofmontanawestern-my.sharepoint.com%2F%3Ab%3A%2Fg%2Fpersonal%2Fleslie_johnson_umwestern_edu%2FEb_FY9OuE2ZDoy1AxXvTqMABfBKgdbgmiXgkx3gzBkCPwA%3Fdownload%3D1&via.sec=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJoIjoib3BJc21GQVJmc0p2SzY3N1lheEwiLCJleHAiOjE2Njk5MzIwMDB9.XGHK8_DEAhhNZiHaBM2MaAzBfGiXf7--I2YKsZChOuo

Third party URL: https://universityofmontanawestern-my.sharepoint.com/:b:/g/personal/leslie_johnson_umwestern_edu/Eb_FY9OuE2ZDoy1AxXvTqMABfBKgdbgmiXgkx3gzBkCPwA?download=1

```

image

### LMS

Moodle

### Additional details

After meeting with the instructional designer, we learned that the Microsoft administrator for the school quit earlier that day, having recently done some work related to security policies...

Links to tickets: 

- https://app.hubspot.com/contacts/6291320/ticket/1294759463
- https://app.hubspot.com/contacts/6291320/ticket/1298575489

robertknight:
This PR works around a limitation in Musl libc's DNS resolver, lack of support for TCP retries, by temporarily switching the base image of Via to be based on Debian Slim (which uses glibc instead of Musl) rather than Alpine Linux. This limitation caused problems when Via tried to connect to certain Sharepoint-hosted domains, and after some experimentation/exploration, I concluded that changing the base image in the short term is probably going to be the least unpleasant/risky solution (aside from deciding not to fix the problem at all).

See https://github.com/hypothesis/product-backlog/issues/1409#issuecomment-1338447951 for a detailed explanation of the issue and possible solutions.

Musl is due to get support for TCP retries in the next release (v1.2.4?), and will hopefully appear in Alpine 3.18. So we should be able to switch Via back to Alpine at that point, for consistency with our other apps. The Alpine-based image is also smaller. The compressed Debian-based one is ~107MB whereas the compressed Alpine-based image is 58MB.

The first commit updates the Docker file. The second adds a `run-docker` command to assist with testing locally.

Fixes https://github.com/hypothesis/product-backlog/issues/1409.

**Testing:**

1. Run h, viahtml and the client locally
2. Check out this branch, and create a local commit that patches `conf/nginx/includes/resolver.conf` to change the DNS resolver from AWS's resolver to 1.1.1.1 or 8.8.8.8
3. Run `make docker`
4. Run `make run-docker`
5. Visit http://localhost:9083/https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,7 @@ FROM python:3.8.15-slim-bullseye
 LABEL maintainer="Hypothes.is Project and contributors"
 
 # Install nginx & supervisor
-RUN apt-get update && apt-get install --yes nginx nginx-extras gettext-base \
-  && pip install --no-cache-dir supervisor \
+RUN apt-get update && apt-get install --yes nginx nginx-extras gettext-base supervisor \
   && apt-get clean
 
 # Create the hypothesis user, group, home directory and package directory.
@@ -33,4 +32,4 @@ COPY . .
 
 USER hypothesis
 
-CMD /usr/bin/envsubst '$${NGINX_SECURE_LINK_SECRET}' < /var/lib/hypothesis/nginx_envsubst.conf.template > /var/lib/hypothesis/nginx_envsubst.conf && /usr/local/bin/supervisord -c /var/lib/hypothesis/conf/supervisord.conf
+CMD /usr/bin/envsubst '$${NGINX_SECURE_LINK_SECRET}' < /var/lib/hypothesis/nginx_envsubst.conf.template > /var/lib/hypothesis/nginx_envsubst.conf && /usr/bin/supervisord -c /var/lib/hypothesis/conf/supervisord.conf
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,29 @@
-FROM python:3.8.12-alpine3.13
+# Unlike most Hypothesis projects this Docker image is based on Debian,
+# so it can use glibc's DNS resolver which supports TCP retries. It can be
+# reverted back to Alpine when Musl v1.2.4 is released.
+#
+# See https://github.com/hypothesis/product-backlog/issues/1409
+FROM python:3.8.15-slim-bullseye
 LABEL maintainer="Hypothes.is Project and contributors"
 
 # Install nginx & supervisor
-RUN apk add --no-cache nginx gettext supervisor
+RUN apt-get update && apt-get install --yes nginx nginx-extras gettext-base \
+  && pip install --no-cache-dir supervisor \
+  && apt-get clean
 
 # Create the hypothesis user, group, home directory and package directory.
-RUN addgroup -S hypothesis && adduser -S -G hypothesis -h /var/lib/hypothesis hypothesis
+RUN groupadd --system hypothesis && useradd --system -g hypothesis --home /var/lib/hypothesis hypothesis
 WORKDIR /var/lib/hypothesis
 
 # Ensure nginx state and log directories writeable by unprivileged user.
-RUN chown -R hypothesis:hypothesis /etc/nginx/conf.d /var/log/nginx /var/lib/nginx
+RUN chown -R hypothesis:hypothesis /etc/nginx/conf.d /var/log/nginx /var/lib/nginx /var/lib/hypothesis
 
 # Copy minimal data to allow installation of python dependencies.
 COPY requirements/requirements.txt ./
 
 # Install build deps, build, and then clean up.
-RUN apk add --no-cache --virtual build-deps \
-  && pip install --no-cache-dir -U pip \
-  && pip install --no-cache-dir -r requirements.txt \
-  && apk del build-deps
+RUN pip install --no-cache-dir -U pip \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,24 +1,29 @@
...
+#
+# See https://github.com/hypothesis/product-backlog/issues/1409
+FROM python:3.8.15-slim-bullseye
 LABEL maintainer="Hypothes.is Project and contributors"
 
 # Install nginx & supervisor
-RUN apk add --no-cache nginx gettext supervisor
+RUN apt-get update && apt-get install --yes nginx nginx-extras gettext-base \
+  && pip install --no-cache-dir supervisor \
+  && apt-get clean
```
seanh: Could `supervisor` be added to the `apt-get install` command, instead of having a separate `pip` install command for it? That seems more equivalent to what used to happen with Alpine (using `apk` to install `supervisor`). `pip` may get a more recent version than `apt` does but Supervisor isn't updated often.

This was already wrong in the original Alpine-based `Dockerfile` anyway: we should be creating a virtualenv and installing a pinned version of Supervisor from a requirements file like we do with the rest of our Python dependencies. But that's a separate issue.

seanh (approved):
:+1:

This looks fine to me. I think honestly if I ever get round to doing a review of our Docker containers (and their build times) we may want to switch them all over to Debian or Ubuntu anyway. Those are the most mainstream Linux distros and what the team is familiar with and uses locally. Some good advice on Docker base images for Python apps here: https://pythonspeed.com/articles/base-image-python-docker-images/

Apparently the compressed size of [python:3.8.15-slim-bullseye](https://hub.docker.com/layers/library/python/3.8.15-slim-bullseye/images/sha256-3778adf3c19e35a6cbb7e7ef85c8715e4062f0292e7e8f876a84cc90657d2126) is only 45MB. The compressed the size is presumably what matters for uploading to and downloading from Docker Hub. But perhaps the uncompressed size affects the time taken to build the compressed image in the first place and/or the time taken to start a container?

robertknight:
> Apparently the compressed size of [python:3.8.15-slim-bullseye](https://hub.docker.com/layers/library/python/3.8.15-slim-bullseye/images/sha256-3778adf3c19e35a6cbb7e7ef85c8715e4062f0292e7e8f876a84cc90657d2126) is only 45MB. The compressed the size is presumably what matters for uploading to and downloading from Docker Hub. But perhaps the uncompressed size affects the time taken to build the compressed image in the first place and/or the time taken to start a container?

The compressed size affects upload/downloads from Docker Hub. The uncompressed size will affect anything that involves operating on the bytes of the image. How much of a practical difference it will make in our PR builds and deployment pipeline I'm not sure. There are other factors too such as apk vs apt.

Alpine has served us well for 7+ years, so I'm in no rush to change it everywhere. From what I gather in my reading, what has changed in that time is the availability of slim Ubuntu/Debian images that are more competitive with Alpine size-wise.

PR #845, opened 2022-12-21
Proxy D2L files using python

marcospri:
Needs https://github.com/hypothesis/h-vialib/pull/42 merged and published before CI is green.

For 
- https://github.com/hypothesis/lms/issues/4768

## Testing

Follow the instructions over: https://github.com/hypothesis/lms/pull/4792

### DIFF
--- a/via/requests_tools/headers.py
+++ b/via/requests_tools/headers.py
@@ -2,6 +2,8 @@
 
 from collections import OrderedDict
 
+from h_vialib.secure import SecureSecrets
+
 # A mix of headers we don't want to pass on for one reason or another
 BANNED_HEADERS = {
     # Requests needs to set Host to the right thing for us
@@ -75,11 +77,22 @@ def clean_headers(headers):
     return clean
 
 
-def add_request_headers(headers):
+def add_request_headers(headers, request=None):
     """Add headers for 3rd party providers which we access data from."""
 
     # Pass our abuse policy in request headers for third-party site admins.
     headers["X-Abuse-Policy"] = "https://web.hypothes.is/abuse-policy/"
     headers["X-Complaints-To"] = "https://web.hypothes.is/report-abuse/"
+
+    if request and "via.secret.headers" in request.params:
+        # Pass along any headers sent in the original request
+        secure_secrets = SecureSecrets(
+            request.registry.settings["via_secret"].encode("utf-8")
+        )
+
+        secret_headers = secure_secrets.decrypt_dict(
+            request.params["via.secret.headers"]
+        )
+        headers.update(secret_headers)
 
     return headers
--- a/via/requests_tools/headers.py
+++ b/via/requests_tools/headers.py
@@ -2,7 +2,7 @@
 
 from collections import OrderedDict
 
-from h_vialib.secure import SecureSecrets
+from h_vialib.secure import Encryption
 
 # A mix of headers we don't want to pass on for one reason or another
 BANNED_HEADERS = {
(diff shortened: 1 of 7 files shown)


via/requests_tools/headers.py
```diff
@@ -75,11 +77,22 @@ def clean_headers(headers):
...
 
-def add_request_headers(headers):
+def add_request_headers(headers, request=None):
     """Add headers for 3rd party providers which we access data from."""
 
     # Pass our abuse policy in request headers for third-party site admins.
     headers["X-Abuse-Policy"] = "https://web.hypothes.is/abuse-policy/"
     headers["X-Complaints-To"] = "https://web.hypothes.is/report-abuse/"
 
+    if request and "via.secret.headers" in request.params:
```
marcospri, line 87: Adding this here so every request benefits from it (at least the python ones).

via/services/pdf_url.py
```diff
@@ -73,20 +75,35 @@ def get_pdf_url(self, url):
...
     @classmethod
     def _is_onedrive_url(cls, url):
         return any(regexp.match(url) for regexp in cls._MS_ONEDRIVE_URL)
 
-    def _proxy_onedrive_pdf(self, url):
+    @classmethod
+    def _is_d2l_url(cls, url):
+        return any(regexp.match(url) for regexp in cls._D2L_URL)
+
+    def _proxy_python_pdf(self, url, route="proxy_python_pdf"):
```
marcospri, line 93: Making this method a bit generic instead of copy/paste a similar version.

via/views/__init__.py
```diff
@@ -21,10 +21,10 @@ def add_routes(config):  # pragma: no cover
         factory=QueryURLResource,
     )
     config.add_route(
-        "proxy_onedrive_pdf",
-        "/onedrive/proxied.pdf",
-        factory=QueryURLResource,
+        "proxy_onedrive_pdf", "/onedrive/proxied.pdf", factory=QueryURLResource
     )
+    config.add_route("proxy_d2l_pdf", "/d2l/proxied.pdf", factory=QueryURLResource)
+    config.add_route("proxy_python_pdf", "proxied.pdf", factory=QueryURLResource)
```
marcospri, line 27: This last route is a bit extra and not used but I reckon it leads to a nice "just proxy these with python" feature in the future without too much noise.

via/views/view_pdf.py
```diff
@@ -41,11 +41,17 @@ def view_pdf(context, request):
...
+def proxy_python_pdf(context, request):
+    """Proxy a pdf with python (as opposed to nginx).
+
+    Multiple routes point to this view to allow having separate access logs for each of them.
+    """
     url = context.url_from_query()
 
     content_iterable = request.find_service(HTTPService).stream(
-        url, headers=add_request_headers({})
+        url, headers=add_request_headers({}, request=request)
```
marcospri, line 54: Passing the request here get the original headers.

via/services/pdf_url.py
```diff
@@ -73,20 +75,35 @@ def get_pdf_url(self, url):
             return self._google_file_url(file_details, url)
 
         if self._is_onedrive_url(url):
-            return self._proxy_onedrive_pdf(url)
+            return self._proxy_python_pdf(url, route="proxy_onedrive_pdf")
+
+        if self._is_d2l_url(url):
```
marcospri, line 80: I briefly tried the nginx approach but not sure if it was possible to:

- Pass the headers to nginx in a secure way
- Do that in a way that involved sane code.

seanh (approved):
Not much to say, this all looks sensible.

It might be worth having a check in New Relic at some point to see how Via's performance is doing with now proxying lots of files in Python instead of NGINX.

PR #903, opened 2023-04-26
[Poc] Youtube annotations 

marcospri:
Needs: https://github.com/hypothesis/h-vialib/pull/51


via/get_url.py
```diff
@@ -24,6 +25,10 @@ def get_url_details(http_service, url, headers=None):
     if GoogleDriveAPI.parse_file_url(url):
         return "application/pdf", 200
 
+    if YoutubeAPI.parse_file_url(url):
+        # TODO is inventing a mimetype the right approach?
```
marcospri, line 29: We do a lot mimetype, content types dances around here. I'd refactor this using an enum inside the client library to simplifiy all of this.

via/services/__init__.py
```diff
@@ -38,6 +45,12 @@ def create_google_api(settings):
     )
 
 
+def create_youtube_api(settings):
```
marcospri, line 48: Left over from using google's API.

PR #905, opened 2023-04-28
Add frontend infrastructure for video player app

robertknight:
This PR updates the frontend asset building and serving infrastructure in Via to align with how it works in lms and h, using the [h-assets](https://github.com/hypothesis/h-assets) package, and then leverages that to set up a basic skeleton for a client-rendered video player app. Please pardon the large size of this PR, that's mostly due to changes in the npm/yarn lockfiles.

There is some existing Via-specific setup for serving PDF.js from files in `via/static/js/` which I left as-is in this PR. In future it would make sense to unify these two static asset serving systems. PDF.js is a very large JS app however so some changes to our standard build tools might be required to support it.

**Summary of changes:**

1. Set up frontend asset serving using the h-assets package that we also use in lms and h. The use of h-assets in this PR closely follows that of the lms repo.
2. Add frontend build infrastructure using our standard tech stack (Rollup, Preact etc.). As part of this, the npm package manager was changed from npm to yarn (`package-lock.json => yarn.lock`), to match the client/lms repos
3. Add a basic skeleton for a video player app in `via/static/scripts/video_player` using Preact. This doesn't have tests yet. Those will come after I add testing infrastructure in a subsequent PR.
4. Add a Pyramid route for `/video/{id}` that serves the video player app. At some point in future we will need to support video providers other than YouTube. This might look something like `/video/{provider}/{id}`.

There are two pieces of frontend infrastructure that are not included in this PR:

- Tailwind / CSS setup
- Testing + CI using Karma + Chrome (see https://github.com/hypothesis/via/pull/907)

I will add these in subsequent PRs.

**Testing:**

1. Check out this PR and go to `http://localhost:9083/video/{youtube_video_id}` where `youtube_video_id` is a valid ID for any public YouTube video
2. You should see an unstyled page that contains the Hypothesis client, a YouTube video player iframe and a couple of lines from a dummy transcript. Please ignore the lack of styling. The point of this PR is just to get basic infrastructure and plumbing in place.
3. Kill Via's dev server, run `make docker`, then `make run-docker`, then repeat steps (1) and (2) to see that this all works in the Docker container as well.


via/static/scripts/video_player/config.ts
```diff
@@ -0,0 +1,32 @@
+import type { Transcript } from './components/Transcript';
```
robertknight: This module follows the same pattern as `frontend_apps/config.ts` in the LMS repo, but is much simpler since we only have one app with very simple configuration.

via/static/scripts/video_player/components/VideoPlayer.tsx
```diff
@@ -0,0 +1,35 @@
+export type VideoPlayerProps = {
+  videoId: string;
+
+  /** Set the current timestamp of the video. */
+  time: number;
```
robertknight, line 5: This timestamp and the callbacks below are not actually used yet. They will be once more of the UI is fleshed out in subsequent PRs.

gulpfile.mjs
```diff
@@ -0,0 +1,67 @@
+import {
+  buildCSS,
+  buildJS,
+  generateManifest,
+  runTests,
+  watchJS,
+} from '@hypothesis/frontend-build';
+import gulp from 'gulp';
+
+// import tailwindConfig from './tailwind.config.mjs';
```
robertknight, line 10: The CSS and testing parts of this are not used yet. This will come in follow-up PRs.

Dockerfile
```diff
@@ -1,3 +1,15 @@
+# Stage 1: Build frontend assets
```
robertknight, line 1: The structure here matches the lms app's Dockerfile.

acelaya (approved):
👍🏼

PR #906, opened 2023-05-01
Suppress warning about pkg_resources being deprecated

robertknight:
This package is imported by Pyramid and it triggers a warning:

```
DeprecationWarning: pkg_resources is deprecated as an API
```

See also https://github.com/hypothesis/h/pull/7950 and https://hypothes-is.slack.com/archives/C1MA4E9B9/p1682933798408359.


robertknight:
As this change has already been merged into other repos (eg. https://github.com/hypothesis/checkmate/pull/673) I'm going to go ahead and merge this.

PR #907, opened 2023-05-01
Add frontend test infrastructure for video player app

robertknight:
**Depends on https://github.com/hypothesis/via/pull/905**

This PR adds our standard frontend testing infrastructure to Via for use with the video player app, and a simple test module (`HypothesisClient-test.js`). It also adds a GitHub CI workflow to run the tests.

The infrastructure is copied and tweaked from the lms repo, minus a few vestigial things I removed. One intentional change is that the file that sets up the global JS environment in the tests is called `setup-tests.js` in this repo, rather than `bootstrap.js` as in other repos. This seems like a much more obvious name to me.


acelaya (approved):
> One intentional change is that the file that sets up the global JS environment in the tests is called setup-tests.js in this repo, rather than bootstrap.js as in other repos. This seems like a much more obvious name to me.

Agree

PR #908, opened 2023-05-01
Set up ESLint

robertknight:
**Depends on https://github.com/hypothesis/via/pull/907**

Set up JS/TS linting, using the configuration from the hypothesis/lms project as a base. Also fix a few lint errors reported by the initial config.


acelaya (approved):
👍🏼

PR #909, opened 2023-05-02
Add CSS / Tailwind configuration and build process

robertknight:
**Depends on https://github.com/hypothesis/via/pull/908**

Set up the CSS / Tailwind build. This largely follows the structure and config in the hypothesis/lms repo, with the LMS-specific Tailwind extensions removed.

### DIFF
--- a/tailwind.config.mjs
+++ b/tailwind.config.mjs
@@ -1,0 +1,22 @@
+import tailwindConfig from '@hypothesis/frontend-shared/lib/tailwind.preset.js';
+
+export default {
+  presets: [tailwindConfig],
+  content: [
+    './via/static/scripts/**/*.{js,ts,tsx}',
+    './node_modules/@hypothesis/frontend-shared/lib/**/*.js',
+  ],
+  theme: {
+    extend: {
+      fontFamily: {
+        sans: [
+          '"Helvetica Neue"',
+          'Helvetica',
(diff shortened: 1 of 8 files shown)


tailwind.config.mjs
```diff
@@ -0,0 +1,22 @@
...
+
+export default {
+  presets: [tailwindConfig],
+  content: [
+    './via/static/scripts/**/*.{js,ts,tsx}',
+    './node_modules/@hypothesis/frontend-shared/lib/**/*.js',
+  ],
+  theme: {
+    extend: {
+      fontFamily: {
```
robertknight, line 11: I copied the font family stack here from the LMS app, as I didn't see a default in `@hypothesis/frontend-shared/lib/tailwind.preset.js`. I'm not sure that we want this here though. @lyzadanger can you advise when you're back from 🇧🇪 ?

via/static/styles/video_player.css
```diff
@@ -0,0 +1,7 @@
+/*
+CSS entry point for the video player app.
+*/
+
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
robertknight, line 7: The entry point is CSS rather than SASS for now. We can change this as and when we need SASS features.

acelaya (approved):
👍🏼

PR #912, opened 2023-05-03
Use full YouTube URLs instead of just video IDs, and various other miscellaneous YouTube-related changes

marcospri:
Use full YouTube URLs instead of just video IDs as the path params for the `/video/` route.

Add a new `YouTubeService` that handles parsing YouTube URLs in order to extract the video IDs from them. In the future this same `YouTubeService` will also be used to get video transcripts from the YouTube API.

This commit also adds a new `YOUTUBE_CAPTIONS` setting and disables the `/video` route (HTTP Unauthorized) is this setting isn't enabled.

`get_url_details.py` is also turned into a service (`URLDetailsService`) because it now depends on two other services (and because previously it was a random top-level file).

Also, HTTP caching headers are added to the `/video` route.


via/services/youtube.py
```diff
@@ -0,0 +1,23 @@
+from urllib.parse import parse_qs, urlparse
+
+
+class YoutubeService:
+    @classmethod
+    def parse_url(cls, public_url):
+        """Get the youtube video ID from an URL."""
+        parsed = urlparse(public_url)
```
marcospri, line 15: This is a bit wordy compared to just a regexp but easier to get right and to read IMO.

via/views/__init__.py
```diff
@@ -9,7 +9,7 @@ def add_routes(config):  # pragma: no cover
     config.add_route("index", "/", factory=QueryURLResource)
     config.add_route("get_status", "/_status")
     config.add_route("view_pdf", "/pdf", factory=QueryURLResource)
-    config.add_route("view_video", "/video/{id}")
+    config.add_route("view_video", "/video", factory=QueryURLResource)
```
marcospri, line 12: This context support URLs in the URL

via/views/route_by_content.py
```diff
@@ -18,8 +18,9 @@ def route_by_content(context, request):
         request.find_service(HTTPService), url, request.headers
     )
     via_client_svc = request.find_service(ViaClientService)
+    content_type = via_client_svc.content_type(mime_type)
 
-    if via_client_svc.is_pdf(mime_type):
+    if content_type in ["pdf", "video"]:
```
marcospri, line 22: Same headers as for pdfs now.

via/views/view_video.py
```diff
@@ -1,15 +1,18 @@
...
 
+from via.services import YoutubeService
+
 
 @view_config(
     renderer="via:templates/video_player.html.jinja2",
     route_name="view_video",
 )
-def view_video(request):
+def view_video(context, request):
```
marcospri, line 9: We take full youtube URLs now instead of the video id and parse it here.

This is the same approach as for PDFs but there we split this further to support different PDF providers. As we only support youtube we don't need that specialization here.

If we were to add other video providers we'd keep the /video route taking the full URL as the main entry point.

via/services/youtube.py
marcospri, line 15: Once we integrate this into LMS we might have to extend this to support our own `video://youtube/ID` type of URL if we go that route.

For now this allows for easier testing directly in via.

robertknight:
The general approach of using existing Via URLs (`https://{via_url}/{http_url}`) but serving the response in the video player if the URL matches a recognized pattern, or the content type is a video, makes sense. On some sites we may run into an issue where users might want to annotate the content either as regular HTML or a video. We could show some kind of mode-switcher in the UI in that case, and ensure we record enough information with the annotations to know what mode was being used when a particular annotation was made.

tests/conftest.py
```diff
@@ -17,6 +17,7 @@ def pyramid_settings():
         "checkmate_ignore_reasons": None,
         "checkmate_allow_all": False,
         "enable_front_page": True,
+        "youtube_captions": True,
```
marcospri, line 20: New setting to enable/disable this

tests/unit/conftest.py
```diff
@@ -48,3 +49,13 @@ def pyramid_request(
     pyramid_request.checkmate.check_url.return_value = None
 
     return pyramid_request
+
+
+@pytest.fixture
+def call_view(pyramid_request):
```
marcospri, line 55: Moved to the common fixtures

tests/unit/via/services/url_details_test.py
```diff
@@ -36,7 +37,7 @@ def test_it_calls_get_for_normal_urls(
 
         url = "http://example.com"
 
-        result = get_url_details(http_service, url, headers=sentinel.headers)
+        result = svc.get_url_details(url, headers=sentinel.headers)
```
marcospri, line 40: Now a service, not a function

tests/unit/via/views/route_by_content_test.py
```diff
@@ -13,37 +13,39 @@ class TestRouteByContent:
     @pytest.mark.parametrize(
         "content_type,status_code,expected_cache_control_header",
         [
-            ("PDF", 200, "public, max-age=300, stale-while-revalidate=86400"),
-            ("HTML", 200, "public, max-age=60, stale-while-revalidate=86400"),
-            ("HTML", 401, "public, max-age=60, stale-while-revalidate=86400"),
-            ("HTML", 404, "public, max-age=60, stale-while-revalidate=86400"),
-            ("HTML", 500, "no-cache"),
-            ("HTML", 501, "no-cache"),
+            ("pdf", 200, "public, max-age=300, stale-while-revalidate=86400"),
```
marcospri, line 16: These were only compared in the test against "PDF" directly. Lowercase values the ones expected for content type.

via/services/url_details.py
```diff
@@ -0,0 +1,59 @@
...
+import cgi
+from collections import OrderedDict
+
+from via.requests_tools.headers import add_request_headers, clean_headers
+from via.services import HTTPService
+from via.services.google_drive import GoogleDriveAPI
+from via.services.youtube import YoutubeService
+
+
+class URLDetailsService:
```
marcospri, line 11: Made this into a service.

This was a function which already had a dependency on HTTPService and I was about to add one for YoutubeService.

Made into a service in one commit and extended with the YoutubeService dependency next.

via/services/via_client.py
```diff
@@ -5,19 +5,23 @@
...
     """A wrapper service for h_vialib.ViaClient."""
 
+    _mime_types_content_type = {
+        "application/x-pdf": "pdf",
+        "application/pdf": "pdf",
+        "video/x-youtube": "video",
+    }
+
+    def content_type(self, mime_type):
+        return self._mime_types_content_type.get(mime_type, "html")
```
marcospri, line 15: One function instead of `is_pdf`, `is_video`...

via/services/youtube.py
```diff
@@ -0,0 +1,30 @@
+from urllib.parse import parse_qs, urlparse
+
+
+class YoutubeService:
+    def __init__(self, enabled: bool):
+        self._enabled = enabled
+
+    @property
+    def enabled(self):
+        return self._enabled
```
marcospri, line 10: This is a bit silly but it will potentially grow with other attributes.

tests/unit/via/services/youtube_test.py
```diff
@@ -0,0 +1,33 @@
...
+        "url,video_id",
+        [
+            ("not_an_url", None),
+            ("https://notyoutube:1000000", None),
+            ("https://notyoutube.com", None),
+            ("https://youtube.com", None),
+            ("https://youtube.com?param=nope", None),
+            ("https://youtube.com?v=", None),
+            ("https://www.youtube.com/watch?v=VIDEO_ID", "VIDEO_ID"),
+            ("https://www.youtube.com/watch?v=VIDEO_ID&t=14s", "VIDEO_ID"),
```
marcospri, line 19: Support more formats see:

https://stackoverflow.com/a/8260383/15031259

And also not there:

https://www.youtube.com/live/4Lb2JoDN70o?feature=share
https://www.youtube.com/shorts/ktl1ncCTBGg

tox.ini
```diff
@@ -44,6 +44,7 @@ setenv =
     dev: VIA_SECRET = not_a_secret
     dev: CHECKMATE_API_KEY = dev_api_key
     dev: ENABLE_FRONT_PAGE = {env:ENABLE_FRONT_PAGE:true}
+    dev: YOUTUBE_CAPTIONS = {env:YOUTUBE_CAPTIONS:true}
```
seanh, line 29: Enable the feature flag by default in dev :+1:

via/services/url_details.py
seanh: Cool. Yeah this was a weird top-level file so I'm glad to see it gone. Turning it into a service seems right especially given the dependency on another service. Definitely the right decision when adding a _second_ dependency on a second other service

via/services/url_details.py
```diff
@@ -0,0 +1,59 @@
...
+        :raise UnhandledException: For all other request based errors
+        """
+        if headers is None:
+            headers = OrderedDict()
+
+        if GoogleDriveAPI.parse_file_url(url):
+            return "application/pdf", 200
+
+        if self._youtube.enabled and self._youtube.parse_url(url):
+            return "video/x-youtube", 200
```
seanh, line 34: As well as being moved into a service this now supports potentially returning `"video/x-youtube"` as a special case

via/services/youtube.py
```diff
@@ -0,0 +1,30 @@
+from urllib.parse import parse_qs, urlparse
+
+
+class YoutubeService:
```
seanh, line 4:
```suggestion
class YouTubeService:
```

via/services/youtube.py
seanh: :+1:

via/services/youtube.py
```diff
@@ -0,0 +1,30 @@
...
+            return None
+
+        if not parsed.query:
+            return None
+
+        query_params = parse_qs(parsed.query)
+        if "v" not in query_params:
+            return None
+
+        return query_params["v"][0]
```
seanh, line 26: I wonder if this should return a data dict for consistency with the `GoogleDriveAPI` service and for extensibility

via/views/debug.py
```diff
@@ -6,7 +6,7 @@
 from pyramid import view
 from pyramid.response import Response
 
-from via.get_url import clean_headers
+from via.requests_tools.headers import clean_headers
```
seanh, line 9: Weird that this was being imported from `get_url`

via/views/proxy.py
```diff
@@ -1,15 +1,16 @@
...
 
 @view_config(route_name="proxy", renderer="via:templates/proxy.html.jinja2")
 def proxy(context, request):
     url = context.url_from_path()
     request.checkmate.raise_if_blocked(url)
 
-    mime_type, _status_code = get_url_details(request.find_service(HTTPService), url)
+    mime_type, _status_code = request.find_service(URLDetailsService).get_url_details(
+        url
+    )
```
seanh, line 13: Just updating the call now that `URLDetailsService` is a service :+1:

via/views/route_by_content.py
```diff
@@ -14,12 +13,13 @@ def route_by_content(context, request):
 
     request.checkmate.raise_if_blocked(url)
 
-    mime_type, status_code = get_url_details(
-        request.find_service(HTTPService), url, request.headers
+    mime_type, status_code = request.find_service(URLDetailsService).get_url_details(
+        url, request.headers
```
seanh, line 17: Just updating the call now that `URLDetailsService` is a service :+1:

via/views/route_by_content.py
seanh: Ok, so we're going to use the same caching headers for videos as we've been using for PDFs :+1:

via/views/__init__.py
seanh: Ok, so we've changed from `/video/{id}` to `/video/{youtube_url}` and using `QueryURLResource` instead of normal route params so that `{youtube_url}` can be a full URL

via/views/view_video.py
seanh: Ok, there was an earlier PR (https://github.com/hypothesis/via/pull/905) that added this `/video/*` route, but I missed that.

So this is a new video player app that's part of Via now. But my understanding is that we're not actually proxying the bytes of YouTube videos through Via. Rather, Via just renders a page that includes the YouTube video player embedded, the video transcript, and the annotation client.

seanh (approved):
YouTube needs to be spelled with a capital T, e.g. `YouTubeService`.

Otherwise this all looks good to me :+1:

seanh:
Rebased and squashed...

tests/unit/via/services/youtube_test.py
seanh: I think I'm probably going to have LMS normalize the YouTube URLs into some custom format (something like: `video://youtube/{id}` before sending them to Via's `/video?url={url}` view. So the `/video` view won't need to support real YouTube URLs, so I think I'll probably end up removing this real YouTube URL parsing from Via (and putting it into LMS instead).

We would still need the real YouTube URL parsing in Via if we ever wanted to add YouTube support to public Via: the root `/<REAL_YOUTUBE_URL>` view would have to parse and normalize real YouTube URLs in the same way that LMS is going to and forward them to `/video?url={normalized_youtube_url}`.

Anyway, I think we *will* need this real YouTube URL parsing code at least in LMS even if we don't add YouTube support to public Via, so I might as well add support for more formats here now and it'll get cut-pasted into LMS later..

seanh:
I *think* I've parcelled everything in this PR out into separate smaller PRs now, so closing this one

PR #914, opened 2023-05-03
Fetch transcripts for youtube videos

marcospri:

PR #915, opened 2023-05-03
Mock transcript data in the same shape as the real one

marcospri:
We could try shape it into that other format in the backend if you prefer but this other format seems to have more information.

### DIFF
--- a/tests/unit/via/views/view_video_test.py
+++ b/tests/unit/via/views/view_video_test.py
@@ -19,12 +19,14 @@ class TestViewVideo:
         assert response["transcript"] == {
             "segments": [
                 {
-                    "time": 0,
                     "text": "First segment of transcript",
+                    "start": 0.12,
+                    "duration": 5.219,
                 },
                 {
-                    "time": 30,
                     "text": "Second segment of transcript",
+                    "start": 3.0,
+                    "duration": 3.48,
                 },
             ],
         }
--- a/via/views/view_video.py
+++ b/via/views/view_video.py
@@ -13,12 +13,14 @@ def view_video(request):
     transcript = {
         "segments": [
             {
-                "time": 0,
                 "text": "First segment of transcript",
+                "start": 0.12,
+                "duration": 5.219,
             },
             {
-                "time": 30,
                 "text": "Second segment of transcript",
+                "start": 3.0,
+                "duration": 3.48,
             },
         ],
     }


tests/unit/via/views/view_video_test.py
```diff
@@ -19,12 +19,14 @@ def test_it(self, pyramid_request, Configuration):
...
                 {
-                    "time": 0,
                     "text": "First segment of transcript",
+                    "start": 0.12,
+                    "duration": 5.219,
                 },
                 {
-                    "time": 30,
                     "text": "Second segment of transcript",
+                    "start": 3.0,
```
robertknight, line 28: I know this is just an example, but presumably this `start` value should be `>= 0.12 + 5.219`, or do you have real examples where the transcript segments can overlap?

marcospri: Those values are from a real video and every video I've tried (argubaly, just 5) has overlapping values.

I've been using https://github.com/hypothesis/via/pull/914 to fetch the subs.

marcospri: >start is the point in time when the subtitle box appears on screen in the youtube video and duration is how long it is displayed. As there often is some overlap between subtitle boxes, this can happen.

https://github.com/jdepoix/youtube-transcript-api/issues/21

robertknight: At present the frontend doesn't have a use for the duration data, it only needs  the `start` value. From https://github.com/jdepoix/youtube-transcript-api/issues/21 and https://github.com/jdepoix/youtube-transcript-api/issues/189 my understanding is that the duration value indicates how long a transcript line should remain on screen, in the captions overlaid on the video. The on-screen UI in YouTube displays the current and previous lines. For the transcript UI that information isn't needed though.

Perhaps the backend should simply omit it to avoid confusion. If we find a use case in future we can re-add it to the data.

Reducing the format to the minimal information will also be helpful if we need to integrate with other providers in future which don't have an equivalent field.

robertknight (commented):
Renaming `time` to `start` to match the YouTube "API" response makes sense. The meaning of the "duration" field seems to be tied into YouTube's particular UI if I understand correctly, and is not what readers might expect (the duration of the video segment in which the words in a segment are being read). I would be inclined to omit it from the backend response until we have a clear use case.

PR #917, opened 2023-05-04
Add logic to observe and control playback state in video player

robertknight:
Add logic in the `YouTubeVideoPlayer` component to sync playback state and position with the `play` and `time` props, and report playback state changes via the `onPlayingChanged` and `onTimeChanged` props. This allows the rest of the app to respond to changes in the playback state and control it. In this PR I added an unstyled button to play/pause the video and an input field that displays the playback time in seconds since the start of the video.

The `YouTubeVideoPlayer` component is YouTube-specific, but the intent is that video players for other sources could implement the same API, making it easy to swap out the player component.

The implementation makes use of the [YouTube Player API](https://developers.google.com/youtube/iframe_api_reference) via a script tag which is added to the page when the `YouTubeVideoPlayer` component mounts.

**Testing:**

1. On this branch, go to a video URL, eg. http://localhost:9083/video/xdkSEodpsC8 (nb. this will change soon to `http://localhost:9083/video/{normal_youtube_video_url}`)
2. Click the Play button in the page (outside the player) and observe that it can be used to play/pause the video. Playing/pausing the video using controls inside the video iframe should also update the state of this button.
3. Observe that when the video is playing, the time value below the play button updates once per second
4. Pause the video, edit the time value and click Play. This will cause the video to seek to the appropriate part of the video.

**Caveats:**

If you seek the video while paused, the time field doesn't update until you start playing again. This is on my TODO list for the future.


via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
```diff
@@ -0,0 +1,174 @@
...
+  onPlayingChanged,
+  onTimeChanged,
+  play = true,
+  videoId,
+  time,
+}: YouTubeVideoPlayerProps) {
+  const [loadError, setLoadError] = useState(null);
+  const encodedId = encodeURIComponent(videoId);
+  const encodedOrigin = encodeURIComponent(window.origin);
+  const playerURL = `https://www.youtube.com/embed/${encodedId}?enablejsapi=1&origin=${encodedOrigin}`;
```
acelaya, line 75: Could you add a comment with a link to the docs that explain how to build this URL?

acelaya (approved):
LGTM, and works as expected.

via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
robertknight: The docs are at https://developers.google.com/youtube/player_parameters#Manual_IFrame_Embeds. I will add a comment with this URL.

PR #918, opened 2023-05-05
Ensure `build/` dir exists before attempting to watch files in it

robertknight:
We encountered an issue where the `build/manifest.json` task was never generated if the `build/` dir did not exist prior to running `make dev`. It seems that `gulp.watch` fails if a watched dir does not exist. Create the dir explicitly first to resolve this.

This fix will need to be replicated in other repos that use h-assets, or we could move the file-watching code into a function in `@hypothesis/frontend-build` which handles this for us.

Slack thread: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1683278304439579

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -1,3 +1,5 @@
+import { mkdirSync } from 'fs';
+
 import {
   buildCSS,
   buildJS,
@@ -28,6 +30,8 @@ gulp.task('watch-css', () => {
 });
 
 gulp.task('watch-manifest', () => {
+  // Ensure build dir exists, otherwise `gulp.watch` may not work.
+  mkdirSync('build', { recursive: true });
   gulp.watch('build/**/*.{css,js,map}', generateManifest);
 });
 


gulpfile.mjs
```diff
@@ -28,6 +30,8 @@ gulp.task('watch-css', () => {
 });
 
 gulp.task('watch-manifest', () => {
+  // Ensure build dir exists, otherwise `gulp.watch` may not work.
+  mkdirSync('build');
```
acelaya: According to the docs, this function throws if the directory already exists, unless `recursive: true` is passed, so we probably want to pass that.

https://nodejs.org/api/fs.html#fsmkdirpath-options-callback

Another option would be to check if the directory exists before trying to create it:

```js
if (!existsSync('build')) {
  mkdirSync('build');
}
```

acelaya (changes requested):
See my comment about checking the directory does not exist

acelaya (approved):
👍🏼

gulpfile.mjs
robertknight: Oh, indeed. Using `{ recursive: true }` is better, as it avoids a potential race condition with any other processes/tasks that might try to create the directory at the same time.

PR #920, opened 2023-05-05
Improve transcript display and sync with current video timestamp

robertknight:
This PR starts work on getting the transcript to display alongside of the video in a scrollable container, per current designs, highlight and scroll to the segment matching the current timestamp, and navigate the video when clicking on timestamps in the transcript.

Please consider the visual design and UX details preliminary. The main focus is to get the transcript scrolling, syncing and timestamps in. Also the timestamps in the transcript appear in the annotation quotes / selected text at present. We'll probably want to remove those in future.

**Changes:**

 - Add dummy transcript data that will allow testing of transcript annotation, scrolling and syncing
 - Align the video, transcript and client in a three-column vertical layout, and make the transcript container scrollable.
 - Format timestamps in the transcript in `m:ss` or `h:mm:ss` strings. This format is intentionally more compact than what docdrop.org uses and matches youtube.com
 - Highlight the segment of the transcript that corresponds to the current video timestamp, and smoothly scroll it into view if needed.
 - Increase the contrast of hovered non-current transcript text segments, to make text selection easier.
 - Support navigating the video by clicking on timestamps in the transcript. We intentionally do not scroll when clicking the transcript text, as I felt that was disruptive to watching the video. This is something we might change.

**Testing:**

2. Click on the video to start playing. The playing/paused state should be reflected in the control bar above the transcript
3. Click on the Play/Pause button in the control bar. The video should play/pause.
4. As the video plays, the transcript segment corresponding to the current video time should be highlighted.
5. Move the video playback to a different point. The segment highlighting should update, _but only when the video starts playing from the new position_ (this is due to limitations of the player API).
6. Click on a timestamp in a segment. The video player should move to that timestamp.
7. As the video plays, the transcript container should scroll so that the current segment is centered in the container, unless there is a text selection in the container, to avoid disrupting the user while selecting text to annotate or copy.

**Preview:**

Screenshot 2023-05-15 at 11 08 03

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -66,7 +66,7 @@ function TranscriptSegment({
       data-is-current={isCurrent}
       data-testid="segment"
     >
-      <button className="pr-5 hover:underline" onClick={() => onSelect()}>
+      <button className="pr-5 hover:underline" onClick={onSelect}>
         {formatTimestamp(time)}
       </button>
       <p className="basis-64 grow" data-testid="transcript-text">
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -67,7 +67,7 @@ function TranscriptSegment({
       data-is-current={isCurrent}
       data-testid="segment"
     >
-      <button className="w-20 hover:underline" onClick={() => onSelect()}>
+      <button className="pr-5 hover:underline" onClick={() => onSelect()}>
         {formatTimestamp(time)}
       </button>
       <p className="basis-64 grow" data-testid="transcript-text">
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -1,3 +1,13 @@
+import {
+  Scroll,
+  ScrollContainer,
+  ScrollContent,
+} from '@hypothesis/frontend-shared';
+import classnames from 'classnames';
+import { useEffect, useRef } from 'preact/hooks';
+
+import { formatTimestamp } from '../utils/time';
+
 export type Segment = {
   /**
    * True if this segment corresponds to the section of the video that is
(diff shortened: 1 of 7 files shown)


via/static/scripts/video_player/utils/time.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Format a video timestamp.
+ *
+ * The timestamp is formatted as `m:ss` for times < 1 hour or `h:mm:ss` for
+ * times over 1 hour. This matches how transcript timestamps are formatted on
+ * youtube.com.
+ *
+ * @param time - Time in seconds since the start of the video
+ */
+export function formatTimestamp(time: number): string {
```
robertknight, line 14: This formatting is implemented manually for the moment without any consideration of the user's locale. I will need to check whether there are `Intl` methods we should be using instead.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -30,24 +30,33 @@ export default function VideoPlayerApp({
...
+          onTimeChanged={setTimestamp}
+        />
+      </div>
+      <div className="w-2/5 h-[80vh] bg-grey-0 border border-grey-3">
+        <div className="p-1 bg-grey-1 border-b border-grey-3">
+          <Button
+            classes="text-xl"
+            onClick={() => setPlaying(playing => !playing)}
+          >
+            {playing ? '⏸' : '⏵'}
```
robertknight, line 50: I'm using the unicode symbols for play/pause, but we probably still need to add ASCII labels for these buttons.

via/static/scripts/video_player/utils/time.ts
robertknight: There is a spec proposal at https://github.com/tc39/proposal-intl-duration-format. See https://chromestatus.com/feature/5193021205774336 for browser engine links. `Intl.DurationFormat` exists in Safari Tech Preview but not in Safari stable, Chrome or Firefox.

robertknight:
Text selection that spans multiple lines looks weird in Safari:

Multi-segment selection

The same selection looks fine in Chrome:

Chrome multi-line selection

In Firefox the selection background looks good, but the annotate toolbar is appearing in an unexpected place. My cursor was positioned at the bottom-right of the selection rect when I released it:

Firefox selection rect

robertknight:
I'm using a [ScrollContainer](https://patterns.hypothes.is/data-scrollbox) component with the `borderless` prop to create shadows at the top/bottom of the content to indicate scrollability. There is an issue at the moment with the surrounding border being clipped:

Scroll container

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -27,43 +37,139 @@ export type TranscriptProps = {
...
+    </li>
   );
 }
 
+/**
+ * Return the offset of `element` from the top of a positioned ancestor `parent`.
+ *
+ * @param parent - Positioned ancestor of `element`
+ */
+function offsetRelativeTo(element: HTMLElement, parent: HTMLElement): number {
```
robertknight, line 84: This function was copied from the Hypothesis client.

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -27,43 +37,141 @@ export type TranscriptProps = {
...
   return (
-    <div>
-      <div>{time}</div>
-      <p>{text}</p>
-    </div>
+    <li
+      className={classnames(
+        'flex flex-row p-1 hover:text-black',
+        isCurrent && 'bg-grey-2',
+        !isCurrent && 'text-grey-6'
```
acelaya: A matter of personal preference, but I find objects easier to read when conditionally setting classes with `classnames`.

```suggestion
        {
           'bg-grey-2': isCurrent,
           'text-grey-6': !isCurrent
        }
```

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -27,43 +37,141 @@ export type TranscriptProps = {
...
+      )}
+      data-is-current={isCurrent}
+      data-testid="segment"
+    >
+      <button className="pr-5 hover:underline" onClick={() => onSelect()}>
```
acelaya: Since the `onSelect` callback is not bind to an object, I think it should be fine to pass it as is, with no wrapping arrow function.

```suggestion
      <button className="pr-5 hover:underline" onClick={onSelect}>
```

acelaya (approved):
Looks great!

via/static/scripts/video_player/components/Transcript.tsx
robertknight: Hmm... in JavaScript, I would use the closure to explicitly discard the `Event` argument that get passed to `onClick`, and ensure upstream code doesn't depend on it. Now that we have TypeScript ensuring callers pass only a compatible callback, it is probably safe to skip the closure here.

via/static/scripts/video_player/components/Transcript.tsx
robertknight: Lyza's code in eg. the pattern library seems to prefer the object style too. I'll make this change.

PR #922, opened 2023-05-17
Add Sync button for scrolling transcript to current segment

robertknight:
This provides a way for the user to scroll the transcript back to the current segment, if they have manually scrolled it to a different location while the video is paused.

[Docdrop's video player](https://docdrop.org/video/G99TR_9KF4E/) and the mocks at https://hypothesis-video-transcriptions.netlify.app/ have the same toolbar button. A known issue in this PR is that the button doesn't have an icon like it does in these prototypes. That will come later.

**Testing:**

2. Scroll the transcript list manually until the current segment is no longer in view
3. Click the "Sync" button in the toolbar. The current segment should be scrolled back into view.

Sync button

### DIFF
--- a/via/static/scripts/test-util/mock-imported-components.ts
+++ b/via/static/scripts/test-util/mock-imported-components.ts
@@ -1,0 +1,80 @@
+import type { FunctionComponent } from 'preact';
+
+/**
+ * Return true if an imported `value` "looks like" a Preact function component.
+ *
+ * This check can have false positives (ie. match values which are not really components).
+ * That's OK because typical usage in a test is to first mock all components with
+ * `$imports.$mock(mockImportedComponents())` and then mock other things with
+ * `$imports.$mock(...)`. The more specific mocks will override the generic
+ * component mocks.
+ */
+function isComponent(value: unknown) {
+  return (
+    typeof value === 'function' &&
+    value.name.match(/^[A-Z]/) &&
+    // Check that function is not an ES class. Note this only works with real
+    // ES classes and may not work with ones transpiled to ES5.
+    !value.toString().match(/^class\b/)
+  );
+}
+
+/**
+ * Return the display name of a component, minus the names of any wrappers
+ * (eg. `withServices(OriginalName)` becomes `OriginalName`).
+ *
+ * @param component - A Preact component
+ */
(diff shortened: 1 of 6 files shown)


robertknight:
One of the tests failed with an error about deleting `window.YT`. I suspect this might relate to the VideoPlayerApp tests loading the real YouTube player component instead of mocking it out. I'm going to look into this to avoid test flakiness.

via/static/scripts/test-util/mock-imported-components.ts
```diff
@@ -0,0 +1,80 @@
+import type { FunctionComponent } from 'preact';
```
robertknight, line 1: This was copied from the lms repo, converted to TS and given some type refinements. In time I'd like to put these utilities in a shared package. See https://hypothes-is.slack.com/archives/C1M8NH76X/p1684247252232699.

via/static/scripts/video_player/utils/test/youtube-test.js
```diff
@@ -10,7 +10,7 @@ describe('loadYouTubeIFrameAPI', () => {
       ready: callback => setTimeout(() => callback(), 0),
       Player: sinon.stub(),
     };
-    return window.PT;
+    return window.YT;
```
robertknight, line 13: This is a fix for a typo.

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -145,7 +173,19 @@ export default function Transcript({
...
+    scrollToCurrentSegment();
+  }, [currentIndex, transcript, scrollToCurrentSegment]);
+
+  useImperativeHandle(
+    controlsRef || { current: null },
+    () => ({
+      scrollToCurrentSegment,
+    }),
+    [scrollToCurrentSegment]
+  );
```
acelaya, line 188: First time I see this hook out in the wild 😄

acelaya (approved):
👍🏼

via/static/scripts/video_player/components/Transcript.tsx
robertknight: Heh. I think I've seen this used in exactly one Hypothesis frontend PR before, and it looks like that PR either didn't get merged or got replaced subsequently.

PR #923, opened 2023-05-17
Ensure border on scrolling transcript doesn't get cut off

lyzadanger:
Contain element with height of 80vh with a flex-column layout. This will prevent the bottom of the border on the transcript from disappearing.

Oh! It does occur to me to ask what the purpose is of constraining that element to 80vh, but we can tackle that separately!

Before:

image

After:

image

Also, just for style elimination: default border color should be `grey-3` without having to set; I removed the border-color styles there because they were right in front of me :).

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -41,8 +41,8 @@ export default function VideoPlayerApp({
           onTimeChanged={setTimestamp}
         />
       </div>
-      <div className="w-2/5 h-[80vh] bg-grey-0 border border-grey-3">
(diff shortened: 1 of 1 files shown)


robertknight:
> It does occur to me to ask what the purpose is of constraining that element to 80vh, but we can tackle that separately!

I think it makes sense to size the transcript to some fraction of the viewport size, maybe with min/max limits. This balances ensuring the whole scrollable container fits in the viewport, and taking advantage of however tall the screen is. The 80% value is an arbitrary choice.

robertknight (approved):
LGTM. Following this fix I increased the transcript height from 80vh to 90vh so that overall it is roughly the same height as before.

PR #924, opened 2023-05-18
Use undocumented `onVideoProgress` event to detect video progress changes

robertknight:
Instead of polling the player every second to get the current timestamp, use the undocumented `onVideoProgress` event that the player emits as it plays. This event has the advantage of being emitted immediately when the user manually seeks the video, whereas the previous approach only synced the transcript once the video had played for a second after beeing seeked.

I [filed an issue with YouTube](https://issuetracker.google.com/issues/283097094) to confirm that this API is something we can continue to rely on. If it ever gets removed, we'll have to go back to polling. There are some major sites (eg. [Khan Academy](https://www.khanacademy.org/math/linear-algebra/matrix-transformations/linear-transformations/v/a-more-formal-understanding-of-functions)) that use YouTube's player and need the same functionality, so I don't think this is too risky.

**Testing:**

2. Start playing the video
3. Pause the video and seek to a different location using the progress bar inside the player

On main, the transcript will not update to reflect the new position until after the video starts playing again. On this branch it will update immediately.

### DIFF
--- a/via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
+++ b/via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
@@ -92,6 +92,8 @@ export default function YouTubeVideoPlayer({
 
   const playerController = useRef<YT.Player>();
 
+  // Capture last-seen values of props so we can invoke callbacks without
+  // re-running effects when they change.
   const onPlayingChangedCallback = useRef<(playing: boolean) => void>();
   onPlayingChangedCallback.current = onPlayingChanged;
 
--- a/via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
+++ b/via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
@@ -56,6 +56,22 @@ function isPlaying(state: YT.PlayerState) {
 }
 
 /**
+ * Additional events that the YouTube player supports, which are missing from
+ * {@link YT.Events}.
+ */
+interface PlayerEvents extends YT.Events {
+  /**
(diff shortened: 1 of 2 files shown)


robertknight:
Update: The issue at https://issuetracker.google.com/issues/283097094 has a response from someone with an @google.com account, but it looks like the response is coming from someone who doesn't know the real answer. I will try some other channels, but in the meantime since this change is easy to revert if needed, I'd like to continue using this undocumented API.

via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
```diff
@@ -79,6 +95,9 @@ export default function YouTubeVideoPlayer({
   const onPlayingChangedCallback = useRef<(playing: boolean) => void>();
   onPlayingChangedCallback.current = onPlayingChanged;
 
+  const onTimeChangedCallback = useRef<(timestamp: number) => void>();
+  onTimeChangedCallback.current = onTimeChanged;
```
acelaya, line 101: What's the point of "wrapping" the callback in a ref? 🤔

acelaya (approved):
👍🏼

via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
robertknight: This is to avoid re-running the effect unnecessarily if the callback changes. Instead we just want to run the effect once and then use whatever the last value was.

robertknight: I added a comment to explain this for future readers.

PR #925, opened 2023-05-18
Replace Lorem Ipsum transcript with a sample from a real video

robertknight:
Replace the Lorem Ipsum text with a sample transcript from a real video, https://www.youtube.com/watch?v=x8TO-nrUtSI, generated using the steps at https://github.com/hypothesis/via/pull/914#issuecomment-1547841471. This will allow us to test search and work through other UX issues while the backend transcript fetching/caching etc. is being worked on.

In the process the transcript types were moved out of the `Transcript.tsx` component and into a separate non-UI module where they can be consumed from non-UI code without introducing a utilties => UI dependency. The `time` field of `Segment` was also renamed to `start` to match the structure of the data in the API we are using. The sample data also has a `duration` field, but this is not used by the frontend, so isn't included in the `Segment` type.

**Testing:**

2. You should see the video alongside the correct transcript


acelaya (approved):
👍🏼

PR #929, opened 2023-05-18
Add initial version of search bar to video player

robertknight:
Add an initial version of a search bar to the video player. This filters the transcript to segments whose text matches the current filter. Additionally the current segment is always displayed, so the user can more easily tell where the current video place is relative to the video segments that are visible.

If the browser supports the [CSS Custom Highlights API](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Custom_Highlight_API) (currently only Chrome), that is used to underline matching spans in segments in a way that doesn't interfere with highlights inserted by the Hypothesis client. In other browsers the matching segments are displayed with no underlines. There is a known issue that these highlights may not display in segments that also have Hypothesis highlights.

Search bar preview

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -9,11 +9,13 @@ import {
   useCallback,
   useEffect,
   useImperativeHandle,
+  useMemo,
   useRef,
 } from 'preact/hooks';
 
 import { formatTimestamp } from '../utils/time';
-import type { Segment, TranscriptData } from '../utils/transcript';
+import type { MatchOffset, Segment, TranscriptData } from '../utils/transcript';
+import { filterTranscript } from '../utils/transcript';
 
 /**
  * Interface for interacting with the transcript view, beyond what is available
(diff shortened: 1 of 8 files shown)


via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -45,38 +47,95 @@ export type TranscriptProps = {
...
+
+    const ranges = matches.map(({ start, end }) => {
+      const textNode = contentRef.current!.childNodes[0];
+      const range = new Range();
+
+      // FIXME - If the Hypothesis client has inserted highlights, these will
+      // break up the single text node child into multiple children.
+      if (textNode instanceof Text && textNode.length >= end) {
+        range.setStart(contentRef.current!.childNodes[0], start);
+        range.setEnd(contentRef.current!.childNodes[0], end);
```
acelaya: If `textNode` is created from `contentRef.current!.childNodes[0]`, and there's already a check making sure it is a `Text` instance, could we use it in the `setStart` and `setEnd` calls?

```suggestion
        range.setStart(textNode, start);
        range.setEnd(textNode, end);
```

acelaya (approved):
Looks good. Just left a small comment

via/static/scripts/video_player/components/Transcript.tsx
robertknight: Oh, indeed. I forgot to do this after extracting `contentRef.current!.childNodes[0]` into a variable.

PR #931, opened 2023-05-19
Exclude timestamps from selected text in transcripts

robertknight:
This PR solves the problem of segment timestamps appearing in the middle of annotation quotes. The Hypothesis client creates quotes from the text content of the selected range (effectively `window.getSelection().getRangeAt(0).toString()`). Since timestamp "buttons" appear at the start of each segment in DOM layout order, this resulted in the timestamps appearing in the quotes, as well as the user-visible selection.

The approach taken here is to render formatted timestamps using a `::before` pseudo-element rather than the text content of timestamp `<button>`s. This way the timestamp text doesn't appear in the selected range's text, and hence doesn't get a selection background color or appear in annotation quotes or highlights.

An alternative approach would be to introduce some general mechanism in the client for excluding certain elements from the selection. For example maybe we could exclude all elements with a `user-select: none` style from the quote, or exclude elements with a certain `data-` attribute. We'd have to be careful about the impact of this on the wider web though.

**Testing:**

2. Make a selection that spans multiple lines of the transcript. The selection background should not include the timestamps.
3. Annotate the selected text. The timestamps at the start of each segment should not appear in the annotation quote in the sidebar or be highlighted.

Exclude timestamps from selection

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -118,6 +118,8 @@ function TranscriptSegment({
     };
   }, [highlight, matches]);
 
+  const timestamp = formatTimestamp(time);
+
   return (
     <li
       className={classnames('flex flex-row p-1 hover:text-black', {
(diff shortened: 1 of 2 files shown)


robertknight:
>  I also need to check the accessibility impact of this approach.

I tested this with VoiceOver. Since we're setting an `aria-label` for the button, that gets used as expected.

acelaya:
> maybe we could exclude all elements with a `user-select: none` style from the quote

I think this would make sense in general. The problem is that we could break existing annotations that have already been created including those.

acelaya (approved):
LGTM, and works as expected

PR #932, opened 2023-05-22
Clear transcript filter when scrolling a highlight into view

robertknight:
**Depends on https://github.com/hypothesis/client/pull/5484**

The client's normal logic for scrolling a highlight into view does not work if the transcript segment is hidden due to a search filter being active. Listen for the `scrolltorange` event that the client emits before it scrolls a highlight into view and respond by clearing the filter.

We currently also then have to scroll the highlight into view ourselves, rather than let the client do it, because the client can currently only scroll to highlights that are visible immediately after the `scrolltorange` event is dispatched.

**Testing:**

Make sure you have https://github.com/hypothesis/client/pull/5484 checked out.

2. Annotate some text in the middle of the transcript
3. Enter a filter query in the search bar which does not match the text annotated in step (2)
4. Click the annotation card in the sidebar for the annotation created in step 2

In step 4, clicking the annotation card should cause the filter entered in step 3 to be cleared and the transcript should be scrolled to the location of the highlight.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -1,5 +1,5 @@
 import { Button, Input } from '@hypothesis/frontend-shared';
-import { useRef, useState } from 'preact/hooks';
+import { useEffect, useMemo, useRef, useState } from 'preact/hooks';
 
 import type { TranscriptData } from '../utils/transcript';
 import HypothesisClient from './HypothesisClient';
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -61,19 +105,18 @@ export default function VideoPlayerApp({
...
           <Input
-            aria-label="Filter transcript"
+            aria-label="Transcript filter"
             data-testid="filter-input"
-            onInput={e =>
-              setFilter((e.target as HTMLInputElement).value.trim())
-            }
+            onInput={e => setFilter((e.target as HTMLInputElement).value)}
             placeholder="Search..."
+            value={filter}
```
robertknight, line 120: This ensures that clearing the filter state internally also updates the input control.

acelaya (approved):
LGTM

PR #933, opened 2023-05-22
Set play/pause button label explicitly

robertknight:
This ensures assistive technology can present a proper description for the button. The Unicode symbols that are currently used as "icons" for this button are read as eg. "right-pointing arrow" rather than "play".

Play-pause buttons

### DIFF
--- a/via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
+++ b/via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
@@ -54,14 +54,18 @@ describe('VideoPlayerApp', () => {
     let player = wrapper.find('YouTubeVideoPlayer');
     assert.isFalse(player.prop('play'));
 
-    const playButton = wrapper.find('button[data-testid="play-button"]');
+    let playButton = wrapper.find('button[data-testid="play-button"]');
     assert.equal(playButton.text(), '⏵');
+    assert.equal(playButton.prop('title'), 'Play');
 
     playButton.simulate('click');
 
     player = wrapper.find('YouTubeVideoPlayer');
     assert.isTrue(player.prop('play'));
+
+    playButton = wrapper.find('button[data-testid="play-button"]');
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
```diff
@@ -54,14 +54,18 @@ describe('VideoPlayerApp', () => {
...
+    let playButton = wrapper.find('button[data-testid="play-button"]');
     assert.equal(playButton.text(), '⏵');
+    assert.equal(playButton.prop('title'), 'Play');
 
     playButton.simulate('click');
 
     player = wrapper.find('YouTubeVideoPlayer');
     assert.isTrue(player.prop('play'));
+
+    playButton = wrapper.find('button[data-testid="play-button"]');
```
robertknight, line 66: It is necessary to re-find the play button here to ensure we get an Enzyme wrapper with a current view of the props. Some methods (eg. `text()`) might work without this because they operate on the underlying DOM node, but the `prop` method doesn't.

lyzadanger: That's just fiddly enough that I wonder if you should add a comment to the test?

lyzadanger (approved):
With the caveat that the Unicode character for play (`U+23F5`) isn't displaying on my Mac (Monterey/12.5) in any browser I've tried (pause works fine), this seems fine. We'll be replacing these glyphs very shortly with SVG icon components.

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
robertknight: I think I'd find myself repeating this in a lot of places. For now this is just a general quirk that the team will need to get familiar with. Maybe with some future changes in Enzyme a sub-wrapper could have a method to remember how it was obtained and refresh itself.

PR #936, opened 2023-05-23
Work around Firefox issue that prevented annotating across segments

### LINKED ISSUE #930: Annotating text that spans multiple transcript segments doesn't work in Firefox
**Steps to reproduce:**

1. Annotate a video in Via using Firefox (current URL https://via.hypothes.is/video/x8TO-nrUtSI, this format will change in future)
2. Make a selection that spans multiple segments of the transcript (ie. sections with separate timestamps)
3. Click Annotate

**Expected result:** New annotation is created. Annotation quote includes the whole selection.
**Actual result:**

- In Firefox the new annotation's quote and highlight only includes part of the selection
- In Safari and Chrome the whole selection is included in the annotation's quote and highlight

**Notes:**

After making the selection, the adder toolbar is displayed in the wrong location, which is very likely to be related:

Firefox adder

In the above screenshot, the toolbar should be shown below the bottom of the selection, instead of below the first line.

robertknight:
The video player uses `<button>` elements for timestamps, and these have a default `user-select: none` style. This causes selections spanning multiple segments, which include these buttons, to be split into multiple ranges in Firefox. This FF behavior is not compliant with the Selection API spec, and the Hypothesis client currently doesn't handle it.

If a general solution is implemented in the client in future, we can remove this workaround.

Fixes https://github.com/hypothesis/via/issues/930

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -128,7 +128,22 @@ function TranscriptSegment({
       data-is-current={isCurrent}
       data-testid="segment"
     >
-      <button className="pr-5 hover:underline" onClick={onSelect}>
+      <button
+        className={classnames(
+          'pr-5 hover:underline',
+
+          // Workaround for a Firefox issue that prevented annotating across
+          // multiple segments [1]. Buttons have a default `user-select: none`
+          // style in FF and selections that include elements with this style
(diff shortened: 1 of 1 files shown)


acelaya (approved):
👍🏼

robertknight:
I have a general solution for the client in the works at https://github.com/hypothesis/client/pull/5487, but I will merge this PR and can easily remove it later when no longer needed.

PR #942, opened 2023-05-25
Add "Auto-scroll" checkbox to toggle automatic transcript scrolling

robertknight:
Add a checkbox to control whether the transcript automatically scrolls as the video plays. Turning this off is useful if the user wants to read ahead or back in the transcript while the video is playing.

The focus in this PR is adding the functionality. The location and styling is preliminary.

auto-scroll checkbox

**Testing:**

2. Auto-scroll should initially be enabled. The transcript should scroll as the video plays.
3. Toggling the checkbox off should prevent the auto-scroll behavior and allow the user to scroll up and down while the video is playing
4. Toggling the checkbox back on should re-enable auto-scrolling and immediately scroll to the current location

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -30,6 +30,12 @@ export type TranscriptControls = {
 };
 
 export type TranscriptProps = {
+  /**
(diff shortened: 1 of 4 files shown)


acelaya (approved):
👍🏼

Issue #939, opened 2023-05-25
Add keyboard shortcuts for video player


robertknight:
Implement keyboard shortcuts for basic functionality in the viewer. I suggest we use the [YouTube shortcuts](https://support.google.com/youtube/answer/7631406?hl=en-GB) as a reference.

I don't think we need to implement all the shortcuts that YouTube has, but a few basic ones that would be useful:

- Play/pause the video (Space/k in YT)
- Mute/unmute ("m" in YT)
- Focus the transcript search bar ("/" in YT focuses the general search bar)

Also some that our custom to our app:

- Sync the transcript location with the video
- Turn automatic scrolling of transcript on/off

robertknight:
A basic set of shortcuts was added in https://github.com/hypothesis/via/pull/943. We can add more as we think of them.

PR #943, opened 2023-05-25
Add some basic keyboard shortcuts for video player app

robertknight:
Add a few basic keyboard shortcuts for common actions.

 - "k" plays/pauses the video
 - "/" focuses the search field. Pressing Esc afterwards will remove focus, allowing the use of other single-letter shortcuts again.
 - "a" toggles auto-scroll
 - "s" syncs the transcript

Shortcuts are chosen to match YouTube where feasible. YouTube also supports space to toggle playback, but that is trickier because it must be ignored when an interactive element (eg. a button) is focused, so is not supported.

Note that our own code to handle shortcuts will not work when the iframe is focused, although the YouTube player does support some of the same shortcuts itself (eg. "k").

Part of https://github.com/hypothesis/via/issues/939.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -1,5 +1,11 @@
 import { Button, Checkbox, Input } from '@hypothesis/frontend-shared';
-import { useEffect, useMemo, useRef, useState } from 'preact/hooks';
+import {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'preact/hooks';
 
 import type { TranscriptData } from '../utils/transcript';
 import HypothesisClient from './HypothesisClient';
@@ -54,6 +60,7 @@ export default function VideoPlayerApp({
 
   const [filter, setFilter] = useState('');
   const trimmedFilter = useMemo(() => filter.trim(), [filter]);
+  const filterInputRef = useRef<HTMLInputElement>();
 
   // Listen for the event the Hypothesis client dispatches before it scrolls
   // a highlight into view. If a filter is currently active, clear it first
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -90,6 +97,45 @@ export default function VideoPlayerApp({
...
+      // Shortcuts that match the YouTube player, or perform a similar action.
+      // See https://support.google.com/youtube/answer/7631406.
+      if (matchesKey('k')) {
+        setPlaying(playing => !playing);
+      } else if (matchesKey('/')) {
+        const input = filterInputRef.current;
+        if (input) {
+          input.focus();
+          input.select();
+        }
```
lyzadanger, line 123: There's no chance the filter input would ever be `disabled` at any point, correct?

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -111,16 +157,20 @@ export default function VideoPlayerApp({
...
           <div className="flex-grow" />
           <Input
             aria-label="Transcript filter"
             data-testid="filter-input"
+            elementRef={filterInputRef}
+            onKeyUp={e => {
+              // Allow user to easily remove focus from search input.
+              if (e.key === 'Escape') {
+                (e.target as HTMLElement).blur();
+              }
```
lyzadanger, line 172: Have you considered attempting to restore focus to the previously-focused element? Can you help me understand the aim of the blur behavior here? Have you seen this behavior (i.e. blurring only) elsewhere, like on YouTube?

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
```diff
@@ -242,4 +242,96 @@ describe('VideoPlayerApp', () => {
...
+        { attachTo: document.body }
+      );
+      wrappers.push(wrapper);
+      return wrapper;
+    }
+
+    function sendKey(wrapper, key, target = document.body) {
+      act(() => {
+        target.dispatchEvent(
+          new KeyboardEvent('keyup', { key, bubbles: true })
```
lyzadanger, line 274: Can you help me understand why you had to explicitly set bubbling here in this test?

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -90,6 +97,45 @@ export default function VideoPlayerApp({
...
 
+  const syncTranscript = useCallback(
+    () => transcriptControls.current!.scrollToCurrentSegment(),
+    []
+  );
+
+  // Handle app-wide keyboard shortcuts.
+  useEffect(() => {
+    const keyListener = (e: KeyboardEvent) => {
+      // Disable single letter shortcuts if focused element is a text input.
```
lyzadanger, line 108: Ah, wise. Makes me think of perhaps adding an option for this to the `useKeyPress` hook in `frontend-shared`...

lyzadanger (approved):
These keyboard shortcuts work as described!

With the context of "I'm sure you did your homework and looked at how YouTube and others are doing it," I had a question about the combination of _sync_ and _auto-scroll_.

If, as a user, I had disabled both auto-scroll and sync, when I re-enabled sync, I might expect auto-scroll to re-enable automatically. Have you seen examples of how these two controls behave with respect to each other elsewhere?

I also had a question about the search-input blurring (comments inline).

via/static/scripts/video_player/components/VideoPlayerApp.tsx
robertknight: Not at present. If the transcript was loaded asynchronously (it isn't at present) then perhaps we should disable it while loading?

via/static/scripts/video_player/components/VideoPlayerApp.tsx
robertknight: YouTube doesn't appear to restore focus to the previous element, although that does seem like it would be an improvement and in-line with how eg. dialogs should ideally work. The main goal here though is to allow the user to return to a context where they can re-use single-letter shortcuts (eg. "k" to toggle play/pause) which can't be used while the search input has focus. For that purpose re-focusing the previous element is not required.

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
robertknight: If a key event is dispatched at the search input in the real app, it will bubble up. However the video player app might need to ignore it (eg. if the user pressed "k" while the search input is focused, that should not trigger the play/pause shortcut).

robertknight:
> If, as a user, I had disabled both auto-scroll and sync, when I re-enabled sync, I might expect auto-scroll to re-enable automatically. 

"Sync" is a one-time _action_ which scrolls the transcript to the position where the video is playing. "Auto-scroll" is a boolean _state_ that causes the transcript to be synced when the video's current time changes. The Sync action is generally not useful when auto-scroll is active and the video is actively playing. We could choose to disable it in that case, although it doesn't do any harm either.

PR #944, opened 2023-05-30
Add icons for play/pause/sync actions

robertknight:
Replace the placeholder emoji "icons" for play/pause actions with proper SVG icons, and add an icon for the sync button.

The SVG files were taken from the UI prototype at https://hypothesis-video-transcriptions.netlify.app and converted to Preact components using the `generate-icons.js` script in the hypothesis/frontend-shared repo (see also https://github.com/hypothesis/frontend-shared/pull/1064).

The layout of the top bar has also been modified to more closely match the prototype. I'm not sure that we'll want to have the search bar always be visible and whether it should be above/below the buttons, but it is useful for it to have its own row where it constrained for horizontal space by the buttons around it.

**Before:**

(nb. Appearance of play/pause button varied by platform)

Video player before

**After (Paused state):**

Video player - Paused state

**After (Playing state):**

Video player - Playing state

PR #945, opened 2023-05-30
Add button to copy transcript in video player

robertknight:
~~**Depends on https://github.com/hypothesis/via/pull/944**~~

Add a button to the video player toolbar to copy the video's transcript to the clipboard. This is based on the design in https://hypothesis-video-transcriptions.netlify.app and similar functionality in Docdrop's video player.

The implementation uses the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#browser_compatibility). Support for [_writing_](https://caniuse.com/async-clipboard) using this API is now supported widely enough that I think we can rely on it. Support for _reading_ is more limited according to caniuse.

Error handling is rather rudimentary. If the browser decides to reject the clipboard write for any reason, we just log a console warning. Ideally we'd show a toast message or something else instead so the user is aware the write failed.

**Preview:**

Copy transcript button


acelaya (approved):
👍🏼

PR #948, opened 2023-05-31
Pause video playback when Hypothesis client scrolls to highlight

### LINKED ISSUE #947: Handle scrolling to an annotation while video is playing (esp. if auto-scroll is enabled)
In the video player we need to decide what to do if the user clicks on an annotation in the sidebar while the video is playing. In particular scrolling the transcript to the location corresponding to the annotation conflicts with auto-scroll.

The current behavior is that clicking an annotation card starts scrolling the transcript to the location of the annotation, but leaves the video playing at its current position. If auto-scroll is enabled and the video is playing, then as soon as the video advances enough to change the current segment (usually that only takes a few seconds) the video player will try to scroll back to the location where it is playing.

Some ways that we could potentially handle this:

1. Pause the video before scrolling to a highlight
2. Disable auto-scroll before scrolling to a highlight
3. Change the current timestamp of the video to the segment corresponding to the highlight
4. Some combination of the above

robertknight:
This prevents the video player from immediately scrolling away from the highlight back to wherever it is currently playing.

Fixes https://github.com/hypothesis/via/issues/947

**Testing:**

1. Go to a video (eg. http://localhost:9083/video/x8TO-nrUtSI)
2. Create an annotation somewhere in the transcript that is not near the start
3. Start playing the video. It will soon scroll back up to where the transcript is currently playing
4. Click the annotation card in the sidebar

At step (4), playback should pause and the transcript should scroll to the highlight.

The same behavior ought to happen if you click the up/down-pointing icons at the top/bottom of the bucket bar, but this is not happening due to a [client bug](https://github.com/hypothesis/client/issues/5505).

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -65,20 +65,22 @@ export default function VideoPlayerApp({
   const filterInputRef = useRef<HTMLInputElement>();
 
   // Listen for the event the Hypothesis client dispatches before it scrolls
-  // a highlight into view. If a filter is currently active, clear it first
-  // to ensure the highlight is visible.
+  // a highlight into view.
+  //
+  // - If a filter is currently active, clear it first to ensure the highlight
+  //   is visible.
+  // - Pause playback to prevent transcript quickly scrolling away from the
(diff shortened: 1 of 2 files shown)


acelaya (approved):
Works as expected 👍🏼

PR #949, opened 2023-05-31
Sync transcript immediately after playback starts

robertknight:
When autoscroll is active, sync the transcript immediately after the video starts playing. Previously the transcript did not sync until playback reached the start of the next segment after the current one.

**Testing:**

1. Go to a video (eg. http://localhost:9083/video/x8TO-nrUtSI)
2. Scroll the transcript down
3. Start playing the video by any of the available means (clicking in the player, pressing the "Play" button, pressing the "k" shortcut)

At step 3, the transcript should immediately scroll to the current location. On `main` it won't scroll few a few seconds until playback reaches the next segment.

If the "auto-scroll" checkbox is unchecked, this scroll-on-play behavior shouldn't happen.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -104,6 +104,15 @@ export default function VideoPlayerApp({
     []
   );
 
+  // If we transition from paused to playing while autoscroll is active,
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
Good work. Once I got my head around the `useImperativeHandle` hook involved in the tests, this was straightforward enough.

PR #950, opened 2023-06-01
Fix scrolling to annotation, in video player, when there is no filter

robertknight:
Fix an issue where clicking an annotation card did not scroll the transcript if
there was no filter.

This happened because the player always instructed Hypothesis to wait until the
next render before scrolling. If there was no filter and the video was paused,
no re-render happened and so Hypothesis would wait indefinitely until some other
user action caused a re-render. The fix is to only make the client wait if
there is a filter to clear.

- The first commit extracts the logic for waiting for the next render into a shared `useNextRender` hook, and also fixes an issue where it could fail to resolve if called multiple times before the next render happened
- The second commit changes `VideoPlayerApp` to use this hook, and avoiding waiting if there is no filter

**Testing:**

1. Go to a video, eg. http://localhost:9083/video/x8TO-nrUtSI
2. Create an annotation
3. Scroll the transcript so the highlight is off-screen
4. Click on the annotation card. On `main` the transcript won't scroll unless there is a filter. On this branch it should always scroll.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -7,6 +7,7 @@ import {
   useState,
 } from 'preact/hooks';
 
+import { useNextRender } from '../utils/next-render';
 import type { TranscriptData } from '../utils/transcript';
 import { formatTranscript } from '../utils/transcript';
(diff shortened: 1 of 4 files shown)


acelaya (approved):
Looks good, and works as expected.

PR #951, opened 2023-06-01
Apply the cookiecutter to Via

seanh:
Apply our pyramid-app cookiecutter to Via. Depends on https://github.com/hypothesis/cookiecutters/pull/121

These cookiecutter PRs are by their nature kind of hard to review by looking at the diff. Some recommendations on how to test it:

* Take a look at [the new `ci.yml` passing on this branch](https://github.com/hypothesis/via/actions/workflows/ci.yml?query=branch%3Aapply-cookiecutter)
* Run `make help` to see the available commands and try them out
* Try `make clean sure` on your machine


jon-betts (commented):
Similar to the LMS change, this seems to have recompiled the dependencies.

This was a black box test of the commands, not a view of the code. I ran through some quick shake down type tests:

* `make clean`
* `make sure` (ran twice)
* `make docker`
* `make docker-run` - Errors, couldn't exit cleanly with Ctrl-C had to kill with docker
* `make dev`

seanh:
@jon-betts Could you re-review?

1. `make template` should be clean now that https://github.com/hypothesis/cookiecutters/pull/121 is merged

2. The changes to the requirements files in this PR are deliberate and look correct to me

3. The problem with <kbd>Ctrl</kbd> + <kbd>c</kbd> not killing `make docker-run` also exists on `main` so I'm not going to look into that for this PR

4. I've fixed a couple of problems with `make docker-run` so that the processes do now start up successfully:

   1. On `main` there are [a bunch of envvars in the `docker run` command in the `Makefile`](https://github.com/hypothesis/via/blob/ddfc763e1b9ec8b9d81e0bf41d2437210616b272/Makefile#L145-L152). On this branch / with the cookiecutter these needed to be [added to `.docker.env` file](https://github.com/hypothesis/via/pull/951/files#diff-1d0df5054f8c449c8fe67e2e27c2d76815e1e68543a28b71fed4590b5e094f77)
   2. On `main` Via also [mounts a `via-data` directory](https://github.com/hypothesis/via/blob/ddfc763e1b9ec8b9d81e0bf41d2437210616b272/Makefile#LL144C3-L144C38) which is unique to Via. Added this back using a cookiecutter include

Makefile
```diff
@@ -104,50 +97,42 @@ requirements/lint.txt: requirements/tests.txt requirements/functests.txt
...
+	@git archive --format=tar HEAD | docker build -t hypothesis/via:$(DOCKER_TAG) -
+
+.PHONY: docker-run
+$(call help,make docker-run,"run the app's docker image")
+docker-run:
+	@docker run \
+		--add-host host.docker.internal:host-gateway \
+		--env-file .docker.env \
+		--env-file .devdata.env \
+		-p 9082:9082 \
```
jon-betts: I think some important stuff has been lost here. The main port has changed form 9083 to 9082, so if you go to it you get no content. There might be some other missing bits too.

jon-betts (commented):
The `make template` now seems clean.

The `make docker-run` now executes, but you can't get any results back from it. There are quite a lot of changes in that section, but one that jumps out is the change of port from 9082 to 9083. I think we are trying to point to the HTTP end-point of the app instead of NGINX, but in the docker container there is no such end-point as the app exposes itself via a unix socket.

Makefile
seanh: I think the port thing is the only change: it's the only difference I see and if I fix that one thing locally then it does work. But it's kind of a pain to fix properly: all our other projects use their Gunicorn port here in the Docker command except Via which uses a different port. I'll have to tweak the cookiecutter to allow for this. I wonder if it might be worth moving this whole command out into a `bin/make_docker_run` file that `Makefile` calls so that projects can take over this entire file for themselves if they want to

seanh: Cookiecutter PR to enable fixing this: https://github.com/hypothesis/cookiecutters/pull/124

seanh: ^ Ok, that's merged and I've applied the change to this PR: `make docker-run` works now

seanh:
I'm gonna take the liberty of merging this because it's blocking work on the video annotations feature. The broken `make docker-run` was the only remaining issue, we discussed the solution to that in https://github.com/hypothesis/cookiecutters/pull/124 and I've just applied it to this PR

PR #964, opened 2023-06-06
Fix stopping Via's NGINX process in dev

seanh:
Problem: run `make dev` then kill it with <kbd>Ctrl</kbd> + <kbd>c</kbd>
then run `make dev` again. You'll see this error from the `nginx` process:

    Bind for 127.0.0.1:9083 failed: port is already allocated

This happens because the `nginx` process from the first run of
`make dev` is still running. This commit attempts to fix this by
configuring Supervisor to stop NGINX using the `TERM` signal instead
of `KILL` so that Docker has time to stop NGINX properly.

### DIFF
--- a/conf/supervisord-dev.conf
+++ b/conf/supervisord-dev.conf
@@ -13,7 +13,7 @@ stopasgroup=true
 command=docker compose run --rm --service-ports nginx-proxy
 stdout_events_enabled=true
 stderr_events_enabled=true
-stopsignal=KILL
+stopsignal=TERM
 stopasgroup=true
 
 [program:assets]
--- a/.cookiecutter/includes/conf/supervisord-dev.conf.json
+++ b/.cookiecutter/includes/conf/supervisord-dev.conf.json
@@ -4,7 +4,8 @@
       "command": "gunicorn -c conf/gunicorn/dev.conf.py --paste conf/development.ini"
     },
     "nginx": {
-      "command": "docker compose run --rm --service-ports nginx-proxy"
+      "command": "docker compose run --rm --service-ports nginx-proxy",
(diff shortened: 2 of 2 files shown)


conf/supervisord-dev.conf
```diff
@@ -13,7 +13,7 @@ stopasgroup=true
 command=docker compose run --rm --service-ports nginx-proxy
 stdout_events_enabled=true
 stderr_events_enabled=true
-stopsignal=KILL
+stopsignal=TERM
```
robertknight, line 16: According to http://supervisord.org/configuration.html, `TERM` is the default. Do you know why it was set to `KILL` originally?

robertknight: I guess that may have been done in ad354ab230cd9ac5aa6e959614c7739f4d726dea to try and make `make dev` terminate more quickly?

robertknight (approved):
This works for me, although `TERM` is the default stopsignal, so doesn't need to be explicitly specified.

conf/supervisord-dev.conf
seanh: Yes, we set it to `KILL` everywhere (across all our projects) as part of making sure that all the processes definitely get stopped (no zombie processes running) and stop quickly. We used to have a lot of problems with <kbd>Ctrl</kbd> + <kbd>c</kbd> taking a long time to return you to your prompt and then leaving zombie gunicorn and other processes running and occupying ports

seanh:
> This works for me, although `TERM` is the default stopsignal, so doesn't need to be explicitly specified.

Because of the way the cookiecutter templates the file we do actually need to specify `TERM`

PR #971, opened 2023-06-07
Add YouTube URL parsing

seanh:
Add YouTube URL parsing to extract the video ID from a YouTube URL. Nothing calls this yet.

Depends on https://github.com/hypothesis/via/pull/967


tests/unit/via/services/youtube_service_test.py
```diff
@@ -10,6 +10,26 @@ class TestYouTubeService:
     def test_enabled(self, enabled):
         assert YouTubeService(enabled=enabled).enabled == enabled
 
+    @pytest.mark.parametrize(
```
marcospri: Now that this is a separate PR we might want to tackle all of these here:

https://github.com/hypothesis/via/pull/912#discussion_r1184868371

Or merge this to get all the pieces together first

seanh: I think I'm probably going to have LMS normalize the YouTube URLs into some custom format (something like: `video://youtube/{id}` before sending them to Via's `/video?url={url}` view. So the `/video` view won't need to support real YouTube URLs, so I think I'll probably end up removing this real YouTube URL parsing from Via (and putting it into LMS instead).

We would still need the real YouTube URL parsing in Via if we ever wanted to add YouTube support to public Via: the root `/<REAL_YOUTUBE_URL>` view would have to parse and normalize real YouTube URLs in the same way that LMS is going to and forward them to `/video?url={normalized_youtube_url}`.

Anyway, I think we *will* need this real YouTube URL parsing code at least in LMS even if we don't add YouTube support to public Via, so I've added support for a few more formats and it'll get cut-pasted into LMS later.

marcospri: > Anyway, I think we will need this real YouTube URL parsing code at least in LMS even if we don't add YouTube support to public Via

The client is already doing that so if we don't care about support in public Via we could completely remove this and don't need a python implementation of it.

If we want to support it in via and for some other reason we also need it in the LMS backend we could consider adding the parsing to h-vialib. But we'll see when we get there.

marcospri (approved):

PR #972, opened 2023-06-08
Display bucket bar alongside transcript

### LINKED ISSUE #952: Explore showing bucket bar next to transcript container
The Hypothesis client's "bucket bar" indicating where annotations are is currently attached to the sidebar. In the context of HTML and this looks reasonable as the annotatable content spans the full size of the screen. In the context of video transcript annotation however, this results in the buckets feeling "detached" from the content that can be annotated:

Transcript and bucket bar

We think it might work better if the bucket bar could somehow be attached to the edge of the scrollable transcript instead. This issue exists to explore that possibility.

robertknight:
Use the Hypothesis client's `bucketBarContainer` config to position the bucket bar alongside the transcript, instead of the left edge of the client's sidebar.

**For the changes to work, you'll need to check out https://github.com/hypothesis/client/pull/5524**. Without the client changes, this PR has no visible effect for end users and the bucket bar is displayed along the edge of the sidebar as before.

Fixes https://github.com/hypothesis/via/issues/952.

**Before:**

**After:**

Transcript buckets

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -56,7 +56,7 @@ function isScrollToRangeEvent(e: Event): e is ScrollToRangeEvent {
 export default function VideoPlayerApp({
   videoId,
   clientSrc,
-  clientConfig,
+  clientConfig: baseClientConfig,
   transcript,
 }: VideoPlayerAppProps) {
   // Current play time of the video, in seconds since the start.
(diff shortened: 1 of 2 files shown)


robertknight:
This is what this currently looks like in a "stacked" layout where the transcript is displayed below the video, with the sidebar alongside:

Stacked layout

lyzadanger (approved):
👍🏻 

We can tweak local styling as we go and iterate...

Issue #953, opened 2023-06-02
Apply layout and design to YouTube video transcript annotation interface


lyzadanger:
This work pertains to the video-annotation interface (in via).

This work includes, but is not limited to:

* [x] basic layout and dimensions for baseline (i.e. desktop) viewport
* [x] adjustment of transcript (entries, spacing, typography, etc.)
* [x] adjustment of toolbar, buttons, and actions
* [x] alignment, both vertical and horizontal, of transcript with sidebar
* [x] branding
* [x] adjustments of scrolling and alignment per bucket bar
* [x] layout at different viewport sizes (breakpoints) (first pass only; iteration will be tracked separately)
* ~~error and messaging styling (potentially with toast messaging)~~ Tracked separately 

This work is done when we have had a design review and revisions can be tracked separately.

## Separately-tracked and follow-up issues directly related to this theme (not exhaustive)

* [Figure out what to do with the "Auto-scroll" control (design)](https://github.com/orgs/hypothesis/projects/118/views/1#)
* https://github.com/hypothesis/via/issues/954
* [Add a toast message component for use in video-annotations application](https://github.com/hypothesis/frontend-shared/issues/1075) (errors)

lyzadanger:
The initially-tracked tasks in this ticket have been completed, and all remaining design tasks are tracked in separate issues.

PR #975, opened 2023-06-08
Apply basic layout to video-annotation interface at target baseline viewport

### LINKED ISSUE #974: Apply basic layout to video-annotation app at baseline (desktop) widths
Apply styling for a basic layout that works best at our current target baseline viewport, which is a desktop or laptop browser window. The layout should be appropriate for both two-column (video, transcript) and three-column (side-by-side mode).

The current target viewport width for "baseline" desktop is 1448px.

Other target viewport layouts (breakpoints) will come later.

lyzadanger:
This PR applies a very basic layout for the video-annotation application to best suit (in broad strokes) the target baseline viewport of ~1448px as per https://github.com/hypothesis/via/issues/974

A few salient points:

* `AspectRatio` is used here to allow the video content to fill available space while retaining the correct aspect ratio
* The columnar layout is updated such that the column containing the transcript and controls is full height and fixed-width, and always flush against the right
* A subtle background color has been applied to the page. This is just author preference :)

Absolutely nothing here is carved in stone and should be considered iterative, as will be the case with several subsequent design-related PRs.

Many things are not addressed here. This only attempts to satisfy the first item in the UI/design list captured in https://github.com/hypothesis/via/issues/953

Before, 1448px:

Screen Shot 2023-06-08 at 9 04 25 AM

Screen Shot 2023-06-08 at 9 04 42 AM

After, 1448px:

Screen Shot 2023-06-08 at 9 02 53 AM

Screen Shot 2023-06-08 at 9 02 30 AM

Fixes https://github.com/hypothesis/via/issues/974
Part of #953

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -1,4 +1,5 @@
 import { Button, Checkbox, CopyIcon, Input } from '@hypothesis/frontend-shared';
+import classnames from 'classnames';
 import {
   useCallback,
   useEffect,
@@ -158,7 +159,15 @@ export default function VideoPlayerApp({
 
   return (
     <div className="w-full flex">
-      <div className="grow flex flex-col p-4">
+      <div
+        className={classnames(
+          // This column will grow in width per the parent flex container and
+          // allow the contained media (video) to scale with available space.
+          // The column flex layout established here ensures this container fills
+          // the full height of the parent flex container
+          'grow flex flex-col p-3'
+        )}
+      >
         <YouTubeVideoPlayer
           videoId={videoId}
           play={playing}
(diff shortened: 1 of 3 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -167,8 +176,25 @@ export default function VideoPlayerApp({
...
       </div>
-      <div className="w-2/5 h-[90vh] flex flex-col bg-grey-0 border">
-        <div className="p-1 bg-grey-1 border-b flex flex-col">
+      <div
+        className={classnames(
+          // Full-height column with a width allowing comfortable line lengths
+          'h-[100vh] w-[50ch] flex flex-col',
+          'bg-grey-0 border-x',
+          // TODO: This is a stopgap measure to prevent controls from being
+          // interfered with (overlaid) by sidebar controls and toolbar
```
lyzadanger, line 185: I'll explain some of the layout challenges vis-a-vis side-by-side mode in our next check-in call.

acelaya (approved):
Looking good!

PR #977, opened 2023-06-09
Design: Apply styling to transcript entries

lyzadanger:
This PR applies some styling to transcript entries, including fonts, segmentation, hover and active states.

Note that font sizes at present are smaller than they were, but are one tick larger than the sidebar and mockups because those felt extra tiny to me. But we can go smaller if desired.

All of these choices can be evaluated in our next design review...

More to come...

Before:

Screen Shot 2023-06-09 at 9 48 52 AM

After:
Screen Shot 2023-06-09 at 9 48 29 AM

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -1,8 +1,4 @@
-import {
-  Scroll,
-  ScrollContainer,
-  ScrollContent,
-} from '@hypothesis/frontend-shared';
+import { Scroll, ScrollContainer } from '@hypothesis/frontend-shared';
 import classnames from 'classnames';
 import type { Ref } from 'preact';
 import {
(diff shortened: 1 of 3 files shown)


via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -128,20 +124,37 @@ function TranscriptSegment({
...
+          // TODO: Use shared Button to get these styles for free
+          'flex transition-colors focus-visible-ring rounded-sm',
+          // `peer` allows Tailwind styling based on sibling state
+          'peer',
+          'font-medium hover:underline',
+          {
+            // Colors are one tick lighter than segment text
+            'text-stone-700': isCurrent,
+            'text-stone-500': !isCurrent,
+            'hover:text-stone-800': true,
```
acelaya, line 156: Why adding this here with `true` instead of adding it as a plain string before the object itself?

acelaya (approved):
👍🏼

PR #979, opened 2023-06-12
Use full YouTube URLs not video IDs in the `/video` route

seanh:
[Slack thread](https://hypothes-is.slack.com/archives/C054CG5PH5M/p1686231916609859), [another Slack thread](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1686563922909299).

Change Via's `/video` endpoint to accept full YouTube URLs in a `url` query param instead of accepting just YouTube video IDs in a URL path part.

This changes the format of Via's `/video` URLs from `/video/YOUTUBE_VIDEO_ID` to `/video?url=YOUTUBE_VIDEO_URL`.

This change is needed because the `/video` endpoint may need the full YouTube URL not just the video ID:

* The `/video` endpoint may need to pass the full YouTube URL to Checkmate to check whether the video has been blocked
* The `/video` endpoint may need the full YouTube URL for unforeseen functionality reasons. For example we may find out in future that our video player needs to handle YouTube `/live/` URLs differently from `/watch` ones

This commit also adds **webargs** validation for Via views and makes `/video` show an error page if the `url` query param is missing or is not a valid YouTube URL.

Testing
=======

I've used HTTP status **400** for the error pages rather than webargs's default of 422 because 400 is that Via was already using for other "bad URL" error pages.

No `url` query param: <http://localhost:9083/video>, status is 400
--------------------

image

`url` query param isn't a valid URL: <http://localhost:9083/video?url=foo>, status is 400
-----------------------------------

image

`url` query param isn't a YouTube URL: <http://localhost:9083/video?url=https://example.com>, status is 400
-------------------------------------

image

`url` query param is a valid YouTube URL: <http://localhost:9083/video?url=https://www.youtube.com/watch?v=jfKfPfyJRdk>
----------------------------------------

image

### DIFF
--- a/tests/unit/via/views/conftest.py
+++ b/tests/unit/via/views/conftest.py
@@ -1,0 +1,2 @@
+# Import this so that our webargs customization is registered in the view tests.
+import via.views.webargs  # pylint: disable=unused-import
--- a/tests/unit/via/views/view_video_test.py
+++ b/tests/unit/via/views/view_video_test.py
@@ -3,13 +3,14 @@ from unittest.mock import sentinel
 import pytest
 from pyramid.httpexceptions import HTTPUnauthorized
 
+from via.exceptions import BadURL
 from via.views.view_video import view_video
 
 
 @pytest.mark.usefixtures("youtube_service")
 class TestViewVideo:
-    def test_it(self, pyramid_request, Configuration):
-        pyramid_request.matchdict["id"] = "abcdef"
+    def test_it(self, pyramid_request, Configuration, youtube_service):
+        youtube_service.get_video_id.return_value = sentinel.video_id
 
         response = view_video(pyramid_request)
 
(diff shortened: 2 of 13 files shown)


tests/unit/via/views/conftest.py
```diff
@@ -0,0 +1,2 @@
+# Import this so that our webargs customization is registered in the view tests.
+import via.views.webargs  # pylint: disable=unused-import
```
seanh, line 2: Otherwise webargs's Pyramid parser's default error responses (which we've disabled) would still be active in the unit tests and validation errors would be `HTTPUnprocessableEntity` exceptions with no access to the underlying Marshmallow `ValidationError`

tests/unit/via/views/view_video_test.py
```diff
@@ -1,36 +1,65 @@
...
+                        "time": 0,
+                        "text": "First segment of transcript",
+                    },
+                    {
+                        "time": 30,
+                        "text": "Second segment of transcript",
+                    },
+                ],
+            },
+            "video_id": sentinel.youtube_video_id,
```
seanh, line 38: I combined the several separate `assert`'s here into one but the only real difference is that `"video_id"` is now `sentinel.youtube_video_id`

marcospri (approved):

PR #980, opened 2023-06-12
Change Via's `/video` endpoint to `/video/youtube`

seanh:
The reason for this is that we may want to have different query params
for different video platforms. For example lets say we have a `foo`
param that's required for YouTube videos but not allowed for Vimeo
videos. If the endpoint is `/video?url=YOUTUBE_OR_VIMEO_URL` then the
`foo` param is either required or not allowed depending on the value of
the `url` param. This could make validation difficult. It'll be easier
if `/video/youtube` and `/video/vimeo` are two different endpoints that
can easily be validated differently.


marcospri (approved):
Makes sense :+1:

PR #981, opened 2023-06-12
Remove the unused hardcoded YouTube transcript

seanh:
The frontend doesn't actually use this, and the plan is for real
transcripts to come from a separate API endpoint not from the
`view_video()` view.


marcospri (approved):
I see the frontend has:

https://github.com/hypothesis/via/blob/main/via/static/scripts/video_player/sample-transcript.ts

as an example anyway, :+1:

PR #982, opened 2023-06-12
Adapt video application layout based on size of app container element

lyzadanger:
This draft PR demonstrates a mechanism for adapting the layout of the video-annotation app based on the available size of the app's containing element. We cannot use media queries here because the viewport may be considerably wider than the app element when the sidebar side-by-side mode is activated. Container queries would be just the ticket but are not yet widely enough supported.

The concept I'd like to evaluate is the encapsulation of a certain amount of this logic in a hook (called `useLayoutHook` here), which internally uses a `ResizeObserver`. This hook returns a relative width value (e.g. 'sm', 'xl') that the consuming component can use to style itself. The breakpoint values in `useAppLayout` are a best guess first swag but shouldn't be construed final recommendations.

There some design challenges yet to sort out that I'm deferring until after we have a bucket-bar-on-transcript feature sorted and landed.

### Testing

On this branch, view the video player app. Resize your browser winder[^1] and open and close the sidebar at different sizes.

The layout stays multicolumn until an aggressively narrow amount of space is available[^2], and then will fall down to single-column. The transcript column width is adjusted at certain breakpoints to make this more possible. 

[^1]: This was a typo but I'm leaving it because I like the idea of cowgirl-me saying "Resize yer browser wind-er".

[^2]: The design "brief" implies that we should use a multi-column layout when it is at all possible.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -8,6 +8,7 @@ import {
   useState,
 } from 'preact/hooks';
 
+import { useAppLayout } from '../hooks/use-app-layout';
 import { useNextRender } from '../utils/next-render';
 import type { TranscriptData } from '../utils/transcript';
 import { formatTranscript } from '../utils/transcript';
@@ -65,6 +66,8 @@ export default function VideoPlayerApp({
   const [filter, setFilter] = useState('');
   const trimmedFilter = useMemo(() => filter.trim(), [filter]);
   const filterInputRef = useRef<HTMLInputElement>();
+  const appContainerRef = useRef<HTMLDivElement | null>(null);
+  const appSize = useAppLayout(appContainerRef);
 
   // Listen for the event the Hypothesis client dispatches before it scrolls
   // a highlight into view.
@@ -157,15 +160,26 @@ export default function VideoPlayerApp({
     }
   };
 
+  const multicolumn = appSize !== 'sm';
+
   return (
-    <div className="w-full flex">
+    <div
+      data-app-size={appSize}
+      data-multicolumn={multicolumn}
+      className={classnames('w-full flex', {
+        'flex-col min-h-0 h-[100vh]': !multicolumn,
+        'flex-row': multicolumn,
+      })}
+      ref={appContainerRef}
+    >
       <div
         className={classnames(
           // This column will grow in width per the parent flex container and
           // allow the contained media (video) to scale with available space.
           // The column flex layout established here ensures this container fills
           // the full height of the parent flex container
-          'grow flex flex-col p-3'
+          'flex flex-col p-3',
+          { grow: multicolumn }
         )}
       >
         <YouTubeVideoPlayer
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -157,15 +160,26 @@ export default function VideoPlayerApp({
...
+  const multicolumn = appSize !== 'sm';
+
   return (
-    <div className="w-full flex">
+    <div
+      data-app-size={appSize}
+      data-multicolumn={multicolumn}
+      className={classnames('w-full flex', {
+        'flex-col min-h-0 h-[100vh]': !multicolumn,
+        'flex-row': multicolumn,
```
lyzadanger, line 171: I'll organize and comment styles better if this moves ahead.

via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -0,0 +1,53 @@
...
+    onSizeChanged(element.clientWidth, element.clientHeight)
+  );
+  observer.observe(element);
+  return () => observer.disconnect();
+}
+
+/**
+ * Return a string representing the relative size of `appContainer`
+ */
+export function useAppLayout(appContainer: RefObject<HTMLDivElement | null>) {
```
robertknight, line 23: The API design of this hook currently doesn't allow for reacting to changes in the ref's value during the lifetime of the component. I think that's an acceptable limitation, as the hook doesn't pretend to be generic, but we should be clear about it.

via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -0,0 +1,53 @@
+import type { RefObject } from 'preact';
+import { useCallback, useLayoutEffect, useState } from 'preact/hooks';
+
+type AppSize = 'sm' | 'md' | 'lg' | 'xl' | '2xl';
+
+/**
+ * Watch for resizes on `element` and invoke `onSizeChanged` callback
+ */
+export function observeElementSize(
```
robertknight, line 9: If we're relying on `ResizeObserver`'s availability, this helper doesn't do much so it could be inlined into the hook below.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -157,15 +160,26 @@ export default function VideoPlayerApp({
     }
   };
 
+  const multicolumn = appSize !== 'sm';
+
   return (
-    <div className="w-full flex">
+    <div
+      data-app-size={appSize}
+      data-multicolumn={multicolumn}
```
robertknight, line 168: How do you plan to use these attributes? In tests? In CSS selectors? Both?

robertknight (commented):
The general strategy of using `ResizeObserver`, wrapped in a hook, seems good to me.

via/static/scripts/video_player/hooks/use-app-layout.ts
lyzadanger: Can you help me understand the implications of this a little better? The internal hooks have `appContainer` as a dependency, and this general setup reflects other recent hooks that take refs as an argument/option. Can you provide a scenario (not necessarily one we'd run into with this project, but in general) where this could pose an issue?

via/static/scripts/video_player/hooks/use-app-layout.ts
lyzadanger: Yeah, I had already considered that. Figured I'd refactor the hook structure once we agreed on the general sentiment.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
lyzadanger: I'm not even sure I'll retain these `data-` attributes yet. Need to take a deeper squint here and figure out how to manage styling (as somewhat convoluted tailwind classes based on `data-` or other attributes on the app element or using component logic based on `appSize`). TBD.

lyzadanger:
See #992

PR #988, opened 2023-06-13
Allow `via.*` query params to `view_video()`

seanh:
The URLs that h-vialib generates to the `view_video()` view include various `via.*` query params for configuring the client. Fix the view's new webargs schema to allow these `via.*` query params rather than raising a `ValidationError`.

This actually allows *any* unknown query params to be passed to the view in the `kwargs` argument, not just `via.*` params. But all the view does with `kwargs` is pass them to h-vialib's `extract_from_params()` method which only returns known params:

https://github.com/hypothesis/h-vialib/blob/e3d518e382d944024772dca7b3ee7a8b53be5767/src/h_vialib/configuration.py#L38C1-L48

### DIFF
--- a/tests/unit/via/views/view_video_test.py
+++ b/tests/unit/via/views/view_video_test.py
@@ -20,6 +20,9 @@ class TestViewVideo:
         response = youtube(pyramid_request)
 
         youtube_service.get_video_id.assert_called_once_with(video_url)
+        Configuration.extract_from_params.assert_called_once_with(
+            {"via.foo": "foo", "via.bar": "bar"}
+        )
         assert response == {
             "client_embed_url": "http://hypothes.is/embed.js",
             "client_config": Configuration.extract_from_params.return_value[1],
@@ -58,20 +61,22 @@ class TestViewVideo:
             youtube(pyramid_request)
 
     @pytest.fixture
-    def Configuration(self, patch):
-        Configuration = patch("via.views.view_video.Configuration")
-        Configuration.extract_from_params.return_value = (
-            sentinel.via_config,
-            sentinel.h_config,
-        )
-
-        return Configuration
-
-    @pytest.fixture
     def video_url(self):
         return "https://example.com/watch?v=VIDEO_ID"
 
     @pytest.fixture
     def pyramid_request(self, pyramid_request, video_url):
-        pyramid_request.params["url"] = video_url
+        pyramid_request.params.update(
+            {"url": video_url, "via.foo": "foo", "via.bar": "bar"}
+        )
         return pyramid_request
+
+
+@pytest.fixture(autouse=True)
+def Configuration(patch):
+    Configuration = patch("via.views.view_video.Configuration")
+    Configuration.extract_from_params.return_value = (
+        sentinel.via_config,
+        sentinel.h_config,
+    )
+    return Configuration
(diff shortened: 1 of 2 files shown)


tests/unit/via/views/view_video_test.py
```diff
@@ -57,21 +60,24 @@ def test_it_with_YouTube_transcripts_disabled(
...
 
-    @pytest.fixture
-    def Configuration(self, patch):
-        Configuration = patch("via.views.view_video.Configuration")
-        Configuration.extract_from_params.return_value = (
-            sentinel.via_config,
-            sentinel.h_config,
-        )
-
-        return Configuration
```
seanh, line 68: This fixture got moved down below

tests/unit/via/views/view_video_test.py
```diff
@@ -20,6 +20,9 @@ def test_it(self, pyramid_request, Configuration, youtube_service, video_url):
         response = youtube(pyramid_request)
 
         youtube_service.get_video_id.assert_called_once_with(video_url)
+        Configuration.extract_from_params.assert_called_once_with(
+            {"via.foo": "foo", "via.bar": "bar"}
+        )
```
seanh, line 25: Added a missing assertion. This assertion might have caught the mistake when my change that broke this changed this line to pass `{}` to `extract_from_params()` instead of the actual params

tests/unit/via/views/view_video_test.py
```diff
@@ -57,21 +60,23 @@ def test_it_with_YouTube_transcripts_disabled(
...
+
+
+@pytest.fixture(autouse=True)
+def Configuration(patch):
+    Configuration = patch("via.views.view_video.Configuration")
+    Configuration.extract_from_params.return_value = (
+        sentinel.via_config,
+        sentinel.h_config,
+    )
+    return Configuration
```
seanh, line 82: Moved this fixture to the bottom of the file (to make it global) and added `autouse=True` so that all tests in this file definitely use the mock `Configuration` object. Otherwise you can end up with tests that should fail passing because they're accidentally calling the real h-vialib

via/views/view_video.py
```diff
@@ -9,8 +10,10 @@
 
 
 @view_config(renderer="via:templates/view_video.html.jinja2", route_name="youtube")
-@use_kwargs({"url": fields.Url(required=True)}, location="query")
-def youtube(request, url):
+@use_kwargs(
+    {"url": fields.Url(required=True)}, location="query", unknown=marshmallow.INCLUDE
+)
+def youtube(request, url, **kwargs):
```
seanh, line 16: Added `unkwown=mardhsmallow.INCLUDE` so that unknown query params get passed through. See <https://webargs.readthedocs.io/en/latest/advanced.html#setting-unknown> and <https://marshmallow.readthedocs.io/en/stable/quickstart.html#handling-unknown-fields>. Previously the schema was raising `ValidationError` if there were any query params other than `url`.

Any unknown query params get passed to the view as kwargs, hence the `**kwargs`.

via/views/view_video.py
```diff
@@ -21,7 +24,7 @@ def youtube(request, url):
     if not video_id:
         raise BadURL(f"Unsupported video URL: {url}", url=url)
 
-    _, client_config = Configuration.extract_from_params(request.params)
+    _, client_config = Configuration.extract_from_params(kwargs)
```
seanh, line 27: We're only passing the unknown query params to `exctract_from_params()` rather than all query params as before, so the `url` query param (the onw known one) doesn't get passed. It doesn't seem to be necessary to pass `url` so this works.

marcospri (approved):

PR #989, opened 2023-06-13
Add Checkmate to `view_video()`

### LINKED ISSUE #978: Enable blocking of particular YouTube URLs

seanh:
URLs like `https://via.hypothes.is/video/youtube?url=<YOUTUBE_URL>` can be visited directly by users or might be embedded directly by LMS and would proxy a YouTube video even if that video were blocked by Checkmate.

Add a Checkmate call to the `view_video()` route so that these URLs will error if given a YouTube video that's blocked by Checkmate.

Fixes https://github.com/hypothesis/via/issues/978

I haven't tested this manually because I don't know of an easy way to add a URL to Checkmate's blocklist (and it doesn't seem to be documented in Checkmate's README).

### DIFF
--- a/tests/unit/via/views/view_video_test.py
+++ b/tests/unit/via/views/view_video_test.py
@@ -1,6 +1,7 @@
 from unittest.mock import sentinel
 
 import pytest
+from checkmatelib import BadURL as CheckmateBadURL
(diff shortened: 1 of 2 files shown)


marcospri (approved):
Looks very similar to the usage over the PDF view :+1:

Issue #991, opened 2023-06-14
Red "loading bar" showing forever in video-annotation app


lyzadanger:
After recent changes, there is a red "loading bar" that partially fills and never resolves at the top of the screen on the video annotation app (this is as seen in my localhost):

<img width="1579" alt="image" src="https://github.com/hypothesis/via/assets/439947/bdc0e17f-de83-4f0b-8228-681992990150">

This loading bar is outside of the `iframe` that contains the proxied content/video app.

_Update_: This behavior only happens when accessing the video-annotation app through URL patterns like:

http://localhost:9083/https://www.youtube.com/watch?v=eOJ6L_snYnA

And does not manifest with URL patterns like:

http://localhost:9083/video/youtube?url=https://www.youtube.com/watch?v=eOJ6L_snYnA

It should be noted that the first URL pattern is what Via will redirect you to if you paste a YouTube URL in the input field on via's landing page. See more discussion about the two different URL formats here: https://github.com/hypothesis/via/issues/941#issuecomment-1600764487

seanh:
FYI this only happens in public Via (where I don't think we have immediate plans to release the new YouTube player), doesn't happen in LMS.

PR #992, opened 2023-06-15
Adapt video-application layout at different sizes

lyzadanger:
This PR adds a mechanism for adapting video-annotation application layout at different app-container sizes and applies some initial layout adaptation. It also adds a new “top bar” with branding and the search/filtering input. This top bar extends the top bar of the client’s sidebar, visually.

While the mechanism for adapting layout is solid here, the applied layout/design (e.g. breakpoints) should be considered first-pass only. There are still some details to iron out. Expect at least one more iteration of the breakpoints/adaptations themselves.

NB: The red loading bar you see here is being [tracked separately](https://github.com/hypothesis/via/issues/991). 

A few screenshots:

Screen Shot 2023-06-15 at 3 18 04 PM

Screen Shot 2023-06-15 at 3 19 32 PM

Screen Shot 2023-06-15 at 3 21 10 PM

Screen Shot 2023-06-15 at 3 21 15 PM

### Known awkwardness

There are viewport widths around ~900px where the layout is awkward with the sidebar open. This is a band of widths at which side-by-side is not applied. We may choose to come at this by adjusting the minimum-width threshold of side-by-side for this application. Or we may still want to solve it from the video-app side. In any case, I will track it separately.

Screen Shot 2023-06-15 at 3 21 02 PM

Part of https://github.com/hypothesis/via/issues/953


via/static/scripts/test-util/wait.ts
lyzadanger, line 1: I know we've been talking about centralizing these test utils in a shared package, but I needed them nooowwwww :)

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -158,46 +169,40 @@ export default function VideoPlayerApp({
...
-        <YouTubeVideoPlayer
-          videoId={videoId}
-          play={playing}
-          time={timestamp}
-          onPlayingChanged={setPlaying}
-          onTimeChanged={setTimestamp}
-        />
-      </div>
+    <div
+      data-testid="app-container"
```
lyzadanger, line 173: I acknowledge that this component's layout is getting complicated. To help a tiny bit for now, I've attempted to comment the heck out of styling and add `data-testid`s to elements.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -158,46 +169,40 @@ export default function VideoPlayerApp({
...
-        <div
-          className={classnames(
-            'p-1 bg-grey-1',
-            // TODO: This is a stopgap measure to prevent the right side of the
-            // search input from being inpinged on by sidebar controls
-            'pr-2'
-          )}
-          data-testid="search-bar"
-        >
+        <LogoIcon />
```
lyzadanger: Presumably this should link to somewhere and have a `title`, etc. We'll get there.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -158,46 +169,40 @@ export default function VideoPlayerApp({
...
-            aria-label="Transcript filter"
             data-testid="filter-input"
+            aria-label="Transcript filter"
+            classes={classnames(
+              // Match height of search input in sidebar
+              'h-[32px]',
+              // TODO: Temporary prevention of sidebar controls overlapping
+              'mr-[22px]',
+              {
+                // Adapt width to match width of transcript
```
lyzadanger, line 207: If I can find a straightforward way to abstract some of this away in a future iteration, I will. I tried a couple of abortive things today but felt in danger of imposing premature abstraction. So for now please have patience with verbose styling until we get into more of a "clean-up" mode.

via/static/scripts/video_player/components/icons/Pause.tsx
```diff
@@ -14,6 +14,7 @@ export default function PauseIcon(props: PauseIconProps) {
       height="16"
       fill="none"
       data-component="PauseIcon"
+      viewBox="0 0 16 16"
```
lyzadanger, line 16: This prevents the SVG from getting clipped when rendered at smaller sizes.

via/static/scripts/video_player/hooks/test/use-app-layout-test.js
```diff
@@ -0,0 +1,82 @@
...
+        data-testid="appContainer"
+      >
+        <button>Hi</button>
+      </div>
+    );
+  }
+
+  function renderComponent() {
+    act(() => {
+      render(<FakeComponent />, container);
```
lyzadanger: NB: Using `preact`'s `render` here. I could not for the life of me find a test structure that worked with enzyme's `mount`.

via/static/scripts/video_player/hooks/test/use-app-layout-test.js
```diff
@@ -0,0 +1,82 @@
...
+    const steps = [
+      [800, 'md'],
+      [1000, 'lg'],
+      [1200, 'xl'],
+      [1500, '2xl'],
+      [400, 'sm'],
+    ];
+    for (const [width, expected] of steps) {
+      appContainerEl.style.width = `${width}px`;
+      await waitFor(() => appSize === expected, /* timeout */ 50);
```
lyzadanger: These tests regularly exceed the default `10ms` timeout locally. Is `50` enough leeway? I hope so? Advice welcome.

via/static/scripts/video_player/components/icons/Pause.tsx
robertknight: This is an auto-generated component, using the script from the frontend-shared repo. Can you update the source SVG and regenerate instead. I thought I added a README.md in `components/icons` to explain this, but it looks like I forgot.

via/static/scripts/test-util/wait.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Wait for a condition to evaluate to a truthy value.
+ *
+ * @return {Promise<T>} - Result of the `condition` function
+ */
+export async function waitFor<T>(
+  condition: () => T,
+  timeout = 10,
+  what = condition.toString()
+) {
```
robertknight:
```suggestion
): Promise<T> {
```

via/static/scripts/test-util/wait.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Wait for a condition to evaluate to a truthy value.
+ *
+ * @return {Promise<T>} - Result of the `condition` function
```
robertknight:
```suggestion
 * @return Result of the `condition` function
```

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -158,46 +169,40 @@ export default function VideoPlayerApp({
...
         className={classnames(
-          // Full-height column with a width allowing comfortable line lengths
-          'h-[100vh] w-[450px] flex flex-col',
-          'bg-grey-0 border-x',
-          // TODO: This is a stopgap measure to prevent controls from being
-          // interfered with (overlaid) by sidebar controls and toolbar
-          'mr-[30px]'
+          // Match height of sidebar's top-bar
+          'h-[40px] min-h-[40px]',
+          // Background is one shade darker than sidebar top-bar
```
robertknight, line 185: Maybe separate the layout and color classes here to make it more obvious what this comment is referring to.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
robertknight: Yes. See the logo link on eg. https://hypothes.is/search.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -209,21 +214,87 @@ export default function VideoPlayerApp({
...
+              'w-[450px]': appSize === 'xl',
+              'w-[410px]': appSize === 'lg',
+              'w-[380px]': appSize === 'md',
+            }
+          )}
+        >
+          <div
+            data-testid="transcript-controls"
+            className={classnames(
+              // Echo height of top-bar
```
robertknight: Can you expand this to clarify what "top-bar" is referring to here.

via/static/scripts/video_player/components/icons/Play.tsx
```diff
@@ -14,6 +14,7 @@ export default function PlayIcon(props: PlayIconProps) {
       height="16"
       fill="none"
       data-component="PlayIcon"
+      viewBox="0 0 16 16"
```
robertknight, line 16: See comment about icon component generation for Pause icon.

via/static/scripts/video_player/hooks/test/use-app-layout-test.js
robertknight: I think I see the issue you ran into. There are two things that need to be done to make this work using our normal way of writing tests:

1. Sync the Enzyme _root_ wrapper with the DOM when polling in `waitFor`, using `wrapper.update()`. Note this doesn't work on a child wrapper.
2. Re-fetch the child wrapper from the root after updating

```diff
diff --git a/via/static/scripts/video_player/hooks/test/use-app-layout-test.js b/via/static/scripts/video_player/hooks/test/use-app-layout-test.js
index 698fa9e..3b79835 100644
--- a/via/static/scripts/video_player/hooks/test/use-app-layout-test.js
+++ b/via/static/scripts/video_player/hooks/test/use-app-layout-test.js
@@ -1,22 +1,17 @@
-import { render } from 'preact';
+import { mount } from 'enzyme';
 import { useRef } from 'preact/hooks';
-import { act } from 'preact/test-utils';
 
 import { waitFor } from '../../../test-util/wait';
 import { useAppLayout } from '../use-app-layout';
 
 describe('useAppLayout', () => {
   let container;
-  let appSize;
-
-  function findElementByTestId(testId) {
-    return container.querySelector(`[data-testid=${testId}]`);
-  }
+  let wrappers;
 
   // Create a fake component to mount in tests that uses the hook
   function FakeComponent() {
     const myRef = useRef();
-    appSize = useAppLayout(myRef);
+    const appSize = useAppLayout(myRef);
     return (
       <div
         ref={myRef}
@@ -30,20 +25,21 @@ describe('useAppLayout', () => {
   }
 
   function renderComponent() {
-    act(() => {
-      render(<FakeComponent />, container);
-    });
-
-    return findElementByTestId('appContainer');
+    const wrapper = mount(<FakeComponent />, { attachTo: container });
+    wrappers.push(wrapper);
+    return wrapper;
   }
 
   beforeEach(() => {
     container = document.createElement('div');
     container.setAttribute('id', 'container');
     document.body.append(container);
+
+    wrappers = [];
   });
 
   afterEach(() => {
+    wrappers.forEach(w => w.unmount());
     container.remove();
   });
 
@@ -55,17 +51,19 @@ describe('useAppLayout', () => {
   ].forEach(({ containerWidth, expected }) => {
     it('should provide relative app size when component is rendered', () => {
       container.style.width = `${containerWidth}px`;
-
-      renderComponent();
-      assert.equal(appSize, expected);
+      const appContainerEl = renderComponent().find(
+        '[data-testid="appContainer"]'
+      );
+      assert.equal(appContainerEl.prop('data-app-size'), expected);
     });
   });
 
   it('should update app size if relative size of container changes', async () => {
     container.style.width = '200px';
 
-    const appContainerEl = renderComponent();
-    assert.equal(appSize, 'sm');
+    const wrapper = renderComponent();
+    const appContainerEl = wrapper.find('[data-testid="appContainer"]');
+    assert.equal(appContainerEl.prop('data-app-size'), 'sm');
 
     const steps = [
       [800, 'md'],
@@ -75,8 +73,13 @@ describe('useAppLayout', () => {
       [400, 'sm'],
     ];
     for (const [width, expected] of steps) {
-      appContainerEl.style.width = `${width}px`;
-      await waitFor(() => appSize === expected, /* timeout */ 50);
+      container.style.width = `${width}px`;
+
+      await waitFor(() => {
+        wrapper.update();
+        const appContainerEl = wrapper.find('[data-testid="appContainer"]');
+        return appContainerEl.prop('data-app-size') === expected;
+      }, /* timeout */ 50);
     }
   });
 });
```

via/static/scripts/video_player/hooks/test/use-app-layout-test.js
robertknight: I'm not familiar with `ResizeObserver`'s reporting frequency. It might be synced to `requestAnimationFrame` in which case it would likely be 16ms (60 frames/sec). 50ms might be OK.

via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -0,0 +1,55 @@
...
+type AppSize = 'sm' | 'md' | 'lg' | 'xl' | '2xl';
+
+/**
+ * Return a string representing the relative size of `appContainer` that can
+ * be used to make decisions about app layout and sizing of components. This
+ * would typically be something handled by media queries, but it is an element
+ * (`body`) width that needs to be queried, not the viewport. We could replace
+ * this with container queries once browser support is sufficient.
+ */
+export function useAppLayout(appContainer: RefObject<HTMLDivElement | null>) {
```
robertknight: I suggest to explicitly specify the return type here for the reader, and also export the `AppSize` type.

via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  function observeElementSize(
+    element: Element,
+    onSizeChanged: (width: number, height: number) => void
+  ): () => void {
+    const observer = new ResizeObserver(() =>
+      onSizeChanged(element.clientWidth, element.clientHeight)
+    );
+    observer.observe(element);
+    return () => observer.disconnect();
```
robertknight: The `observeElementSize` helper seems redundant to me if we're only using `ResizeObserver` internally. It made more sense when it was abstracting over different approaches.

via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    setAppSize(containerSize);
+  }, [appContainer]);
+
+  useLayoutEffect(() => {
+    const cleanup = observeElementSize(appContainer.current!, updateWidth);
+    updateWidth();
+    return cleanup;
+  }, [updateWidth, appContainer]);
```
robertknight, line 51: This layout effect will never re-run if the `appContainer.current` value changes, because the hook is watching the ref object rather than the `current` property.

Some options:

- Accept this limitation, but call it out in comments. I think this is the easiest option.
- Re-work hook to take an observable ref of some sort and then do the work to handle changes in the value internally. This effort is not justified at present IMO, unless we can find a really easy way of doing it.

robertknight (commented):
The general shape of this makes sense. I left some notes on the tests, icons and a couple of other things. I understand the layouts are not finalized yet. The one major comment I have on it is that I think we should make more space for the transcript in stacked layout and also have the transcript search field nearer the actual transcript in that case. It seems weird to have the video in the middle between them.

via/static/scripts/video_player/hooks/test/use-app-layout-test.js
lyzadanger: The `requestAnimationFrame` involvement here hadn't occurred to me. Thanks for the note.

via/static/scripts/video_player/hooks/use-app-layout.ts
lyzadanger: Good point. I kind of looked past it since it was copied code. I'll unfurl it.

via/static/scripts/video_player/components/icons/Pause.tsx
lyzadanger: I'm also adding a README

via/static/scripts/video_player/components/VideoPlayerApp.tsx
lyzadanger: It should just be worded "Same height as top-bar". I'm updating the comment.

lyzadanger:
Updated:

* Icon components regenerated, and README added
* Updated return types in `wait`
* Explicit return type in hook; export `AppSize` type; document limitation with `appContainer.current`; unrolled internal function
* Hypothesis icon linked
* Updated styling comments in `VideoPlayerApp`
* Applied suggested test changes (and wrapped head around them)

Deferring, but tracked per suggestion:

#993
#994

via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    setAppSize(containerSize);
+  }, [appContainer]);
+
+  useLayoutEffect(() => {
+    const element = appContainer.current;
+    /* istanbul ignore if */
+    if (!element) {
+      return () => {};
+    }
```
lyzadanger, line 46: This state should never occur. Should I remove this guard?

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -158,46 +169,48 @@ export default function VideoPlayerApp({
...
-            'p-1 bg-grey-1',
-            // TODO: This is a stopgap measure to prevent the right side of the
-            // search input from being inpinged on by sidebar controls
-            'pr-2'
-          )}
-          data-testid="search-bar"
+        <a
+          href="https://web.hypothes.is"
+          target="_blank"
+          rel="noreferrer"
```
lyzadanger, line 192: This is primarily to jailbreak out of via

via/static/scripts/video_player/hooks/use-app-layout.ts
robertknight: I would probably leave as is, or maybe log a warning. You could make an argument that having `appContainer.current` be null is likely a mistake and so the code should fail loudly. On the other hand, the hook is slightly more flexible in its current state.

robertknight (approved):

Issue #955, opened 2023-06-05
Explore combining transcript segments together


robertknight:
The video player transcript currently presents each segment, as it appears in the raw caption data from YouTube, as a separate row:

<img width="579" alt="Video player transcript" src="https://github.com/hypothesis/via/assets/2458/385ff851-7362-47c0-8edd-0e128ddfa44f">

In docdrop.org's video player, multiple segments are combined together.

<img width="524" alt="docdrop transcript" src="https://github.com/hypothesis/via/assets/2458/4d849916-8d05-46f5-a447-35483181c364">

Combining segments together can improve readability, by making the transcript read more like prose. However it can also make it more difficult to match what is currently being said in the video with the corresponding text, since the text block that is highlighted as "current" becomes larger.

This issue is about doing some exploration here and settling on something reasonable for the first version of the video player

robertknight:
A constraint is that auto-generated transcripts from YouTube don't include reliable information about the natural boundaries of sentences or paragraphs that you would find in written prose. This means that we can't trivially start / end segments at "natural" places.

robertknight:
An initial stab at this has been implemented in https://github.com/hypothesis/via/pull/998. We'll tweak this over time.

PR #998, opened 2023-06-19
Combine each group of 3 consecutive segments in the raw transcript

robertknight:
Segments in typical YouTube transcripts are very short. Grouping them can improve readability, at the cost of making it less precise which part of the transcript corresponds to the current video position. As an initial iteration of grouping, just combine every 3 segments together.

Ideally it might be nice to have segments be tidy linguistic units such as sentences, but auto-generated YouTube transcripts often do not contain the information to do this.

I also adjusted the layout to justify text and add padding to segments to avoid them being covered by buckets from the client's bucket bar.

Part of https://github.com/hypothesis/via/issues/955.

----

**Before:**

Ungrouped segments

**After:**

Grouped segments

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -181,10 +181,16 @@ function TranscriptSegment({
         data-timestamp={timestamp}
       />
       <p
-        className={classnames('grow peer-hover:text-stone-900', {
-          'text-stone-600': !isCurrent,
-          'text-stone-800': isCurrent,
-        })}
+        className={classnames(
+          'grow text-justify peer-hover:text-stone-900',
+          {
+            'text-stone-600': !isCurrent,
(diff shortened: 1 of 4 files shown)


via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -181,10 +181,16 @@ function TranscriptSegment({
         data-timestamp={timestamp}
       />
       <p
-        className={classnames('grow peer-hover:text-stone-900', {
-          'text-stone-600': !isCurrent,
-          'text-stone-800': isCurrent,
-        })}
+        className={classnames(
+          'grow text-justify peer-hover:text-stone-900',
```
lyzadanger, line 185: I have a task in https://github.com/hypothesis/via/issues/953 to do styling work around the buckets: that's still an open task. This is fine for now, however.

lyzadanger (approved):
LGTM.

PR #1000, opened 2023-06-19
Add transcript API

### LINKED ISSUE #990: Add a Via backend API for getting YouTube video transcripts
To be called by Via's own frontend in order to get the transcript of the current video.

seanh:
Add a new API endpoint for the frontend code to call (in a future commit) that returns the transcript of a given YouTube video.

Fixes <https://github.com/hypothesis/via/issues/990>. See <https://github.com/hypothesis/via/issues/990#issuecomment-1589700767> for notes on the design of the new API.

The endpoint currently returns a hard-coded fake transcript.

This is Via's first API endpoint so I had to set up the framework for an API:

* PyJWT is added as a requirement, for encoding and decoding JWT's for API authentication and authorization
* A new required setting `API_JWT_SECRET` is added, used for encoding and decoding JWT's
* A new Pyramid security policy is added for authenticating and authorizing requests based on bearer token JWT's in the request's `Authorization` header
* The new API view is added as `/api/youtube/transcript/{id}`
* New data is added to the `js-config` object that the `view_video()` view passes to the frontend telling the frontend about the transcript API call: HTTP method, URL, and `Authorization` header (including an encoded JWT that's passed to the frontend)
* A new API exception views are added for 404s and 403s

The JWT's that the backend generates and passes to the frontend grant a general `"api"` permission, so any token can be used to call any API with any arguments. Currently the transcript API is the only API that exists to call. But this does mean that any token can be used to access any transcript. This seems acceptable for the use-case.

The JWT's expire after a generous timeout.

## Testing

Visit a YouTube page (e.g. <http://localhost:9083/https://www.youtube.com/watch?v=kQNqmUZrEOk>) and inspect the source code and check the API config. You should see something like this:

```
<script type="application/json" class="js-config">
{
  ...,
  "api": {
    "transcript": {
      "doc": "Get the transcript of the current video",
      "headers": {"Authorization": "Bearer eyJ***19Y"},
      "method": "GET",
      "url": "http://localhost:9083/api/youtube/transcript/1"
    }
  }
}
</script>
```

Try copy-pasting the authorization header out of the API config and using it to make an API request. For example using [HTTPie](https://httpie.io/):

```terminal
$ http GET 'http://localhost:9083/api/youtube/transcript/1' Authorization:"Bearer eyJ***4hs"
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 21814
Date: Mon, 19 Jun 2023 15:53:09 GMT
Server: nginx/1.17.6
X-Via: compute

{
    "segments": [
        {
            "start": 0.0,
            "text": "[Music]"
        },
        ...
        {
            "start": 755.2,
            "text": "you"
        }
    ]
}
```

See what happens if you change the transcript ID to something other than `1`: you should get a JSON 404 error:

```terminal
$ http GET 'http://localhost:9083/api/youtube/transcript/2' Authorization:"Bearer eyJ***4hs"
HTTP/1.1 404 Not Found
Connection: keep-alive
Content-Length: 4
Date: Mon, 19 Jun 2023 15:54:36 GMT
Server: nginx/1.17.6

null
```

See what happens if you change the authorization header to something invalid or remove it entirely, you should get a JSON 403 error:

```terminal
$ http GET 'http://localhost:9083/api/youtube/transcript/1' Authorization:"Bearer foo"
HTTP/1.1 403 Forbidden
Connection: keep-alive
Content-Length: 4
Date: Mon, 19 Jun 2023 15:55:28 GMT
Server: nginx/1.17.6

null
```

### DIFF
--- a/tests/unit/via/security_test.py
+++ b/tests/unit/via/security_test.py
@@ -1,0 +1,46 @@
+from unittest.mock import sentinel
+
+import pytest
+from h_matchers import Any
+from pyramid.security import Allowed, Denied
+
+from via.security import ViaSecurityPolicy
+
+
+class TestViaSecurityPolicy:
+    def test_permits_when_theres_no_authorization_header(self, pyramid_request):
+        result = ViaSecurityPolicy().permits(pyramid_request, sentinel.context, "api")
+
+        assert result == Any.instance_of(Denied).with_attrs(
+            {"msg": "No Authorization header"}
+        )
+
+    def test_permits_when_the_jwt_is_invalid(self, pyramid_request):
+        pyramid_request.headers["Authorization"] = "Bearer invalid_token"
+        result = ViaSecurityPolicy().permits(pyramid_request, sentinel.context, "api")
+
+        assert result == Any.instance_of(Denied).with_attrs({"msg": "Invalid JWT"})
+
+    @pytest.mark.usefixtures("with_valid_authorization_header")
+    def test_permits_when_the_permission_is_unknown(self, pyramid_request):
+        result = ViaSecurityPolicy().permits(
+            pyramid_request, sentinel.context, "unknown_permission"
+        )
+
+        assert result == Any.instance_of(Denied).with_attrs(
+            {"msg": "Unknown permission"}
+        )
+
+    @pytest.mark.usefixtures("with_valid_authorization_header")
+    def test_permits_when_the_jwt_and_permission_are_valid(self, pyramid_request):
+        result = ViaSecurityPolicy().permits(pyramid_request, sentinel.context, "api")
+
+        assert result == Any.instance_of(Allowed).with_attrs(
+            {"msg": "Valid JWT provided"}
+        )
+
+    @pytest.fixture
+    def with_valid_authorization_header(self, pyramid_request):
+        pyramid_request.headers[
+            "Authorization"
+        ] = f"Bearer {ViaSecurityPolicy.encode_jwt(pyramid_request)}"
--- a/via/security.py
+++ b/via/security.py
@@ -1,0 +1,39 @@
+from datetime import datetime, timedelta, timezone
+
+import jwt
+from pyramid.security import Allowed, Denied
+
+
+class ViaSecurityPolicy:
+    def permits(self, request, _context, permission):
+        try:
+            encoded_jwt = request.headers["Authorization"][len("Bearer ") :]
+        except KeyError:
+            return Denied("No Authorization header")
+
+        try:
+            jwt.decode(
+                encoded_jwt,
+                self._get_jwt_secret(request),
+                algorithms=["HS256"],
+                options={"require": ["exp"]},
+            )
+        except jwt.InvalidTokenError:
+            return Denied("Invalid JWT")
+
+        if permission == "api":
+            return Allowed("Valid JWT provided")
+
+        return Denied("Unknown permission")
+
+    @classmethod
+    def encode_jwt(cls, request):
+        return jwt.encode(
+            {"exp": datetime.now(tz=timezone.utc) + timedelta(hours=48)},
+            cls._get_jwt_secret(request),
+            algorithm="HS256",
+        )
+
+    @staticmethod
+    def _get_jwt_secret(request):
+        return request.registry.settings["api_jwt_secret"]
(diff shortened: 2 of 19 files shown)


tests/unit/via/security_test.py
```diff
@@ -0,0 +1,46 @@
...
+from via.security import ViaSecurityPolicy
+
+
+class TestViaSecurityPolicy:
+    def test_permits_when_theres_no_authorization_header(self, pyramid_request):
+        result = ViaSecurityPolicy().permits(pyramid_request, sentinel.context, "api")
+
+        assert result == Any.instance_of(Denied).with_attrs(
+            {"msg": "No Authorization header"}
+        )
```
seanh, line 16: Note that we're asserting about the message strings that the security policy passes to `Denied()`. This is to defend against bugs in the tests. For example a test that the security policy denies a request with an invalid JWT might have a bug in the test code such that it doesn't actually put _any_ JWT in the request. The security policy would still return `Denied()` (because there's no JWT) and the test would pass. The security policy could actually be failing to deny requests that do contain an invalid JWT and the test would still be passing. Having the tests assert about the denied reason defends against this.

The use of `h_matchers` here is because `Denied("foo") == Denied("bar")` is `True`!

Similar to asserting about exception messages when testing exception raising

via/views/api/youtube.py
```diff
@@ -0,0 +1,1328 @@
...
+
+    if transcript_id != "1":
+        raise HTTPNotFound()
+
+    # pylint:disable=too-many-lines
+    return {
+        "data": {
+            "type": "transcripts",
+            "id": transcript_id,
+            "attributes": {
```
seanh: The `"data"`, `"type"`, `"id"` and `"attributes"` are [JSON:API](https://jsonapi.org/)-style design. See https://github.com/hypothesis/via/issues/990#issuecomment-1589700767. If we add more JSON:API-style endpoints to Via we might want to add some helper code for this, rather than each view having to return the right format of JSON manually.

via/views/view_video.py
```diff
@@ -32,4 +33,14 @@ def youtube(request, url, **kwargs):
...
+        "api": {
+            "transcript": {
+                "doc": "Get the transcript of the current video",
+                "url": request.route_url("api.youtube.transcript", transcript_id="1"),
+                "method": "GET",
+                "headers": {
+                    "Authorization": f"Bearer {ViaSecurityPolicy.encode_jwt(request)}"
+                },
+            }
+        },
```
seanh, line 43: As we add more endpoints we might want to write some helper code for generating these API config objects, rather than each view doing it manually

via/security.py
```diff
@@ -0,0 +1,39 @@
...
+
+        return Denied("Unknown permission")
+
+    @classmethod
+    def encode_jwt(cls, request):
+        return jwt.encode(
+            {"exp": datetime.now(tz=timezone.utc) + timedelta(hours=48)},
+            cls._get_jwt_secret(request),
+            algorithm="HS256",
+        )
```
seanh, line 35: We might want to add something more than just the expiry time to the JWT payload. For example I wonder if we could encode the permissions that the JWT grants into the JWT?

via/views/__init__.py
```diff
@@ -28,6 +28,10 @@ def add_routes(config):  # pragma: no cover
     config.add_route("proxy_d2l_pdf", "/d2l/proxied.pdf", factory=QueryURLResource)
     config.add_route("proxy_python_pdf", "proxied.pdf", factory=QueryURLResource)
 
+    config.add_route(
+        "api.youtube.transcript", "/api/youtube/transcript/{transcript_id}"
```
marcospri, line 32: Is this a new view?

seanh: This is the new route for the new API endpoint, and the new view for it is in `views/api/youtube.py`

marcospri (approved):
Looks good, done test testing as works as described.

My only feedback is about using `transcript_id` as a new concept. The API is already scoped to youtube so why not use the video_id itself?

I reckon once we introduce language selection it should be video_id + language. Is that the rationale for the transcript ID? I reckon I'd still use video id explicitly there.

seanh:
> My only feedback is about using `transcript_id` as a new concept. The API is already scoped to youtube so why not use the video_id itself?

Yeah I this this should be `video_id`, I'll change it.

> I reckon once we introduce language selection it should be video_id + language.

I think it might be `video_id` + `transcript_id` rather than language, although I haven't actually looked into the YouTube API that we're going to be calling yet, but I suspect it may be possible to have multiple transcripts for the same language? So it'll need a unique transcript ID rather than a language.

I think maybe the `video_id` will be in the path but the `transcript_id` will be an optional query param, and if you don't specify a `transcript_id` then Via chooses one automatically.

Anyway we'll come to this when we add transcript/language selection

PR #1002, opened 2023-06-20
Update frontend tooling

acelaya:
This PR makes some updates around frontend tooling

* Updates to `@hypothesis/frontend-build` v2, so that we take advantage of the latest improvement in the library, while we also get rid of a peer dependency mismatch warning when running `yarn install`.
* Update `sass` package, which is required for one of the latest changes in `@hypothesis/frontend-build`
* Add missing frontend make tasks (`frontend-lint`, `frontend-test` and `frontend-format`)
* Add missing `yarn` scripts, required for some of the point above to work (`test` and `test:watch`)
* Reformat two icons that were not currently following our coding style.

### DIFF
--- a/via/static/images/icons/pause.svg
+++ b/via/static/images/icons/pause.svg
@@ -1,4 +1,10 @@
-<svg xmlns="https://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16" fill="none">
+<svg
+  xmlns="https://www.w3.org/2000/svg"
+  width="16"
+  height="16"
+  viewBox="0 0 16 16"
+  fill="none"
+>
   <path
     fill="currentColor"
     fill-rule="evenodd"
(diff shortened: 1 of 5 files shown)


acelaya:
Converting to draft, as some tasks are failing and need to investigate a bit

robertknight (approved):
LGTM. I notice that `make frontend-format` is causing a couple of icons to be re-formatted. This suggests that: a) we are missing a frontend-formatting check in our CI and b) I suggest to run `make frontend-format` and commit the changes in this PR. It could also be an artifact of a Prettier update.

acelaya:
> I suggest to run `make frontend-format` and commit the changes in this PR

I actually had it committed, but ended up rolling it back 😅

via/static/images/icons/pause.svg
acelaya, line 1: This is just a formatting change, result of running `make frontend-format`.

You can see the actual changes by clicking here: 
image

via/static/images/icons/play.svg
acelaya, line 1: This is just a formatting change, result of running `make frontend-format`.

You can see the actual changes by clicking here: 
image

PR #1003, opened 2023-06-21
Provide right-channel buckets styling for all layouts

lyzadanger:
This PR attempts to solve the problem of space along the right margin and add a channel backdrop for buckets for all layouts. It was a technical challenge!

This is probably the last "messy" applied-design PR before extracting, cleaning up and consolidating layout-related components and logic.

#### With sidebar closed

**Before**:

Screen Shot 2023-06-21 at 10 29 07 AM

**After**:
Screen Shot 2023-06-21 at 10 27 45 AM

#### With sidebar open

**Before**:

Screen Shot 2023-06-21 at 10 29 16 AM

**After**:
Screen Shot 2023-06-21 at 10 28 03 AM

## Single-column layout

(No change when sidebar open)

**Before**:

Screen Shot 2023-06-21 at 10 29 26 AM

**After**:

Screen Shot 2023-06-21 at 10 28 14 AM

NB: This may render https://github.com/hypothesis/client/issues/5554 non-applicable as we've solved it in a different way here. It's still a bit of a shame to lose 20px at the right edge in narrow screens, but not a crisis.

Part of https://github.com/hypothesis/via/issues/953

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -1,6 +1,6 @@
 import { Scroll, ScrollContainer } from '@hypothesis/frontend-shared';
 import classnames from 'classnames';
-import type { Ref } from 'preact';
+import type { ComponentChildren, Ref } from 'preact';
 import {
   useCallback,
   useEffect,
@@ -31,6 +31,11 @@ export type TranscriptProps = {
    * the segment corresponding to `currentTime` in view.
    */
   autoScroll?: boolean;
+
+  /** Other content to render in scrolling container after the transcript
+   * segments.
+   */
+  children?: ComponentChildren;
 
   transcript: TranscriptData;
 
@@ -182,14 +187,14 @@ function TranscriptSegment({
       />
       <p
         className={classnames(
-          'grow text-justify peer-hover:text-stone-900',
+          'grow peer-hover:text-stone-900',
           {
             'text-stone-600': !isCurrent,
             'text-stone-800': isCurrent,
           },
 
           // Avoid buckets overlapping highlighted text.
-          'pr-[30px]'
+          'pr-3'
         )}
         data-testid="transcript-text"
         ref={contentRef}
@@ -222,6 +227,7 @@ function offsetRelativeTo(element: HTMLElement, parent: HTMLElement): number {
 
 export default function Transcript({
   autoScroll = true,
+  children,
   controlsRef,
   currentTime,
   filter = '',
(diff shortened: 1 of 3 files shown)


acelaya (approved):

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -32,6 +32,8 @@ export type TranscriptProps = {
    */
   autoScroll?: boolean;
 
+  children?: ComponentChildren;
```
robertknight, line 38: I think it would be helpful here to say what the `Transcript` does with its children and what assumptions the caller is allowed to make about the layout into which it is rendered, since that isn't immediately obvious.

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -182,14 +184,14 @@ function TranscriptSegment({
       />
       <p
         className={classnames(
-          'grow text-justify peer-hover:text-stone-900',
+          'grow peer-hover:text-stone-900',
```
robertknight, line 190: What was the rationale for the text justification change? Is that an aesthetic preference or was there something about the new layout which made it work less well?

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -245,8 +249,10 @@ export default function VideoPlayerApp({
             data-testid="filter-container"
             className={classnames(
               'text-right',
-              // TODO: Temporary prevention of sidebar controls overlapping
-              'mr-[22px]'
+              // Put space to the right of the filter input so it is not
+              // overlaid by sidebar controls. NB: Cannot use margin because it
+              // gets "consumed" in side-by-side mode
```
robertknight, line 254: I think we've agreed to enable the video player to take more control over how side-by-side mode works, and probably entirely replace the client's default implementation. Would you mind filing an issue in the frontend and linking it here?

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -385,18 +389,35 @@ export default function VideoPlayerApp({
...
               onSelectSegment={segment => setTimestamp(segment.start)}
-            />
+            >
+              <div
+                data-testid="bucket-bar-channel"
+                className={classnames(
+                  // Provide a backdrop for bucket bar buttons. 20px of this
+                  // is overlaid by the sidebar's semi-transparent bucket
+                  // channel.
+                  'bg-gradient-to-r from-white to-grey-1 border-l w-[40px]'
```
robertknight, line 399: Do you have an idea of what a better solution would look like with additional changes in the client?

robertknight:
As an observation about the single-column layout, it feels odd that we currently have very little space for the transcript when the viewport is just narrow enough to use the layout, but the space for the transcript grows as you continue to make the viewport narrower and the video shrinks.

robertknight (approved):
Functionally this looks to work as intended. I might have some thoughts on visual design, but I don't know if now is the time to discuss that or whether you have other changes you want to land first.

via/static/scripts/video_player/components/Transcript.tsx
lyzadanger: Justification is a tricky beast (this is my typesetting days coming through). You have to have a reliable (reasomably-long) line-character length to avoid big white rivers of space, and though justified text is sometimes aesthetically satisfying for the designer because of its tight edges, it's actually easier to read ragged-right text (in left-to-right languages, anyway). When working with the layout, I spotted some lines that were getting gappy and awkward.

I can cite references if you'd like me to take the time to do so!

lyzadanger:
> As an observation about the single-column layout, it feels odd that we currently have very little space for the transcript when the viewport is just narrow enough to use the layout, but the space for the transcript grows as you continue to make the viewport narrower and the video shrinks.

Perfectly reasonable feedback! I've been thinking about tackling this by establishing a reasonable max-sizing on the video so that it doesn't take up too much space (vertical _or_ horizontal). I'd consider it orthogonal to these changes, but am happy to pick it up soon.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
lyzadanger: Yep. I need to gather my thoughts, but, yes.

PR #1004, opened 2023-06-22
Replace `request.checkmate` with `CheckmateService`

seanh:
Replace `request.checkmate` (the `ViaCheckmateClient` class which is [implemented in a top-level `via/checkmate.py` file](https://github.com/hypothesis/via/blob/d19c83df23de7b4d67b0d8aba37d218b3b74ade0/via/checkmate.py#L8-L43)) with a new `CheckmateService`.

This is going to enable future PRs (e.g. https://github.com/hypothesis/via/pull/1009) where I need other services to depend on `CheckmateService`, but I think it stands alone as an improvement by itself.

The new `CheckmateService` also uses composition to wrap `checkmatelib.CheckmateClient` rather than inheriting from it.

In general this approach is longer (in terms of line-count) but I think it's much clearer and separates various concerns better. Also a big part of the reason why this is longer is because it has more detailed unit test coverage.


tests/unit/conftest.py
```diff
@@ -39,12 +37,5 @@ def pyramid_request(
     pyramid_config,  # pylint:disable=unused-argument
 ):
     pyramid_request = testing.DummyRequest()
-
     apply_request_extensions(pyramid_request)
-
-    pyramid_request.checkmate = create_autospec(
-        ViaCheckmateClient, spec_set=True, instance=True
-    )
-    pyramid_request.checkmate.check_url.return_value = None
```
seanh, line 48: We no longer need `request.checkmate` in the tests, since it isn't used/doesn't exist anymore

tests/unit/services.py
```diff
@@ -34,6 +35,11 @@ def via_client_service(mock_service):
     return mock_service(ViaClientService)
 
 
+@pytest.fixture
+def checkmate_service(mock_service):
+    return mock_service(CheckmateService)
```
seanh, line 40: Add a mock for the new `CheckmateService` to the tests instead

tests/unit/via/services/checkmate_test.py
```diff
@@ -0,0 +1,133 @@
...
+class TestCheckmateService:
+    def test_check_url(self, checkmate_client, svc):
+        result = svc.check_url(sentinel.url)
+
+        checkmate_client.check_url.assert_called_once_with(
+            sentinel.url,
+            allow_all=sentinel.allow_all,
+            blocked_for=sentinel.blocked_for,
+            ignore_reasons=sentinel.ignore_reasons,
+        )
```
seanh, line 22: By using composition instead of inheritance these assertions are much clearer. You can clearly see that it's asserting that a method on `checkmate_client` was called. And if you're not sure what `checkmate_client` is you can look down at that fixture and see exactly what it is, but the clear and consistent naming should make it pretty obvious: it's the `checkmate_client` argument to `CheckmateService.__init__()`, saved as `self._checkmate_client` within `CheckmateService`, and it's an instance of `checkmatelib`'s `CheckmateClient` class.

Because the old `ViaCheckmateClient` class **inherited** from `checkmatelib.CheckmateClient` rather than composing it had to do something much more confusing here:

For example there was this test that has no assertions and that writes some private attributes of `checkmatelib` (an external library's) `CheckmateClient`:

https://github.com/hypothesis/via/blob/d19c83df23de7b4d67b0d8aba37d218b3b74ade0/tests/unit/via/checkmate_test.py#L14-L17

The tests also had a fixture that patches one of the methods of the `ViaCheckmateClient` object-under-test, and even more confusingly the method being patched is a method that's not actually implemented in `ViaCheckmateClient` itself, it's inherited from `checkmatelib.CheckmateClient`, so what this is actually doing is reaching into `checkmatelib` to patch one of its methods:

https://github.com/hypothesis/via/blob/d19c83df23de7b4d67b0d8aba37d218b3b74ade0/tests/unit/via/checkmate_test.py#L64-L68

The tests then had assertions like `check_url.assert_called_once_with()` which are confusing because there is no `check_url()` method defined anywhere in Via (it's a method that Via implicitly _inherits_ from `checkmatelib`):

https://github.com/hypothesis/via/blob/d19c83df23de7b4d67b0d8aba37d218b3b74ade0/tests/unit/via/checkmate_test.py#L28-L33

via/app.py
```diff
@@ -88,9 +87,6 @@ def create_app(_=None, **settings):
     config.add_static_view(name="static", path="via:static")
     config.add_cache_buster("via:static", cachebust=cache_buster)
 
-    # Make the CheckmateClient object available as request.checkmate.
-    config.add_request_method(ViaCheckmateClient, reify=True, name="checkmate")
```
seanh, line 92: No more `request.checkmate`.

I think this is good partially because services are a more familiar pattern for us, and also because it's good to keep custom `request` methods to a minimum: the `request` is kind of a global variable and with `request.checkmate` anything that wants to call Checkmate needs to have the `request` object. Whereas with `CheckmateService` you only need to have the service, for example future PRs will inject `CheckmateService` into other services as a dependency.

via/services/checkmate.py
```diff
@@ -5,13 +5,20 @@
...
-    def __init__(self, request):
-        super().__init__(
-            host=request.registry.settings["checkmate_url"],
-            api_key=request.registry.settings["checkmate_api_key"],
+class CheckmateService:
+    def __init__(self, checkmate_client, allow_all, blocked_for, ignore_reasons):
+        self._checkmate_client = checkmate_client
+        self._allow_all = allow_all
+        self._blocked_for = blocked_for
+        self._ignore_reasons = ignore_reasons
```
seanh, line 16: By using the service pattern the construction of the `CheckmateClient` and `CheckmateService` objects and reading the settings and params that these objects need from the `request` is all moved into the `factory()` function (and its unit tests) which allows the `CheckmateService` class itself to be much simpler: it no longer has access to the `request` and no longer reads `request.registry.settings` and `request.params`, and it's much easier to see the four specific dependencies that it actually has (`checkmate_client`, `allow_all`, `blocked_for` and `ignore_reasons`).

Note that I've changed this from inheritance to composition: instead of inheriting from `Checkmatelib.CheckmateClient` the new `CheckmateService.__init__()` accepts a `CheckmateClient` instance as an argument. This makes things much more clearly decoupled which makes the code easier to cleanly unit test and also easier to understand. Previously you'd see `check_url()` being called both as `self.check_url()` within the old `ViaCheckmateClient` itself and also as `request.checkmate.check_url()` in the `status.py` view, but this was confusing because no `check_url()` was defined anywhere in Via, it's defined in `checkmatelib` but you won't see `check_url()` being explicitly imported anywhere in Via either.

via/services/checkmate.py
```diff
@@ -5,13 +5,20 @@
...
+
+    def check_url(self, url):
+        return self._checkmate_client.check_url(
+            url,
+            allow_all=self._allow_all,
+            blocked_for=self._blocked_for,
+            ignore_reasons=self._ignore_reasons,
```
seanh: Because we're using composition `ViaClientService` has to have an explicit `check_url()` method that forwards to `checkmatelib`'s `CheckmateClient.check_url()`.

As usual with composition this can seem a bit wordier (having to have this boilerplate forwarding method and tests for it) but it's much clearer: you can see what the API of the `CheckmateService` class it by just looking at the code of the class itself, without also having to look at the `CheckmateClient` class (which is in an external library), and no confusion about calls to a `check_url()` method that doesn't seem to be defined anywhere.

This `check_url()` method also isn't *just* forwarding to `CheckmateClient.check_url()`: it's also injecting the `allow_all`, `blocked_for` and `ignore_reasons` values. This means that `raise_if_blocked()` (below) doesn't need to inject those and makes that method shorter and clearer by separating the concerns.

via/services/checkmate.py
```diff
@@ -24,20 +31,23 @@ def raise_if_blocked(self, url):
...
+def factory(_context, request):
+    return CheckmateService(
+        checkmate_client=CheckmateClient(
+            host=request.registry.settings["checkmate_url"],
+            api_key=request.registry.settings["checkmate_api_key"],
+        ),
+        allow_all=request.registry.settings["checkmate_allow_all"],
+        blocked_for=request.params.get("via.blocked_for"),
+        ignore_reasons=request.registry.settings["checkmate_ignore_reasons"],
+    )
```
seanh, line 61: All the `request` stuff grouped together in the factory

via/services/checkmate.py
```diff
@@ -5,13 +5,20 @@
...
-            host=request.registry.settings["checkmate_url"],
-            api_key=request.registry.settings["checkmate_api_key"],
+class CheckmateService:
+    def __init__(self, checkmate_client, allow_all, blocked_for, ignore_reasons):
+        self._checkmate_client = checkmate_client
+        self._allow_all = allow_all
+        self._blocked_for = blocked_for
+        self._ignore_reasons = ignore_reasons
+
+    def check_url(self, url):
```
jon-betts:
```suggestion
    def check_url(self, url): -> Optional[BlockResponse]
```

I think this is handy to add as the block response isn't just a boolean or something, but has structure.

A doc-string here would be good. I'd be in favour of re-introducing doc-strings as a required part of our linting. I don't think missing them out is an exceptional case at the moment.

via/services/checkmate.py
```diff
@@ -24,20 +31,23 @@ def raise_if_blocked(self, url):
...
             raise BadURL(exc, url=url) from exc
-
         except CheckmateException:
             blocked = None
 
         if blocked:
             raise HTTPTemporaryRedirect(location=blocked.presentation_url)
+
+
+def factory(_context, request):
```
jon-betts, line 52:
```suggestion
def service_factory(_context, request):
```

A quick future compatibility tip. I don't think it'll ever get there, but I think we are standardizing on `service_factory` as the name to allow for splitting big services into multiple files.

It's a bit of a nightmare to import `blah.factory` if it's both a method and a module name.

via/views/proxy.py
```diff
@@ -14,7 +14,9 @@ def static_fallback(_context, _request):
 @view_config(route_name="proxy", renderer="via:templates/proxy.html.jinja2")
 def proxy(context, request):
     url = context.url_from_path()
-    request.checkmate.raise_if_blocked(url)
+    checkmate_service = request.find_service(CheckmateService)
+
+    checkmate_service.raise_if_blocked(url)
```
jon-betts:
```suggestion
    request.find_service(CheckmateService).raise_if_blocked(url)
```

Doesn't look like we use this again?

via/views/route_by_content.py
```diff
@@ -4,15 +4,21 @@
...
 
 
 @view.view_config(route_name="route_by_content", decorator=(has_secure_url_token,))
 def route_by_content(context, request):
     """Routes the request according to the Content-Type header."""
     url = context.url_from_query()
+    checkmate_service = request.find_service(CheckmateService)
 
-    request.checkmate.raise_if_blocked(url)
+    checkmate_service.raise_if_blocked(url)
```
jon-betts: Similar suggestion to above, but I can't do it in GH because of the removed line.

tests/unit/via/services/checkmate_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @pytest.fixture
+    def pyramid_request(self, pyramid_request):
+        pyramid_request.registry.settings["checkmate_url"] = sentinel.url
+        pyramid_request.registry.settings["checkmate_api_key"] = sentinel.api_key
+        pyramid_request.registry.settings["checkmate_allow_all"] = sentinel.allow_all
+        pyramid_request.registry.settings[
+            "checkmate_ignore_reasons"
+        ] = sentinel.ignore_reasons
+        return pyramid_request
```
jon-betts, line 124:
```suggestion
    @pytest.mark.parametrize("blocked_for", (None, sentinel.blocked_for))
    def test_it(self, pyramid_request, CheckmateClient, CheckmateService, blocked_for):
        pyramid_request.registry.settings = {
            "checkmate_url": sentinel.url,
            "checkmate_api_key": sentinel.api_key,
            "checkmate_allow_all": sentinel.allow_all,
            "checkmate_ignore_reasons": sentinel.ignore_reasons,
        }
        if blocked_for:
            pyramid_request.params["via.blocked_for"] = sentinel.blocked_for

        svc = factory(sentinel.context, pyramid_request)

        CheckmateClient.assert_called_once_with(
            host=sentinel.url, api_key=sentinel.api_key
        )
        CheckmateService.assert_called_once_with(
            checkmate_client=CheckmateClient.return_value,
            allow_all=sentinel.allow_all,
            blocked_for=blocked_for,
            ignore_reasons=sentinel.ignore_reasons,
        )
        assert svc == CheckmateService.return_value
```

I find the back and forth between the fixture, sentinels and tests a little complicated. I think we can do this as above in a oner.

Another option might be to assert with statements similar to we do with `blah.return_value`:

```python
assert blah == pyramid_request.registry.settings["checkmate_url"]

# vs.

assert blah == sentinel.checkmate_url
```

jon-betts (approved):
This looks good. A service fits here pretty well.

I have some stylistic tips:

 * Docstrings are good, I think the new method should have one
 * I think we can look-up and directly use the service in a oner in all the places I've seen it, instead of assigning to a variable
 * I think we are starting to use `service_factory` instead of `factory` to get around import problems if we ever want to split services up
 * Some suggestions to make the tests a bit shorter

via/services/checkmate.py
seanh: :+1: https://github.com/hypothesis/via/pull/1004/commits/cf2604d2bfd54dd4e46625a7eff2fd87e9177ad9 It's going to be so much nicer when we can write `-> BlockResponse | None` instead of using `Optional` (Python 3.10)

seanh: Added a docstring too: https://github.com/hypothesis/via/pull/1004/commits/dc41000eac2c40de9f8815c2278d42704c842936

via/services/checkmate.py
seanh: I'm gonna leave this because they're actually _all_ called `factory()` not `service_factory()` in Via currently, so perhaps there needs to be a separate PR to rename them all consistently. I suspect splitting things into multiple _services_ would be a better pattern, rather than splitting something into a single big service with multiple files.

via/views/proxy.py
seanh: :+1: https://github.com/hypothesis/via/pull/1004/commits/b9feee694badad46fe59d6600da8bd33b91937ba

via/views/route_by_content.py
seanh: :+1: https://github.com/hypothesis/via/pull/1004/commits/131a64b66c923a4a8a4e53c56f06e38c84e592e3

tests/unit/via/services/checkmate_test.py
seanh: The registry settings etc are in a fixture because there's more than one test for the factory, so if you don't put the set up in a shared fixture you end up having to duplicate it in each test.

I actually don't mind the pattern of putting test _setup_ in fixtures but then putting the test logic itself in the tests: it nicely separates setup and test and can make the test logic a lot clearer because the tests are stuffed full of distracting setup code. As long as it's done clearly with simple fixtures and with the fixtures sticking to setup and not containing test logic.

And I think when using clear sentinel values it's fairly obvious:

```python
CheckmateService.assert_called_once_with(
    ...
    allow_all=sentinel.allow_all_setting,
    ...
)
```

You may not be able to see exactly where it comes from without scanning the rest of the test class for `sentinel.allow_all_setting`, but doesn't the name alone makes it pretty clear that this is a sentinel value standing in for the `allow_all` setting? And it's eminently greppable too.

(This suggests that maybe longer names like `sentinel.allow_all_setting` might be better than just `sentinel.allow_all`, but the less-specific `sentinel.allow_all` name is one that wouldn't have to change if we e.g. moved the value from a setting to a request param or something.)

jon-betts: >so if you don't put the set up in a shared fixture you end up having to duplicate it in each test.

The suggestion above replaces the two tests with a single parameterized test

PR #1006, opened 2023-06-22
Fetch video transcript from API instead of using hard-coded data

robertknight:
Use the API added in https://github.com/hypothesis/via/pull/1000 to fetch video transcripts, instead of hard-coded data. There is currently a known issue where there is a race condition between the client loading and the transcript loading. If the client loads first, it may cause existing annotations to fail to anchor. I will be working to fix that after this lands.

**Summary of changes:**

 - Copy `waitForElement` utility from lms repo, and convert to TypeScript
 - Add types for API responses from the server and `callAPI` utility to make
   calls to API.
 - Change `VideoPlayerApp` component to accept either a pre-fetched transcript
   or API call configuration as the transcript source. If passed an API call
   config, it displays a loading state, performs the API call and then displays
   the response.

   In the actual app an API call is always used, but the pre-fetched transcript
   is useful in tests.

 - Change video player entry point to read API index from JSON configuration and
   pass the config for the transcript API call to `VideoPlayerApp`.

**Testing:**

1. Go to http://localhost:9083/video/youtube?url=https://www.youtube.com/watch?v=x8TO-nrUtSI. The transcript should load as before.
2. Modify the `callAPI` function in `api.ts` to add a delay and/or fail, then refresh the page. You will see a loading spinner followed by a gloriously unstyled error state.

```ts
export async function callAPI<T = unknown>(api: APIMethod): Promise<T> {
  const headers = { ...api.headers };
  headers['Content-Type'] = 'application/json; charset=UTF-8';

  // Added - Simulate delay and eventual error.
  await new Promise(resolve => setTimeout(resolve, 2000));
  throw new APIError(400);
  ...
}
```

Screenshot 2023-06-27 at 13 10 38

Transcript error state

### DIFF
--- a/via/static/scripts/test-util/wait.ts
+++ b/via/static/scripts/test-util/wait.ts
@@ -7,7 +7,7 @@ export async function waitFor<T>(
   condition: () => T,
   timeout = 10,
   what = condition.toString()
-): Promise<T> {
+): Promise<NonNullable<T>> {
   const result = condition();
   if (result) {
     return result;
@@ -28,6 +28,41 @@ export async function waitFor<T>(
       }
     });
   });
+}
+
+// Minimal Enzyme types needed by `waitForElement`.
+
+// eslint-disable-next-line no-use-before-define
+type Predicate = (wrapper: EnzymeWrapper) => boolean;
+
+type EnzymeWrapper = {
+  length: number;
+  update(): void;
+  find(query: string | Predicate): EnzymeWrapper;
+};
+
+/**
+ * Wait up to `timeout` ms for an element to be rendered.
+ *
+ * @param selector - Selector string or function to pass to `wrapper.find`
+ */
+export function waitForElement(
+  wrapper: EnzymeWrapper,
+  selector: string | Predicate,
+  timeout = 10
+): Promise<EnzymeWrapper> {
+  return waitFor(
+    () => {
+      wrapper.update();
+      const el = wrapper.find(selector);
+      if (el.length === 0) {
+        return null;
+      }
+      return el;
+    },
+    timeout,
+    `"${selector}" to render`
+  );
 }
 
 export function delay(ms: number) {
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -41,7 +41,7 @@ export type VideoPlayerAppProps = {
 
   /**
    * The data source for the transcript. Either an API to call when the player
-   * loads, or pre-fetched data.
+   * loads, or pre-fetched data (mostly useful in tests).
    */
   transcriptSource: APIMethod | TranscriptData;
 };
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -5,6 +5,7 @@ import {
   IconButton,
   Input,
   LogoIcon,
+  Spinner,
 } from '@hypothesis/frontend-shared';
 import classnames from 'classnames';
 import type { Ref } from 'preact';
@@ -17,9 +18,11 @@ import {
 } from 'preact/hooks';
 
 import { useAppLayout } from '../hooks/use-app-layout';
+import { callAPI } from '../utils/api';
+import type { APIMethod, APIError, JSONAPIObject } from '../utils/api';
 import { useNextRender } from '../utils/next-render';
 import type { TranscriptData } from '../utils/transcript';
-import { formatTranscript } from '../utils/transcript';
+import { formatTranscript, mergeSegments } from '../utils/transcript';
 import HypothesisClient from './HypothesisClient';
 import Transcript from './Transcript';
 import type { TranscriptControls } from './Transcript';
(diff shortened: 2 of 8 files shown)


via/static/scripts/test-util/wait.ts
```diff
@@ -7,7 +7,7 @@ export async function waitFor<T>(
   condition: () => T,
   timeout = 10,
   what = condition.toString()
-): Promise<T> {
+): Promise<NonNullable<T>> {
```
robertknight, line 10: The `waitFor` function keeps looping until the `condition` function returns non-null. Encode this in the types.

via/static/scripts/test-util/wait.ts
```diff
@@ -30,6 +30,41 @@ export async function waitFor<T>(
   });
 }
 
+// Minimal Enzyme types needed by `waitForElement`.
```
robertknight, line 33: Enzyme doesn't have its own types, and the third-party types pull in all of React's types. These minimal definitions are good enough for now. I'm happy to write Preact-oriented types for the parts of Enzyme we use when we need them.

via/static/scripts/test-util/wait.ts
```diff
@@ -30,6 +30,41 @@ export async function waitFor<T>(
...
+      const el = wrapper.find(selector);
+      if (el.length === 0) {
+        return null;
+      }
+      return el;
+    },
+    timeout,
+    `"${selector}" to render`
+  );
+}
```
robertknight, line 66: This module should be moved to a shared package soon-ish, since these utilities are clearly generally useful.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -382,24 +428,36 @@ export default function VideoPlayerApp({
...
+                    'bg-gradient-to-r from-white to-grey-1 border-l w-[40px]'
+                  )}
+                />
+              </Transcript>
+            )}
+            {transcript instanceof Error && (
+              <div data-testid="transcript-error">
+                Unable to load transcript: {transcript.message}
+              </div>
+            )}
```
robertknight, line 460: This component is starting to get a little on the large side. We should start to think about factoring out some smaller pieces soonish.

lyzadanger: Yes.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -382,24 +428,36 @@ export default function VideoPlayerApp({
...
-                  // is overlaid by the sidebar's semi-transparent bucket
-                  // channel.
-                  'bg-gradient-to-r from-white to-grey-1 border-l w-[40px]'
-                )}
-              />
-            </Transcript>
+            {isLoading && (
+              <div className="flex justify-center p-8">
+                <Spinner data-testid="transcript-loading-spinner" size="md" />
+              </div>
```
lyzadanger, line 434: This spinner vis-a-vis the height of the transcript column/region needs a little attention, but needs to be done in context with https://github.com/orgs/hypothesis/projects/118/views/1?pane=issue&itemId=31604133

I can pick this up on my next pass.

via/static/scripts/video_player/utils/api.ts
lyzadanger, line 1: I appreciate how simple and clean this is (i.e. compared to, say, the API utils in LMS). Is this fresh-slate or did you crib any of this from elsewhere?

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
```diff
@@ -34,14 +42,15 @@ describe('VideoPlayerApp', () => {
...
-  // TODO - Convert existing tests to use this helper to render the player.
   function createVideoPlayer(props = {}) {
     const wrapper = mount(
       <VideoPlayerApp
         videoId="1234"
         clientSrc="https://dummy.hypothes.is/embed.js"
         clientConfig={{}}
-        transcript={transcriptData}
+        // By default start with the transcript already loaded, instead of
+        // using the API to fetch it.
```
lyzadanger, line 52: Glad for this comment.

via/static/scripts/video_player/index.tsx
```diff
@@ -16,28 +13,18 @@ export function init() {
...
-  // Group segments together for better readability.
-  transcript.segments = mergeSegments(transcript.segments, 3);
-
   render(
     <VideoPlayerApp
       videoId={videoId}
       clientConfig={clientConfig}
       clientSrc={clientSrc}
-      transcript={transcript}
+      transcriptSource={api.transcript}
```
lyzadanger, line 27: Let me ensure I've got this right: `VideoPlayerApp`'s `transcriptSource` prop is typed `APIMethod | TranscriptData`, but the app's entry point will only ever deal in `APIMethod`. Is the `TranscriptData` type allowed entirely as a test affordance or do you anticipate it being more generally useful?

_Nevermind_, the PR description answers this question:

> In the actual app an API call is always used, but the pre-fetched transcript is useful in tests.

Given my distraction here, it might be useful to add a comment somewhere to this effect. Maybe?

lyzadanger (approved):
👍🏻 This adds a bunch of functionality but it's really clear, so there are few comments.

via/static/scripts/video_player/utils/api.ts
robertknight: This is fresh. It follows the pattern of the LMS app utilities, although those have some added complexity due to needing to support retries etc.

PR #1007, opened 2023-06-22
Add caching to `CheckmateService.check_url()`

seanh:
This is so that different parts of the Via code (e.g. different views and services) can blindly call `CheckmateService.check_url()` whenever they need to without worrying about whether something else somewhere else in the code has already called `check_url()` in the current request for the same URL: you can call it multiple times and it won't actually make multiple requests to Checkmate.

This is going to make some upcoming behaviour changes simpler to implement.

### DIFF
--- a/via/services/checkmate.py
+++ b/via/services/checkmate.py
@@ -1,3 +1,5 @@
+from functools import lru_cache
+
 from checkmatelib import BadURL as CheckmateBadURL
 from checkmatelib import CheckmateClient, CheckmateException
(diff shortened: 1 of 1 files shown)


seanh:
My only concern with this is the question of whether it's safe to use `@lru_cache` on instance methods? Do we need to cache results on `self` manually ourselves instead?

seanh:
I think I'm probably not going to need this after all. I'll re-open this if I need it...

PR #1008, opened 2023-06-22
Make `ViaClientService.url_for()`'s `content_type` dependency public

seanh:
`ViaClientService.url_for()` privately depends on the `ContentType`
associated with the given MIME type - `ViaClientService` calls its own
`content_type()` method to get the `ContentType` for the given
`mime_type`:

    def url_for(self, url, mime_type, params):
        return self.via_client.url_for(
            url,
            content_type=self.content_type(mime_type),
            options=params,
        )

In the future code that uses `ViaClientService` is going to have to
treat URLs differently based on their content-type. For example YouTube
URLs are going to be treated differently. Make `content_type` an
explicit argument to `url_for()` so that in future PRs we can write
logic like this:

    content_type = via_client_svc.content_type(mime_type)

    if content_type = ContentType.YOUTUBE:
        # Do something.
    else:
        # Do something else.

    url = via_client_svc.url_for(url, content_type, ...)

    # Do something with `url`.

### DIFF
--- a/tests/unit/via/services/via_client_test.py
+++ b/tests/unit/via/services/via_client_test.py
@@ -19,21 +19,11 @@ class TestViaClientService:
     def test_content_type(self, mime_type, expected_content_type, svc):
         assert svc.content_type(mime_type) == expected_content_type
 
-    @pytest.mark.parametrize(
-        "mime_type,expected_content_type",
-        [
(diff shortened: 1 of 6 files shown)


seanh:
I think I'm probably not going to go in this direction after all. I'll re-open this if I need it...

PR #1009, opened 2023-06-26
Don't call Checkmate for YouTube URLs

seanh:
This should enable us to enable the new YouTube feature in QA and production by allowing it to bypass Checkmate's YouTube block.

This PR does not enable particular YouTube URLs to still be blocked, but it does refactor the code into a shape that should make that easy to achieve. There's a separate issue for adding it: https://github.com/hypothesis/via/issues/978

## Problem

We've added a new YouTube player/annotator page to Via at `/video/youtube?url={youtube_url}` but this doesn't work in QA and production because all of YouTube is blocked by Checkmate.

[The video view itself](https://github.com/hypothesis/via/blob/eccd9f45178a89d541cfc77b77dbc9b8c308d491/via/views/view_video.py#L13-L44) doesn't actually call Checkmate but [the `proxy()` view](https://github.com/hypothesis/via/blob/eccd9f45178a89d541cfc77b77dbc9b8c308d491/via/views/proxy.py#L14-L27) (called when you paste a YouTube URL into Via's front page) and [the `route_by_content()` view](https://github.com/hypothesis/via/blob/eccd9f45178a89d541cfc77b77dbc9b8c308d491/via/views/route_by_content.py#L10-L32) (called when launch a YouTube assignment in the LMS app) both do call Checkmate, and so these views end up showing a Checkmate error rather than redirecting to or embedding the `video()` view.

The reason why YouTube is blocked is that, prior to the new YouTube feature, YouTube HTML pages would get handled by Via HTML because it tries to proxy the actual bytes of the videos through our servers and this blocks up the service.

With the new YouTube feature it's now okay to allow YouTube pages _that will be handled by the new feature_ to be proxied but the new feature only handles YouTube pages that pertain to a single video and not the YouTube front page, channel or user pages, etc. Those other YouTube pages would still get sent to Via HTML and so they must still remain blocked.

## Solution

To solve this we need to:

1. Not call Checkmate anymore for URLs that would be handled by the new YouTube page
2. Still call Checkmate as before for other URLs, including other YouTube URLs

To achieve this, this commit moves the calls to Checkmate from the `proxy()` and `route_by_content()` views into `URLDetailsService.get_url_details()`, which they both call.
This enables us to no longer call Checkmate in the case of YouTube URLs, while still always calling Checkmate before sending any requests to get the content-types and status codes of any other URLs.

### DIFF
--- a/via/services/url_details.py
+++ b/via/services/url_details.py
@@ -3,13 +3,20 @@ import cgi
 from collections import OrderedDict
 
 from via.requests_tools.headers import add_request_headers, clean_headers
+from via.services.checkmate import CheckmateService
 from via.services.google_drive import GoogleDriveAPI
 from via.services.http import HTTPService
 from via.services.youtube import YouTubeService
 
 
 class URLDetailsService:
-    def __init__(self, http_service: HTTPService, youtube_service: YouTubeService):
+    def __init__(
+        self,
+        checkmate_service: CheckmateService,
+        http_service: HTTPService,
+        youtube_service: YouTubeService,
+    ):
+        self._checkmate = checkmate_service
         self._http = http_service
         self._youtube = youtube_service
 
@@ -21,19 +28,19 @@ class URLDetailsService:
         :return: 2-tuple of (mime type, status code)
 
         :raise BadURL: When the URL is malformed
+        :raise checkmatelib.BadURL: if the URL is blocked by Checkmate
         :raise UpstreamServiceError: If we server gives us errors
         :raise UnhandledException: For all other request based errors
         """
-        if headers is None:
-            headers = OrderedDict()
+        if self._youtube.enabled and self._youtube.get_video_id(url):
+            return "video/x-youtube", 200
+
+        self._checkmate.raise_if_blocked(url)
 
         if GoogleDriveAPI.parse_file_url(url):
             return "application/pdf", 200
 
-        if self._youtube.enabled and self._youtube.get_video_id(url):
-            return "video/x-youtube", 200
-
-        headers = add_request_headers(clean_headers(headers))
+        headers = add_request_headers(clean_headers(headers or OrderedDict()))
 
         with self._http.get(
             url,
@@ -54,6 +61,7 @@ class URLDetailsService:
 
 def factory(_context, request):
     return URLDetailsService(
+        checkmate_service=request.find_service(CheckmateService),
         http_service=request.find_service(HTTPService),
         youtube_service=request.find_service(YouTubeService),
     )
--- a/via/services/url_details.py
+++ b/via/services/url_details.py
@@ -27,10 +27,10 @@ class URLDetailsService:
         :param headers: The original headers the request was made with
         :return: 2-tuple of (mime type, status code)
 
-        :raise BadURL: When the URL is malformed
+        :raise BadURL: if the URL is malformed
         :raise checkmatelib.BadURL: if the URL is blocked by Checkmate
-        :raise UpstreamServiceError: If we server gives us errors
-        :raise UnhandledException: For all other request based errors
+        :raise UpstreamServiceError: if we get an error from the upstream server
+        :raise UnhandledException: if we get any other request-based error
         """
         if self._youtube.enabled and self._youtube.get_video_id(url):
             return "video/x-youtube", 200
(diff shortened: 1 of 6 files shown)


via/services/url_details.py
```diff
@@ -21,19 +28,19 @@ def get_url_details(self, url, headers=None):
...
 
         :raise BadURL: When the URL is malformed
+        :raise checkmatelib.BadURL: if the URL is blocked by Checkmate
         :raise UpstreamServiceError: If we server gives us errors
         :raise UnhandledException: For all other request based errors
         """
-        if headers is None:
-            headers = OrderedDict()
+        if self._youtube.enabled and self._youtube.get_video_id(url):
+            return "video/x-youtube", 200
```
seanh: Moved the YouTube check up to the top of the method because this is the one thing that we want to do _before_ calling Checkmate.

Inside this YouTube `if` is where we could do a _different_ call to Checkmate instead, with different ignore reasons, if we wanted to still allow some YouTube videos to be specifically blocked. But that would also require some changes in other files, so that's for a separate PR.

via/services/url_details.py
```diff
@@ -21,19 +28,19 @@ def get_url_details(self, url, headers=None):
...
+
+        self._checkmate.raise_if_blocked(url)
```
seanh, line 38: If it's not YouTube then there's no point in proceeding any further without calling Checkmate first

via/services/url_details.py
```diff
@@ -21,19 +28,19 @@ def get_url_details(self, url, headers=None):
         :return: 2-tuple of (mime type, status code)
 
         :raise BadURL: When the URL is malformed
+        :raise checkmatelib.BadURL: if the URL is blocked by Checkmate
         :raise UpstreamServiceError: If we server gives us errors
         :raise UnhandledException: For all other request based errors
         """
-        if headers is None:
-            headers = OrderedDict()
```
seanh, line 28: Merged this `if` into the `headers = ` line below

via/services/url_details.py
```diff
@@ -21,19 +28,19 @@ def get_url_details(self, url, headers=None):
...
+        self._checkmate.raise_if_blocked(url)
 
         if GoogleDriveAPI.parse_file_url(url):
             return "application/pdf", 200
 
-        if self._youtube.enabled and self._youtube.get_video_id(url):
-            return "video/x-youtube", 200
-
-        headers = add_request_headers(clean_headers(headers))
+        headers = add_request_headers(clean_headers(headers or OrderedDict()))
```
seanh: Merged the initialization of `headers` from above (`if headers is None: ...`) into this assignment

jon-betts: With new Python, do we need `OrderedDict` any more?

via/services/url_details.py
```diff
@@ -21,19 +28,19 @@ def get_url_details(self, url, headers=None):
         :return: 2-tuple of (mime type, status code)
 
         :raise BadURL: When the URL is malformed
+        :raise checkmatelib.BadURL: if the URL is blocked by Checkmate
         :raise UpstreamServiceError: If we server gives us errors
```
jon-betts: "If we server"?

via/services/url_details.py
jon-betts: :+1: Could go one further and inline it

jon-betts (approved):
Yep. This looks good. I spotted some weird doc-string from before.

via/services/url_details.py
seanh: :+1: https://github.com/hypothesis/via/pull/1009/commits/79fd557a9c0adbf895f7cf11299276ff3235367b

via/services/url_details.py
seanh: Good question. Normal dicts now keep the key insertion order since ages ago (Python 3.6) so I think that would have already been true when this code was originally written (Via was never on Python <3.6), so I'm not sure why `OrderedDict` was originally used here.

There are still some differences between dicts and `OrderedDict`'s. For example `OrderedDict`'s consider the order in equality comparisions, dicts don't. `OrderedDict.popitem()` has a new optional argument. `OrderedDict` has `move_to_end()`.

I think I'd rather not mess with it in this PR: without spending more time looking I'm not sure it won't accidentally change some behaviour

via/services/url_details.py
seanh: Fixed this up: https://github.com/hypothesis/via/pull/1009/commits/aeebd90aea2fc3ae4f879cc648d54a751893bf83

PR #1012, opened 2023-06-27
Defer loading client until transcript is loaded

robertknight:
This avoids annotations orphaning if the client tries to anchor before the content is loaded. See https://github.com/hypothesis/client/issues/5568.


lyzadanger (approved):
👍🏻 Seems like a sensible workaround for the moment.

Issue #941, opened 2023-04-26
Capture appropriate metadata with video transcript annotations


robertknight:
As part of https://github.com/hypothesis/product-backlog/issues/1433 we'll need to decide what metadata should be captured when annotating video transcripts, both for the video as a whole and for the specific part of a video that an annotation refers to.

In the existing YouTube video annotation prototype in [docdrop](https://docdrop.org), the Hypothesis client just sees the transcript as a regular web page and annotations of the transcript work like annotating text on any other page. I think we are going to want to capture the timestamp and video length as metadata, so we can display that information in the sidebar/notebook or other places.

Issues to resolve:

- Decide what the annotation's main URL should be
- Decide what document-level metadata should be captured (eg. video title, channel, video length, other?)
- Decide what annotation-level metadata should be captured (eg. start/end timestamp)
- Figure out the mechanisms for the [video player app](https://github.com/hypothesis/via/issues/904) to convey the document-level metadata to the client 
- Figure out the mechanisms for the video player app to convey annotation-level metadata to the client
- Decide if we're going to display any of the video-oriented metadata (eg. timestamps) in the sidebar

Sub-tasks:

- [x] Capture YouTube video URL as annotation URL (see https://github.com/hypothesis/via/pull/1015)
- [x] Capture video title as document title (see https://github.com/hypothesis/via/issues/1071)
- [x] Capture timestamp selectors with annotations (https://github.com/hypothesis/client/pull/5655, https://github.com/hypothesis/via/pull/1116)

robertknight:
I moved this to Via repo because that is where the video player implementation currently lives.

robertknight:
Related to this, we've been discussing showing the video player when proxying a YouTube video URL through Via (`https://via.hypothes.is/https://www.youtube.com/watch?v=abc`), instead of a custom URL (`https://via.hypothes.is/video/abc`). If we do this, and we use eg. `https://www.youtube.com/watch?v=abc` as the URL for a video transcript annotation, we'll have a need to disambiguate from annotations made using eg. the browser extension accessing that URL as a normal web page.

The web annotations specs have a notion of [Rendering Software](https://www.w3.org/TR/annotation-model/#rendering-software) and a `renderedVia` attribute for annotation targets as a way to indicate what tool was used to render a URL. So one option would be to leverage that somehow. There is also a notion of [resource states](https://www.w3.org/TR/annotation-model/#states) as a way to indicate which format of a resource was accessed.

[...]

PR #1015, opened 2023-06-28
Set URL for videos to canonical YouTube video URL

robertknight:
- Add a `<link rel=canonical>` to the video transcript page, which the Hypothesis client will use as the URL for annotations.

 - For YouTube videos, set this URL to `https://www.youtube.com/watch?v={video_id}`.  This is the URL that the various other URLs (`https://youtube.com?v={video_id}`, `https://youtu.be/{video_id}`, `https://youtube.com/v/{video_id}`) ultimately redirect to and is what is used as the `<link rel=canonical>` on YouTube video pages.

----

**Testing:**

1. Go to http://localhost:9083/https://www.youtube.com/watch?v=x8TO-nrUtSI or any of the other recognized YouTube URLs (eg. http://localhost:9083/https://youtu.be/x8TO-nrUtSI)
2. Click the "?" button at the top of the client sidebar and go to "About this version"
3. The document URL displayed in the client should be "https://www.youtube.com/watch?v=x8TO-nrUtSI". There should be a `<link rel=canonical>` tag in the inner iframe's `<head>` with this same value.

If the URL contains additional parameters, such as `t={seconds}` to set the start time, these should *not* appear in the canonical URL.

### DIFF
--- a/tests/unit/via/views/view_video_test.py
+++ b/tests/unit/via/views/view_video_test.py
@@ -23,7 +23,7 @@ class TestViewVideo:
         video_url,
         ViaSecurityPolicy,
     ):
-        youtube_service.canonical_video_url.return_value = sentinel.canonical_video_url
+        # Override default `None` response.
         youtube_service.get_video_id.return_value = sentinel.youtube_video_id
 
         response = youtube(pyramid_request)
@@ -39,8 +39,8 @@ class TestViewVideo:
         assert response == {
             "client_embed_url": "http://hypothes.is/embed.js",
             "client_config": Configuration.extract_from_params.return_value[1],
-            "video_id": sentinel.youtube_video_id,
-            "video_url": sentinel.canonical_video_url,
+            "video_id": youtube_service.get_video_id.return_value,
+            "video_url": youtube_service.canonical_video_url.return_value,
             "api": {
                 "transcript": {
                     "doc": Any.string(),
--- a/tests/unit/via/views/view_video_test.py
+++ b/tests/unit/via/views/view_video_test.py
@@ -23,11 +23,15 @@ class TestViewVideo:
         video_url,
         ViaSecurityPolicy,
     ):
+        youtube_service.canonical_video_url.return_value = sentinel.canonical_video_url
         youtube_service.get_video_id.return_value = sentinel.youtube_video_id
 
         response = youtube(pyramid_request)
 
         youtube_service.get_video_id.assert_called_once_with(video_url)
+        youtube_service.canonical_video_url.assert_called_once_with(
+            sentinel.youtube_video_id
+        )
         Configuration.extract_from_params.assert_called_once_with(
             {"via.foo": "foo", "via.bar": "bar"}
         )
@@ -36,6 +40,7 @@ class TestViewVideo:
             "client_embed_url": "http://hypothes.is/embed.js",
             "client_config": Configuration.extract_from_params.return_value[1],
             "video_id": sentinel.youtube_video_id,
+            "video_url": sentinel.canonical_video_url,
             "api": {
                 "transcript": {
                     "doc": Any.string(),
--- a/via/services/youtube.py
+++ b/via/services/youtube.py
@@ -18,9 +18,6 @@ class YouTubeService:
         This is used as the URL which YouTube transcript annotations are
         associated with.
         """
-        if not video_id:
-            raise ValueError("Invalid video ID")
-
         escaped_id = quote_plus(video_id)
         return f"https://www.youtube.com/watch?v={escaped_id}"
 
--- a/via/services/youtube.py
+++ b/via/services/youtube.py
@@ -1,4 +1,4 @@
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, quote_plus, urlparse
 
 from youtube_transcript_api import YouTubeTranscriptApi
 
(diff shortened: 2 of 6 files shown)


tests/unit/via/views/view_video_test.py
```diff
@@ -23,11 +23,15 @@ def test_it(
         video_url,
         ViaSecurityPolicy,
     ):
+        youtube_service.canonical_video_url.return_value = sentinel.canonical_video_url
```
seanh: You probably don't need to do this, you can just refer to `youtube_service.canonical_video_url.return_value` in the code below. Mocks already return unique values out of the box so unless the sentinel name is really adding anything to the code it's unnecessary

via/services/youtube.py
```diff
@@ -11,6 +11,19 @@ def __init__(self, enabled: bool):
...
 
+    def canonical_video_url(self, video_id: str) -> str:
+        """
+        Return the canonical URL for a YouTube video.
+
+        This is used as the URL which YouTube transcript annotations are
+        associated with.
+        """
+        if not video_id:
+            raise ValueError("Invalid video ID")
```
seanh: I wouldn't bother with this `if`: the `quote_plus()` call below will raise `TypeError` if `video_id` is `None` (and perhaps also if `video_id` is invalid in other ways?). You could catch that `TypeError` and re-raise it as a `ValueError` if you want but given that nothing catches either I wouldn't bother.

We don't generally add validation code like this to our Python methods that're just called by other Python methods of ours. If you're worried about invalid video IDs coming from user-supplied data the place to validate that is at the edge: in a webargs schema on whatever views accept a `video_id` param, schema validation failures will trigger proper error responses

via/services/youtube.py
```diff
@@ -11,6 +11,19 @@ def __init__(self, enabled: bool):
     def enabled(self):
         return self._enabled
 
+    def canonical_video_url(self, video_id: str) -> str:
+        """
+        Return the canonical URL for a YouTube video.
+
+        This is used as the URL which YouTube transcript annotations are
+        associated with.
+        """
```
seanh, line 20: Is this necessary?

I understand that `view_video.html.jinja2` needs to have the `<link rel="canonical" href="{{ video_url }}"/>` but I think you could just pass the non-normalized URL to the template, in `view_video.py::youtube()`:

```diff
diff --git a/via/views/view_video.py b/via/views/view_video.py
index 4c0e4b0..48770da 100644
--- a/via/views/view_video.py
+++ b/via/views/view_video.py
@@ -31,6 +31,7 @@ def youtube(request, url, **kwargs):
         "client_embed_url": request.registry.settings["client_embed_url"],
         "client_config": client_config,
         "video_id": video_id,
+        "video_url": url,
         "api": {
             "transcript": {
                 "doc": "Get the transcript of the current video",
```

I don't think it's Via's job to normalize / canonicalize YouTube URLs? Surely that's the job of the URL normalization and document equivalence code in h? Via isn't the only way that people might annotate YouTube videos and pages, after all.

seanh (commented):
I'm wondering if this should just use the non-normalized YouTube URL as the canonical URL, and do YouTube URL normalization with the rest of the URL normalization in h?

robertknight:
> I'm wondering if this should just use the non-normalized YouTube URL as the canonical URL, and do YouTube URL normalization with the rest of the URL normalization in h?

The URL normalization algorithm in h is effectively frozen at present. See https://github.com/hypothesis/h/issues/6552 (scroll down to bottom for my most recent thinking). I would like to resolve that, but for now I absolutely think using the canonical URL in the video player is the pragmatic option.

via/services/youtube.py
robertknight: > I don't think it's Via's job to normalize / canonicalize YouTube URLs?

For web pages, we do allow and encourage web page authors to indicate what the preferred version of a URL is using `<link rel=canonical>`, which originated as a hint for search engines, and YouTube does set this themselves. They also redirect the various URLs above to this canonical URL.

> Via isn't the only way that people might annotate YouTube videos and pages, after all.

When a user visits YouTube directly, YouTube includes this same `<link rel=canonical>` link on their pages.

tests/unit/via/views/view_video_test.py
robertknight: I followed the pattern of the existing code here (see `sentinel.youtube_video_id` below), although it did look non-idiomatic to me. Is there a reason why the video ID has to be done that way that I've overlooked?

via/services/youtube.py
robertknight: My thinking here was to make it slightly easier to debug if an empty string gets passed. I'm fine with just removing this though.

tests/unit/via/views/view_video_test.py
seanh: Yeah, just try removing it and see what breaks, it's [set to `None` by default](https://github.com/hypothesis/via/blob/a754bdef4f83ab9f924b991209fb19849a1e0962/tests/unit/services.py#L67). It probably shouldn't be set to `None` by default though as that's a bit surprising/unexpected: whatever tests need it to be `None` should probably set that in those particular tests

via/services/youtube.py
seanh: Yeah, I'd say that empty-string video IDs shouldn't be getting into the code in the first place. If they are then we're missing some validation on some view(s)

tests/unit/via/views/view_video_test.py
robertknight: > Yeah, just try removing it and see what breaks,

I did this. It looks like this was added to the mocks so that the default behavior of any tests using it are that a URL is not a YouTube URL. Without this, a bunch of other tests started failing and would have needed to set up an override.

I have removed just the sentinel added in this test, and added a note about why the other one still exists.

via/services/youtube.py
robertknight: OK, I have removed the `video_id` check here.

seanh:
> > I'm wondering if this should just use the non-normalized YouTube URL as the canonical URL, and do YouTube URL normalization with the rest of the URL normalization in h?
> 
> The URL normalization algorithm in h is effectively frozen at present. See [hypothesis/h#6552](https://github.com/hypothesis/h/issues/6552) (scroll down to bottom for my most recent thinking). I would like to resolve that, but for now I absolutely think using the canonical URL in the video player is the pragmatic option.

Ok, this sounds fine to me :+1:

seanh (approved):

PR #1017, opened 2023-06-28
Move API error handling into exception view

seanh:
This arguably doesn't have any benefit whilst there's only one view using it but it gets the code into the right shape: services and views (and their tests) don't need to bother with exception handling or error responses at all, that's all handled by an exception view (which will apply to all views).

### DIFF
--- a/via/views/api/exceptions.py
+++ b/via/views/api/exceptions.py
@@ -1,10 +1,45 @@
-from pyramid.view import forbidden_view_config, notfound_view_config, view_defaults
+import logging
+
+from h_pyramid_sentry import report_exception as report_exception_to_sentry
+from pyramid.view import (
+    exception_view_config,
+    forbidden_view_config,
+    notfound_view_config,
+    view_defaults,
+)
+
+logger = logging.getLogger(__name__)
 
 
 @view_defaults(path_info="^/api/.*", renderer="json")
 class APIExceptionViews:
-    def __init__(self, request):
+    def __init__(self, context, request):
+        self.context = context
         self.request = request
+
+    @exception_view_config(Exception)
+    def exception(self):
+        report_exception_to_sentry(self.context)
+        logger.exception(self.context)
+
+        self.request.response.status_int = 500
+
+        return {
+            "errors": [
+                {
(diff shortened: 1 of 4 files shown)


via/views/api/exceptions.py
```diff
@@ -1,11 +1,46 @@
...
+        return {
+            "errors": [
+                {
+                    "status": self.request.response.status_int,
+                    "code": self.context.__class__.__name__,
+                    "title": str(
+                        getattr(
+                            self.context,
+                            "cause",
+                            "Something went wrong",
```
seanh, line 36: The generic fallback error message [used to be](https://github.com/hypothesis/via/blob/bd1241cab4bbef9ef222c5f82fe1270e2b0cd02d/via/views/api/youtube.py#L36) view-specific (`"Failed to get transcript from YouTube"`) but now it has to be fully generic.

There might be a way for each view to specify a fallback error message for the exception to use but I think we want to avoid every view having to catch and re-raise exceptions as that would lose part of the benefit of exception views (that views and their tests don't need to bother with exception handling). Perhaps something in the `@view_config`?

To be honest I'm not sure `"Failed to get transcript from YouTube"` is actually any more useful than `"Something went wrong"` anyway, so I suspect the best answer may be to just leave this as it is.

If ever the exception does have a `cause` string that will be used, and any service or view can feel free to catch any exception, add a `cause`, and re-raise it

via/views/api/youtube.py
```diff
@@ -19,26 +18,7 @@ def get_transcript(request):
...
-                    "status": request.response.status_int,
-                    "code": exc.__class__.__name__,
-                    "title": str(
-                        getattr(exc, "cause", "Failed to get transcript from YouTube")
-                    ),
-                    "detail": str(exc).strip(),
-                }
-            ]
-        }
+    transcript = request.find_service(YouTubeService).get_transcript(video_id)
```
jon-betts, line 21: Leaves this very clean now

jon-betts (approved):
I've tested a success and failure case, and it works fine.

This certainly clears up the view a lot. We'll see the benefit more as we add more API end-points (if we do).

PR #1018, opened 2023-06-28
Fix `make template`

seanh:
Oops: some changes have been made directly to generated files rather
than using `.cookiecutter/includes/`. Fix this so that `make template`
doesn't make any unwanted changes.


jon-betts (approved):
`make template` is clean now

PR #1020, opened 2023-06-28
Make video player report location and title to top frame

### LINKED ISSUE #991: Red "loading bar" showing forever in video-annotation app

robertknight:
This causes the tab title to update and the loading indicator to hide when the video player app has loaded. This matches code in ViaHTML and `pdf_wrapper.html.jinja2`.

Fixes https://github.com/hypothesis/via/issues/991

### DIFF
--- a/via/static/scripts/video_player/index.tsx
+++ b/via/static/scripts/video_player/index.tsx
@@ -19,6 +19,27 @@ export function init() {
     video_id: videoId,
   } = readConfig();
 
+  // When content is displayed in an iframe, notify top-level of title and
+  // location. This also acts to tell the top-level frame that we finished
+  // loading. See proxy.html.jinja2.
+  if (window !== window.top) {
+    window.parent.postMessage(
+      {
+        type: 'metadatachange',
+        metadata: {
+          // TODO: Once https://github.com/hypothesis/via/pull/1015 lands,
+          // get the URL from `link[rel=canonical]` or pass the video URL as
+          // config.
+          location: `https://www.youtube.com/watch?v=${encodeURIComponent(
+            videoId
+          )}`,
+          title: document.title,
+        },
+      },
+      '*'
+    );
+  }
+
   render(
     <VideoPlayerApp
       videoId={videoId}


lyzadanger:
> This causes the tab title to update and the loading indicator to hide when the video player app has loaded. This matches code in ViaHTML and `pdf_wrapper.html.jinja2`.

Sort of a note to self: this `metadatachange` message is handled in JS code in `proxy.html.jinja2`.

_Edited_: Yes, and this is captured nicely in a comment in the code change itself.

via/static/scripts/video_player/index.tsx
```diff
@@ -19,6 +19,27 @@ export function init() {
...
+      {
+        type: 'metadatachange',
+        metadata: {
+          // TODO: Once https://github.com/hypothesis/via/pull/1015 lands,
+          // get the URL from `link[rel=canonical]` or pass the video URL as
+          // config.
+          location: `https://www.youtube.com/watch?v=${encodeURIComponent(
+            videoId
+          )}`,
+          title: document.title,
```
lyzadanger, line 36: Confirm: expected `title` is `'Via: Hypothesis Video Player'`?

lyzadanger (approved):
The red loading bar is gone on this branch, and the title is "Via: Hypothesis Video Player".

👍🏻

via/static/scripts/video_player/index.tsx
robertknight: Currently, yes. "Hypothesis Video Player" comes from the video player and the "Via:" prefix is always added by Via to distinguish a web page visited directly from a web page proxied through Via.

PR #1023, opened 2023-06-29
Improve transcript scrolling and jumping to segments

robertknight:
This PR contains a couple of improvements related to transcript scrolling and jumping to segments, based on my experience testing out the current version on a few videos:

1. Change transcript scrolling from smooth to immediate. Immediate scrolling is IMO less distracting when watching the video, and is also faster for when the transcript scrolls larger distances.
2. Clear the filter after jumping to a segment. This makes it easier to read the text that comes before or after the segment that you've just jumped to.

See notes in commits for more details.


lyzadanger (approved):
:thumbsup:

PR #1024, opened 2023-06-29
Update transcript styling to better match mocks

lyzadanger:
Before:

Screen Shot 2023-06-29 at 12 20 10 PM

After:

Screen Shot 2023-06-29 at 12 19 53 PM

Reference: https://hypothesis-video-transcriptions.netlify.app/


robertknight (approved):
Looks good to me. One very minor thing, I notice the background at the top of the bottom, when an overscroll "bounce" happens, is still white. If we can match that to the segment background that would be nice.

PR #1026, opened 2023-06-29
Ensure transcript container and contents always use full vertical space

### LINKED ISSUE #1025: Ensure transcript is always fills available vertical space, even if there aren't sufficient transcript entries (bug/nit)
Also ensure the loading spinner is positioned correctly in these cases.

lyzadanger:
Ensure that the transcript container (column) always fills the full available vertical space, even if there are minimal or no transcript entries. The footer and "auto-scroll" control should always be flush against the bottom of the viewport in both wide- and narrow-screen layouts.

Fixes https://github.com/hypothesis/via/issues/1025

Before:

Screen Shot 2023-06-29 at 1 01 37 PM

Screen Shot 2023-06-29 at 1 01 28 PM

After:

Screen Shot 2023-06-29 at 1 00 51 PM

Screen Shot 2023-06-29 at 1 01 02 PM

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -312,7 +312,7 @@ export default function VideoPlayerApp({
       <main
         data-testid="app-layout"
         className={classnames('w-full flex min-h-0', {
-          'flex-col': !multicolumn,
+          'flex-col grow': !multicolumn,
           'flex-row grow h-full': multicolumn,
         })}
         ref={appContainerRef}
@@ -347,7 +347,7 @@ export default function VideoPlayerApp({
         </div>
 
         <div
-          data-testid="transcript-container"
+          data-testid="transcript-and-controls-container"
           className={classnames('flex flex-col bg-grey-0 border-x', {
             // Make transcript fill available vertical space in single-column
             // layouts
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -417,16 +417,12 @@ export default function VideoPlayerApp({
...
-              'relative',
-
-              // Override flex-basis for transcript. Otherwise this container
-              // will be sized to accomodate all transcript elements.
-              'min-h-0',
-
-              // Make the bucket bar fill this container.
-              'flex flex-col'
-            )}
+            className={classnames('relative grow', 'min-h-0', 'flex flex-col', {
```
robertknight, line 420: I think it would be useful to retain the comment about what `min-h-0` is for, for multicolumn layouts, since this is non-obvious for anyone who doesn't know flex really well.

robertknight (approved):

via/static/scripts/video_player/components/VideoPlayerApp.tsx
lyzadanger: A couple of challenges here:

1. I didn't really understand the existing comment: it seemed like it might not actually be true anymore as I parsed it.
2. We use `min-h-0` on a slew of container elements in this layout. Will we need to comment all of them?

Having said that, I will attempt to add _some_ sort of comment here.

lyzadanger: I have a proposal. I'm also spotting here that I don't believe that the `relative` class is necessary, but I don't want to remove it as part of this changeset.

I've created a [task](https://github.com/orgs/hypothesis/projects/118/views/1?pane=issue&itemId=32157417) (and assigned myself to it) to do a pass for CSS and ensure validity and commenting.

PR #1027, opened 2023-06-30
Restrict "More videos" to same channel

### LINKED ISSUE #1022: Hide related videos when video player is paused
When pausing the video player, YouTube may show an overlay of related videos (see "More videos" overlay in the screenshot below). This gets in the way of trying to read whatever is behind it in the video so I'd like to remove this, assuming the player has some API for this.

Related videos overlay

robertknight:
YouTube no longer allows turning off the "More videos" overlay [1] but it does allow recommendations to be restricted to the same channel as the current video. I think it makes sense to enable this restriction as:

 - In an education context, it keeps the content shown "closer" to what the teacher selected / from a source the teacher trusts.

 - Thumbnails from other videos in the same channel will tend to have the same aesthetic as the current video, so it makes the overlay a bit less visually jarring.

[1] https://developers.google.com/youtube/player_parameters#august-23,-2018

Closes #1022

----

**Testing:**

1. On main, start a video and pause it a few seconds later. You will see suggestions from a variety of channels.
2. On this branch, start a video and pause it. You will see suggestions only from the current channel.

### DIFF
--- a/via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
+++ b/via/static/scripts/video_player/components/YouTubeVideoPlayer.tsx
@@ -1,5 +1,5 @@
 import { AspectRatio } from '@hypothesis/frontend-shared';
-import { useEffect, useRef, useState } from 'preact/hooks';
+import { useEffect, useMemo, useRef, useState } from 'preact/hooks';
 
 import { loadYouTubeIFrameAPI } from '../utils/youtube';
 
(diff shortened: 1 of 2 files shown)


lyzadanger (approved):
I see where you're coming from and am happy to move forward with this, but wanted to raise a counter-perspective. To wit, the suggestions for a video I'm using about the GT Corolla (admittedly a personal obsession) previously had these suggestions:

Screen Shot 2023-06-30 at 8 01 53 AM

and now has

Screen Shot 2023-06-30 at 8 02 15 AM

While the thumbnails are perhaps more consistent, the suggested videos are less useful to me as someone interested in a GR Corolla :). I don't really care about this one guy's opinion about other cars (what I see now), but I am interested in other stuff about this topic (what I saw before).

I guess the underlying question is: do we want to show related videos from a single "authority" (these changes) or about a shared "topic" (default behavior from YT)? Your tack is more the former, while my opinion is still slightly towards the latter. Since I am very much n of 1, happy to support your plan here. I am receptive to the notion that we're trying to solve the educational use case and indeed perhaps the "shared authority" walks that line better.

robertknight:
Hmm... I'll ask a wider audience internally to see what folks think.

robertknight:
Nairi provided feedback [in Slack](https://hypothes-is.slack.com/archives/C054CG5PH5M/p1688554114367599?thread_ts=1688127665.477779&cid=C054CG5PH5M) that sticking to same-channel videos seems preferable on the basis that it represents a "trusted" (or at least "known") source of content. On that basis I'm going to merge this. We can easily revisit in future.

Issue #1028, opened 2023-06-30
Display error details if transcript fails to load


robertknight:
If an error occurs loading a transcript, the video player currently displays a generic and unstyled message:

<img width="425" alt="Transcript error" src="https://github.com/hypothesis/via/assets/2458/1983df95-e14f-4505-aceb-d2b548110053">

Instead it should display something more informative using the error details returned from the backend. In the case of https://www.youtube.com/watch?v=dNrTrx42DGQ the details returned are:

(_Update: This video now *does* have a transcript. It is a long video that had only just been uploaded when I initially reported this. Try https://www.youtube.com/watch?v=e-43x-H4CNY instead, which is a video that inherently has no spoken content._)

```json
{
    "errors": [
        {
            "status": 500,
            "code": "TranscriptsDisabled",
            "title": "Subtitles are disabled for this video",
            "detail": "Could not retrieve a transcript for the video https://www.youtube.com/watch?v=dNrTrx42DGQ! This is most likely caused by:\n\nSubtitles are disabled for this video\n\nIf you are sure that the described cause is not responsible for this error and that a transcript should be retrievable, please create an issue at https://github.com/jdepoix/youtube-transcript-api/issues. Please add which version of youtube_transcript_api you are using and provide the information needed to replicate the error. Also make sure that there are no open issues which already describe your problem!"
        }
    ]
}
```

We should avoid including `detail` verbatim here because it references the Python library we currently use.

The value of the `code` field above comes from the names of error classes in https://github.com/jdepoix/youtube-transcript-api/blob/master/youtube_transcript_api/_errors.py.

robertknight:
I had a query about the HTTP status here. That is being discussed in https://hypothes-is.slack.com/archives/C1MA4E9B9/p1688120133798219.

lyzadanger:
Given that this type of error is non-recoverable — the app is not usable when these happen — I'd suggest we not use the toast pattern and instead have a persistent error. We can still style it consistently using the `Callout` component from the `frontend-shared` package[^1], which drops in easily:

```jsx
<div data-testid="transcript-error">
  <Callout status="error">
    Unable to load transcript: {transcript.message}
  </Callout>
</div>
```

That looks like this:

<img width="519" alt="image" src="https://github.com/hypothesis/via/assets/439947/0f439214-1b49-4d94-8403-439297c1c460">


[...]

PR #1030, opened 2023-06-30
Show error summary from API when transcript fails to load

robertknight:
The API returns JSON:API error objects with `code`, `title` and `details` fields. Show the `title` to the user instead of a generic "API call failed" message.

Part of https://github.com/hypothesis/via/issues/1028

**Testing:**

1. Go to http://localhost:9083/https://www.youtube.com/watch?v=e-43x-H4CNY
2. You should see the error message below:

Transcript error display

Note that the error title and contents of the "Error details" box need a bunch of work on the backend to refine. For example subtitles are not so much "disabled" for this video as "not available" because the video doesn't have spoken words. This morning I also encountered the same error message for a video where the transcript just hadn't been generated. In the frontend we are just displaying what the backend sends.

### DIFF
--- a/via/static/scripts/video_player/components/TranscriptError.tsx
+++ b/via/static/scripts/video_player/components/TranscriptError.tsx
@@ -9,12 +9,6 @@ export default function TranscriptError({ error }: TranscriptErrorProps) {
     <div className="p-3">
       <h2 className="text-lg">Unable to load transcript</h2>
       <p className="mb-3">{error.error?.title ?? error.message}</p>
-      {error.error?.detail && (
-        <details>
-          <summary className="mb-2">Error details:</summary>
-          <p>{error.error.detail}</p>
-        </details>
-      )}
     </div>
   );
 }
--- a/via/static/scripts/video_player/components/TranscriptError.tsx
+++ b/via/static/scripts/video_player/components/TranscriptError.tsx
@@ -1,0 +1,20 @@
+import type { APIError } from '../utils/api';
+
+export type TranscriptErrorProps = {
+  error: APIError;
+};
+
+export default function TranscriptError({ error }: TranscriptErrorProps) {
+  return (
+    <div className="p-3">
+      <h2 className="text-lg">Unable to load transcript</h2>
+      <p className="mb-3">{error.error?.title ?? error.message}</p>
+      {error.error?.detail && (
+        <details>
+          <summary className="mb-2">Error details:</summary>
+          <p>{error.error.detail}</p>
+        </details>
+      )}
+    </div>
+  );
+}
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -26,6 +26,7 @@ import { formatTranscript, mergeSegments } from '../utils/transcript';
 import HypothesisClient from './HypothesisClient';
 import Transcript from './Transcript';
 import type { TranscriptControls } from './Transcript';
+import TranscriptError from './TranscriptError';
 import YouTubeVideoPlayer from './YouTubeVideoPlayer';
 import { PauseIcon, PlayIcon, SyncIcon } from './icons';
 
@@ -459,9 +460,7 @@ export default function VideoPlayerApp({
               </Transcript>
             )}
             {transcript instanceof Error && (
-              <div data-testid="transcript-error">
-                Unable to load transcript: {transcript.message}
-              </div>
+              <TranscriptError error={transcript} />
             )}
             <div
               id={bucketContainerId}
--- a/via/static/scripts/video_player/components/test/TranscriptError-test.js
+++ b/via/static/scripts/video_player/components/test/TranscriptError-test.js
@@ -1,0 +1,33 @@
+import { mount } from 'enzyme';
+
+import { APIError } from '../../utils/api';
+import TranscriptError from '../TranscriptError';
+
+describe('TranscriptError', () => {
+  it('displays just `Error.message` if there are no error details', () => {
+    const wrapper = mount(
+      <TranscriptError error={new Error('Something went wrong')} />
+    );
+    assert.equal(
+      wrapper.text(),
+      ['Unable to load transcript', 'Something went wrong'].join('')
+    );
+  });
+
+  it('displays `APIError.error.title` field if present', () => {
+    const error = new APIError(404, {
+      code: 'VideoNotFound',
+      title: 'The video was not found',
+      detail: 'Some long details here',
+    });
+    const wrapper = mount(<TranscriptError error={error} />);
+    assert.equal(
+      wrapper.text(),
+      [
+        'Unable to load transcript',
+        'The video was not found',
+        'Error details:Some long details here',
+      ].join('')
+    );
+  });
+});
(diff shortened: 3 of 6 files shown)


via/static/scripts/video_player/components/test/TranscriptError-test.js
```diff
@@ -0,0 +1,33 @@
...
+import TranscriptError from '../TranscriptError';
+
+describe('TranscriptError', () => {
+  it('displays just `Error.message` if there are no error details', () => {
+    const wrapper = mount(
+      <TranscriptError error={new Error('Something went wrong')} />
+    );
+    assert.equal(
+      wrapper.text(),
+      ['Unable to load transcript', 'Something went wrong'].join('')
```
robertknight, line 13: `[...].join('')` is used here to separate out parts that are on different lines in the UI, but that don't have actual spaces between them.

via/static/scripts/video_player/utils/api.ts
```diff
@@ -57,7 +68,15 @@ export async function callAPI<T = unknown>(api: APIMethod): Promise<T> {
...
 
   if (result.status >= 400 && result.status < 600) {
-    throw new APIError(result.status, resultData);
+    let error;
+    if (
+      resultData &&
+      Array.isArray(resultData.errors) &&
+      resultData.errors.length > 0
+    ) {
+      error = resultData.errors[0];
```
robertknight, line 77: The JSON:API format allows for multiple errors, but we currently only report the first one on the frontend. I'm not aware of a scenario where we'd expect the backend to report multiple yet.

via/static/scripts/video_player/components/TranscriptError.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <div className="p-3">
+      <h2 className="text-lg">Unable to load transcript</h2>
+      <p className="mb-3">{error.error?.title ?? error.message}</p>
+      {error.error?.detail && (
+        <details>
+          <summary className="mb-2">Error details:</summary>
+          <p>{error.error.detail}</p>
+        </details>
+      )}
+    </div>
```
robertknight, line 12: Styling is obvious not final here.

lyzadanger:
@robertknight Can you help me understand the goal of the "details" box once it's more functional?

robertknight:
> @robertknight Can you help me understand the goal of the "details" box once it's more functional?

I was thinking it would work like the "Details" box in LMS error messages. It could include any technical information that we think might be useful for debugging or other purposes. In the LMS context, the support/success team for example will often open this if they are on a call with a user and inspect / make note of what is in there.

lyzadanger:
> I was thinking it would work like the "Details" box in LMS error messages. It could include any technical information that we think might be useful for debugging or other purposes. In the LMS context, the support/success team for example will often open this if they are on a call with a user and inspect / make note of what is in there.

Are you fairly convinced that it adds value? I know you want it to be tweaked, but right now it reads:

> Could not retrieve a transcript for the video https://www.youtube.com/watch?v=e-43x-H4CNY! This is most likely caused by: Subtitles are disabled for this video If you are sure that the described cause is not responsible for this error and that a transcript should be retrievable, please create an issue at https://github.com/jdepoix/youtube-transcript-api/issues. Please add which version of youtube_transcript_api you are using and provide the information needed to replicate the error. Also make sure that there are no open issues which already describe your problem!

AFAICT that doesn't add any information or use beyond what the concise error message says. Can you give me an example of what it might say in future that would be useful?

_Edited_: A plausible path forward could be to omit the `<details>` element until it has something useful to say. Looks like it's easy to drop in?

robertknight:
> AFAICT that doesn't add any information or use beyond what the concise error message says. Can you give me an example of what it might say in future that would be useful?

There is an open issue at https://github.com/hypothesis/via/issues/1044. What I am expecting as a result of resolving that issue is that the backend would omit the `detail` field for this particular error, and so the `<details>` field would not be shown.

We don't currently have a case where we have extra information to show. I would expect the backend to simply omit the `detail` field whenever it has nothing to add.

lyzadanger:
> There is an open issue at #1044. What I am expecting as a result of resolving that issue is that the backend would omit the `detail` field for this particular error, and so the `<details>` field would not be shown.

So, maybe we could punt on the `<details>` element for now until we have a clear need for it?

robertknight:
> So, maybe we could punt on the <details> element for now until we have a clear need for it?

OK, I'll do that.

robertknight:
Done! - The `<details>` is now gone.

via/static/scripts/video_player/components/TranscriptError.tsx
lyzadanger: Yep, don't worry about that for now. I've got it.

via/static/scripts/video_player/components/test/TranscriptError-test.js
```diff
@@ -0,0 +1,29 @@
+import { mount } from 'enzyme';
+
+import { APIError } from '../../utils/api';
+import TranscriptError from '../TranscriptError';
+
+describe('TranscriptError', () => {
+  it('displays just `Error.message` if there are no error details', () => {
```
lyzadanger, line 7: Does this test description need adjustment in light of the `details` removal?

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -459,9 +460,7 @@ export default function VideoPlayerApp({
               </Transcript>
             )}
             {transcript instanceof Error && (
-              <div data-testid="transcript-error">
-                Unable to load transcript: {transcript.message}
-              </div>
+              <TranscriptError error={transcript} />
```
lyzadanger, line 463: I think it's neat that the same reference can accommodate either an actual transcript or an error (and am of the opinion that typing makes this possible versus a bug waiting to happen). But I admit I did a double-take at `error={transcript}` here.

via/static/scripts/video_player/utils/test/api-test.js
```diff
@@ -33,38 +33,71 @@ describe('callAPI', () => {
...
-    assert.instanceOf(error, APIError);
-    assert.equal(error.status, 500);
-    assert.equal(error.data, null);
-    assert.equal(error.message, 'API call failed');
-  });
+  context('when request fails', () => {
+    it('throws exception if result is not JSON', async () => {
+      fakeFetch
+        .withArgs(videoPlayerConfig.api.transcript.url)
+        .resolves(new Response('<b>Oh no</b>', { status: 500 }));
```
lyzadanger, line 40: Just confirming that we intend to invent a 500 here in these cases :). Granted this is not new behavior in these changes, so feel free to ignore me.

lyzadanger (approved):
Thanks for your patience! I'm working on collating some error-styling needs into a single ticket to track, so, I got that part.

via/static/scripts/video_player/components/test/TranscriptError-test.js
robertknight: This is still valid as "error details" includes the title. This case is for when we don't have a server-generated error at all, or it doesn't include even a title.

via/static/scripts/video_player/utils/test/api-test.js
robertknight: Yes, this is mean to simulate something like eg. a 5xx "Bad Gateway" error if the server failed to respond within the various timeouts that the proxies in front have.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
robertknight: Yeah, I might rename this to `transcriptOrError` in future.

PR #1045, opened 2023-07-03
Run Via's unit and functional tests in parallel

seanh:
This might speed up `make test`, `make functests`, `make sure` and CI,
but it depends on how many tests there are: if there are only a small
number of tests (and the tests are fast) then running them in parallel
can actually be slower.

The amount of parallelism you get from `make sure` is more limited
because it's already running several things in parallel (tests,
functests, linting, etc) so there's fewer cores available to further
parallelise the tests and functests within that. This may make it more
likely that parallelising the unit and functional tests within `make
sure` will actually make it slower: you're paying the overhead of
paralleisation but don't actually have much parallism available.

Whenever there are more or slower unit and functional tests then I think
running them in parallel should start to make things much faster.

### DIFF
--- a/.cookiecutter/includes/requirements/functests.in
+++ b/.cookiecutter/includes/requirements/functests.in
@@ -1,0 +1,1 @@
+pytest-xdist[psutil]
--- a/.cookiecutter/includes/requirements/tests.in
+++ b/.cookiecutter/includes/requirements/tests.in
@@ -1,0 +1,1 @@
+pytest-xdist[psutil]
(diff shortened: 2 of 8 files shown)


seanh:
I think coverage is probably broken on CI because the move to `pytest-cov` means I need to update the part of `ci.yml` that passes the coverage files between jobs.

seanh:
Gonna close this for now but I think it may be worth adding this to the cookiecutter as an option

PR #1049, opened 2023-07-05
Style non-dismissable errors

lyzadanger:
This small PR adds UI and styling for non-dismiss-able transcript errors. It bumps to v4.6.0 of `frontend-shared` to get its hands on the new `Callout` component.

Note that I've used a caution icon instead of a cancel (x) icon as @robertknight pointed out that the `x` made it look like it was interactive or dismiss-able. I've also added a `role="alert"` so that this error is picked up and read by screen readers.

We can continue to iterate on the layout of the content _inside_ the box as we go, if we so choose.

image


robertknight (approved):
LGTM

PR #1054, opened 2023-07-10
Support m.youtube.com URLs

### LINKED ISSUE #1053: YouTube mobile links are still blocked by Checkmate
We recently made it so that YouTube links can be "visited" through Via (in the QA environment at least). These "redirect" to our video player. However the mobile versions of these links that start with `m.youtube.com` instead of `www.youtube.com` are still blocked:

http://qa-via.hypothes.is/https://www.youtube.com/watch?v=CjrmuHHEAes works
http://qa-via.hypothes.is/https://m.youtube.com/watch?v=CjrmuHHEAes is blocked

The mobile links are what you get if you visit YouTube's web app on a phone or tablet and copy the URL.

The expected result is that all the YouTube URL variants that we understand can be visited the same way. I haven't checked all the other YouTube URL variants yet.

seanh:
Support m.youtube.com URLs in the new video player.

Fixes https://github.com/hypothesis/via/issues/1053


robertknight (approved):
Works for me 👍

PR #1055, opened 2023-07-10
Fix a false positive when identifying YouTube URLs

seanh:


robertknight (approved):

robertknight:
I'm going to go ahead and merge this tiny fix before I deploy.

PR #1056, opened 2023-07-10
Position current segment towards top of container when auto-scrolling

### LINKED ISSUE #1052: Transcript viewport to 80% read ahead
Pin the current item in the transcript closer to the top of the transcript pane so the user can see more of the transcript coming up (about 80%).

robertknight:
Instead of positioning the current segment in the middle of the transcript container when auto-scrolling, position it closer to the top instead. The rationale is that the user is likely to want read ahead further than they want to read back.

Fixes https://github.com/hypothesis/via/issues/1052

Example of the new vertical positioning:

adjust-autoscroll-pos

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -263,17 +263,17 @@ export default function Transcript({
       return;
     }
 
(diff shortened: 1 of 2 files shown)


acelaya (approved):
👍🏼

PR #1058, opened 2023-07-10
Make segments current when clicked, if not (de-)selecting text

### LINKED ISSUE #1051: Click transcript "card" to seek in video
Users should be able to click a transcript "card" (chunk of text) to seek the video. Currently only able to do via the timestamp

robertknight:
Enable users to seek to a segment by clicking the segment's text, as an alternative to clicking the timestamp, but only if the click does not create or remove a text selection. The aim is to make seeking the video more convenient / intuitive, but without causing the video position to jump if the user is trying to select text to annotate, or de-select text to remove the Annotate/Highlight popup.

Fixes https://github.com/hypothesis/via/issues/1051

**Testing:**

1. Open a video in the video player
2. Click on a segment's text, it should be made current and the video should seek to that position
3. Start selecting the text of a different segment. It should not be made current when the mouse is released.
4. Clear the selection created in step 3 by clicking on a different non-current segment. It should not be made current when the mouse is released

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -87,6 +87,10 @@ type TranscriptSegmentProps = {
   text: string;
 };
 
+function hasSelection() {
+  return Boolean(window.getSelection()?.toString());
+}
+
 function TranscriptSegment({
   hidden = false,
   highlight,
@@ -124,6 +128,8 @@ function TranscriptSegment({
       ranges.forEach(r => highlight.delete(r));
     };
   }, [highlight, matches]);
+
+  const hadSelectionOnPointerDown = useRef(false);
 
   const timestamp = formatTimestamp(time);
 
@@ -194,6 +200,20 @@ function TranscriptSegment({
         )}
         data-testid="transcript-text"
         ref={contentRef}
+        // We have a "click" handler here, but don't set a role because
+        // this is a secondary way to focus the transcript. The timestamp
+        // button is the primary control for this action.
+        onPointerDown={() => {
+          hadSelectionOnPointerDown.current = hasSelection();
+        }}
+        onPointerUp={() => {
+          // Allow the user to easily select a segment by clicking it, but
+          // don't seek the video if they are selecting or de-selecting text
+          // to annotate.
+          if (!hadSelectionOnPointerDown.current && !hasSelection()) {
+            onSelect();
+          }
+        }}
       >
         {text}
         {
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -194,6 +200,20 @@ function TranscriptSegment({
         )}
         data-testid="transcript-text"
         ref={contentRef}
+        // We have a "click" handler here, but don't set a role because
+        // this is a secondary way to focus the transcript. The timestamp
+        // button is the primary control for this action.
```
robertknight, line 205: The result I'm going for here is that the segment timestamp is the accessible way of seeking the video - a button with completely ordinary behavior. Clicking the text is an alternative method, but not marked as a button because it has some non-standard behavior associated with it.

acelaya: Makes sense.

For non sighted users, the other one should be enough, and I would even say, having two controls doing the same, might be confusing for them.

acelaya (approved):
LGTM, and it works as expected on desktop.

I have noticed a slightly awkward behavior on Android, but probably not directly caused by this, but something we may need to check separately.

Basically, if the OS detects you tapped on a word, the default behavior is to select it, which prevents selecting a transcript segment when you tap on it, even if it didn't have anything selected before.

After that, if you tap on other words, it moves selection to them. Only when it detects you tapped on a "blank spot", then it either deselects previous selection (if any), or makes that segment active.

https://github.com/hypothesis/via/assets/2719332/3e72ec39-63dc-4b82-9f35-01d4b38a34d8

acelaya:
> Only when it detects you tapped on a "blank spot", then it either deselects previous selection (if any), or makes that segment active.

Actually, this is not entirely correct. There seems to be a race condition, and sometimes it changes the active segment first, and then selects the tapped word. Other times, the word is selected before, which prevents changing the active segment.

robertknight:
Thanks for the testing. I'm going to land the current version of this as that will make it easier to test across a range of devices. It seems like we will need to do some refinement on mobile.

> Basically, if the OS detects you tapped on a word, the default behavior is to select it, which prevents selecting a transcript segment when you tap on it, even if it didn't have anything selected before.

Is this even for a quick tap, not a longer press?

acelaya:
> Is this even for a quick tap, not a longer press?

Yes, that's how it works now.

I think it didn't use to work like that, but at some point it was changed. I find it a bit annoying.

PR #1060, opened 2023-07-10
Expose dev server's port 9083 for any host, not just 127.0.0.1

acelaya:
This PR changes the behavior of nginx exposed by dev env's `docker compose` config, so that it exposes port `9083` to any IP address, not just 127.0.0.1.

This will allow things like testing the video app from a mobile device that's on the same network as the laptop/computer running via.


robertknight (approved):
Works for me. I tested accessing Via, running on my laptop, on my iPhone.

Issue #954, opened 2023-06-05
Show sidebar alongside transcript


robertknight:
In the video player app, make it possible to see the transcript alongside the sidebar. This will make it possible to read the transcript and annotation threads at the same time, like side-by-side mode for PDFs or web pages.

In the current implementation, the sidebar displays on top of the transcript:

<img width="1407" alt="Video player snapshot" src="https://github.com/hypothesis/via/assets/2458/004fa29e-a166-4ce9-a471-cc93dafbf3b8">

The current tentative design idea is to make it so that when the sidebar is open, the transcript is displayed underneath the video instead.

robertknight:
https://github.com/hypothesis/via/issues/952 is connected to this. If we move the transcript container under the video, it is going to be even more valuable than it is now to display the bucket bar alongside the transcript rather than as part of the sidebar.

robertknight:
This is now implemented, based on Hypothesis's side-by-side mode for HTML, but with some hacks in Via to gain more control over positioning. See https://github.com/hypothesis/client/issues/5571.

lyzadanger:
Update: this ticket will be done when we implement a mechanism for the video player to be able to style its own "side-by-side mode" independent of the client. i.e. it should be possible for the host page to disabled side-by-side mode, and respond to the current layout state of the sidebar in whatever way it sees fit. The layout should also be less awkward across all breakpoints.

lyzadanger:
Complete as of https://github.com/hypothesis/via/pull/1083

nhan000:
How can I have this interface here?
![](https://user-images.githubusercontent.com/2458/243326817-004fa29e-a166-4ce9-a471-cc93dafbf3b8.png)

Is there anything from the `project: video annotations` people can try out yet? If not, where I can follow the development to know when it's ready? Will there be an ability to annotate offline video? I'm somewhat familiar with https://docdrop.org/ but this is a separate thing right? Thanks a lot!

PR #1063, opened 2023-07-10
Disable client sidebyside; apply layout with useSideBySideLayout hook

lyzadanger:
This PR adds a new hook, `useSideBySideLayout`, disables the client's default side-by-side layout handling and implements its own. 

~~In draft until:~~

* [x] Add tests for `useSideBySideLayout` hook

These changes make it so the layout vis-a-vis the relationship between the sidebar and the app content should be relatively sensible across different viewport widths and sidebar states.

The new hook listens for the `hypothesis:layoutchange` event and updates internal state about sidebar layout. When that state changes, or when there is a window-resize, it will compute whether or not to apply a side-by-side layout (make room to the right of the app content). It tunes the exact dimensions allotted to the sidebar such that the video-app UI tucks up nicely against it in various configurations.

## Testing

On this branch, try sizing your browser to different sizes and opening and collapsing the sidebar. There should no longer be awkward ranges in which the layout is two-column, but side-by-side doesn't apply. Translated, the sidebar will only ever lay on top of content (versus beside it) when the content is in a single-column layout, and will no longer ever obscure the transcript column in a two-column layout.

## Not yet

These changes do not address these things, which will be addressed in follow-up PRs:

* Tuning of breakpoints, and re-distributing them (the `md` size only spans about 11px of range as of these changes, but I didn't want to clutter this diff)
* Addressing layout nits beyond the overall layout with the sidebar (e.g. video size/proportion)
* Refactor of what the `useAppLayout` hook returns to encapsulate more and not have `VideoPlayerApp` doing layout computation

Part of #954

### DIFF
--- a/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
+++ b/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
@@ -1,0 +1,135 @@
+import { mount } from 'enzyme';
+import { act } from 'preact/test-utils';
+
+import { useSideBySideLayout } from '../use-side-by-side-layout';
+import { TOOLBAR_WIDTH } from '../use-side-by-side-layout';
+
+describe('useSideBySideLayout', () => {
+  let container;
+  let wrappers;
+  let fakeLayoutDetail;
+
+  // Create a fake component to mount in tests that uses the hook
+  function FakeComponent() {
+    useSideBySideLayout(container);
+    return (
+      <div style="width:100%">
+        <button>Hi</button>
+      </div>
+    );
+  }
+
+  function renderComponent() {
+    const wrapper = mount(<FakeComponent />, { attachTo: container });
+    wrappers.push(wrapper);
+    return wrapper;
+  }
+
+  beforeEach(() => {
+    container = document.createElement('div');
+    container.setAttribute('id', 'container');
+    container.style.width = '1024px';
+    container.style.paddingRight = '0px';
+    document.body.append(container);
+
+    wrappers = [];
+
+    // Layout values here reflect dimensions of the sidebar reported in the
+    // `hypothesis:layoutchange` event at time of authoring.
+    fakeLayoutDetail = {
+      sideBySideActive: false,
+      sidebarLayout: {
+        expanded: true,
+        width: 461,
+        toolbarWidth: 33,
+        height: 0,
+      },
+    };
+  });
+
+  afterEach(() => {
+    wrappers.forEach(w => w.unmount());
+    container.remove();
+  });
+
+  function triggerEvent(detail = fakeLayoutDetail) {
+    const event = new CustomEvent('hypothesis:layoutchange', {
+      detail,
+    });
+    act(() => {
+      document.body.dispatchEvent(event);
+    });
+  }
+
+  function sideBySideState() {
+    const attr = container.getAttribute('data-side-by-side');
+    if (attr === 'true') {
+      return true;
+    } else if (attr === 'false') {
+      return false;
+    } else {
+      return undefined;
+    }
+  }
+
+  describe('initial render', () => {
+    it('does not lay out side-by-side', () => {
+      renderComponent();
+      assert.isFalse(sideBySideState());
+    });
+
+    it('provides room for sidebar controls', () => {
+      renderComponent();
+      assert.equal(container.style.paddingRight, `${TOOLBAR_WIDTH}px`);
+    });
+  });
+
+  describe('when sidebar opens', () => {
+    [875, 876, 1024, 1268].forEach(bodyWidth => {
+      it('applies side-by-side layout when there is enough space', () => {
+        container.style.width = `${bodyWidth}px`;
+        renderComponent();
+        triggerEvent();
+
+        assert.isTrue(sideBySideState());
+      });
+    });
+
+    [450, 600, 874].forEach(bodyWidth => {
+      it('allows sidebar to overlap content on narrow screens', () => {
+        container.style.width = `${bodyWidth}px`;
+        renderComponent();
+        triggerEvent();
+
+        assert.isFalse(sideBySideState());
+      });
+    });
+  });
+
+  describe('when window is resized', () => {
+    it('recalculates side-by-side layout', () => {
+      container.style.width = '450px';
+      renderComponent();
+      triggerEvent();
+
+      assert.isFalse(sideBySideState());
+
+      act(() => {
+        container.style.width = '1450px';
+        window.dispatchEvent(new Event('resize'));
+      });
+      assert.isTrue(sideBySideState());
+    });
+  });
+
+  describe('when sidebar closes', () => {
+    it('does not provide room for sidebar at right', () => {
+      renderComponent();
+      fakeLayoutDetail.sidebarLayout.expanded = false;
+      triggerEvent();
+
+      assert.isFalse(sideBySideState());
+      assert.equal(container.style.paddingRight, `${TOOLBAR_WIDTH}px`);
+    });
+  });
+});
--- a/via/static/scripts/video_player/hooks/use-app-layout.ts
+++ b/via/static/scripts/video_player/hooks/use-app-layout.ts
@@ -1,6 +1,8 @@
 import type { RefObject } from 'preact';
 import { useCallback, useLayoutEffect, useState } from 'preact/hooks';
 
+import { SIDEBYSIDE_THRESHOLD, TOOLBAR_WIDTH } from './use-side-by-side-layout';
+
 /* Relative application container size */
 export type AppSize = 'sm' | 'md' | 'lg' | 'xl' | '2xl';
 
@@ -30,12 +32,11 @@ export function useAppLayout(
       containerSize = '2xl';
     } else if (containerWidth > 1024) {
       containerSize = 'xl';
-    } else if (containerWidth > 875) {
+    } else if (containerWidth > SIDEBYSIDE_THRESHOLD) {
       containerSize = 'lg';
-    } else if (containerWidth > 768) {
+    } else if (containerWidth >= SIDEBYSIDE_THRESHOLD - TOOLBAR_WIDTH) {
       containerSize = 'md';
     }
-
     setAppSize(containerSize);
   }, [appContainer]);
 
--- a/via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
+++ b/via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
@@ -1,0 +1,93 @@
+import { useCallback, useEffect, useState } from 'preact/hooks';
+
+// `hypothesis:layoutchange` is triggered on the host document `body` element by
+// the client's `annotator`. Types are copied from the `client` project. See
+// https://github.com/hypothesis/client/blob/16e37788fdc52fd7e7b8eb9c0eece6a993334424/src/annotator/events.ts
+type SidebarLayout = {
+  /** Whether sidebar is open or closed */
+  expanded: boolean;
+  /** Current width of sidebar in pixels */
+  width: number;
+  /** Current height of sidebar in pixels */
+  height: number;
+  /** Width of controls (toolbar, bucket bar) on the edge of the sidebar */
+  toolbarWidth: number;
+};
+
+type LayoutChangeEventDetail = {
+  sideBySideActive: boolean;
+  sidebarLayout: SidebarLayout;
+};
+
+type LayoutChangeEvent = CustomEvent & { detail: LayoutChangeEventDetail };
+
+// Minimum body width at which laying out side-by-side should be considered
+export const SIDEBYSIDE_THRESHOLD = 875;
+// The sidebar reports its toolbar width as 33px. This is a custom padding
+// amount that allows this app to nestle up more cloesly to the sidebar when
+// closed.
+export const TOOLBAR_WIDTH = 22;
+
+/**
+ * Adjust the available space to the right of app content to make room for
+ * the sidebar when plausible. This mimics the client's "side-by-side" mode,
+ * tuned specifically for this application.
+ *
+ * Side-by-side layout will be re-computed whenever sidebar state via the
+ * `hypothesis:layoutchange` event changes meaningfully, or when the window
+ * is resized.
+ *
+ * @param [container] test seam
+ */
+export function useSideBySideLayout(container = document.body) {
+  const [sidebarOpen, setSidebarOpen] = useState(false);
+  const [sufficientSpace, setSufficientSpace] = useState(
+    container.clientWidth >= SIDEBYSIDE_THRESHOLD
+  );
+  const [sidebarWidth, setSidebarWidth] = useState(0);
+
+  const updateSidebarState = useCallback((e: Event) => {
+    const sidebarInfo: LayoutChangeEventDetail = (e as LayoutChangeEvent)
+      .detail;
+    const layout = sidebarInfo.sidebarLayout;
+
+    setSidebarOpen(layout.expanded);
+    setSidebarWidth(layout.width);
+  }, []);
+
+  // Make room for the sidebar at the right of the app layout when applicable.
+  // Otherwise, provide some room for the sidebar's controls. Padding is used
+  // here (versus margin) because it is part of an element's dimensions
+  // (`clientWidth`) and thus easier to compute with.
+  useEffect(() => {
+    const applySideBySide = sufficientSpace && sidebarOpen;
+    if (applySideBySide) {
+      // prettier-ignore
+      container.style.paddingRight = `${
+        sidebarWidth - (TOOLBAR_WIDTH / 2)
+      }px`;
+    } else {
+      container.style.paddingRight = `${TOOLBAR_WIDTH}px`;
+    }
+    container.setAttribute('data-side-by-side', `${applySideBySide}`);
+  }, [container, sidebarOpen, sidebarWidth, sufficientSpace]);
+
+  useEffect(() => {
+    const onResize = () => {
+      setSufficientSpace(container.clientWidth >= SIDEBYSIDE_THRESHOLD);
+    };
+
+    document.body.addEventListener(
+      'hypothesis:layoutchange',
+      updateSidebarState
+    );
+    window.addEventListener('resize', onResize);
+    return () => {
+      document.body.removeEventListener(
+        'hypothesis:layoutchange',
+        updateSidebarState
+      );
+      window.removeEventListener('resize', onResize);
+    };
+  }, [container, updateSidebarState]);
+}
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -18,6 +18,7 @@ import {
 } from 'preact/hooks';
 
 import { useAppLayout } from '../hooks/use-app-layout';
+import { useSideBySideLayout } from '../hooks/use-side-by-side-layout';
 import { callAPI } from '../utils/api';
 import type { APIMethod, APIError, JSONAPIObject } from '../utils/api';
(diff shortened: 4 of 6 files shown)


via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -30,12 +32,11 @@ export function useAppLayout(
       containerSize = '2xl';
     } else if (containerWidth > 1024) {
       containerSize = 'xl';
-    } else if (containerWidth > 875) {
+    } else if (containerWidth > SIDEBYSIDE_THRESHOLD) {
       containerSize = 'lg';
-    } else if (containerWidth > 768) {
+    } else if (containerWidth >= SIDEBYSIDE_THRESHOLD - TOOLBAR_WIDTH) {
```
lyzadanger, line 37: These need to be tuned, but this update ensures they're "in sync" with the side-by-side mode, at least.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Side-by-side layout will be re-computed whenever sidebar state via the
+ * `hypothesis:layoutchange` event changes meaningfully, or when the window
+ * is resized.
+ */
+export function useSideBySideLayout() {
+  const [sideBySideApplicable, setSideBySideApplicable] = useState(false);
+  const [sufficientSpace, setSufficientSpace] = useState(
+    document.body.clientWidth >= SIDEBYSIDE_THRESHOLD
+  );
+  const [sidebarWidth, setSidebarWidth] = useState(0);
```
lyzadanger: The layout event fires frequently, as could resize, but my aim with these is to have state that don't change value too often. This was the minimum set I could come up with (given that a resize handler would need some retained sidebar state to reference), but I'm open to suggestions.

acelaya: Maybe this could benefit from `useReducer`? It's a hook I think I have used zero times in production code, but dropping the suggestion here in case it makes sense 🤷🏼‍♂️

robertknight:
In testing this I realize there is an issue that the client still needs to know whether side-by-side is active, even when the app is handling the implementation itself. This is because it makes use of this state when deciding whether clicking on the document should close the sidebar or not.

Slack discussion: https://hypothes-is.slack.com/archives/C1M8NH76X/p1689065493060469

robertknight:
There is a [draft client PR](https://github.com/hypothesis/client/pull/5623) that adds a configuration key that the video player can use to communicate the side-by-side state to the client.

Usage in the video player app would look like so:

```js
const clientConfig = useMemo(() => {
  return {
    ...baseClientConfig,
    bucketContainerSelector: '#' + bucketContainerId,
    sideBySide: {
      mode: 'manual',
      isActive: () => {
        // TODO - Return true or false as appropriate.
        return true;
      },
    },
  };
}, [baseClientConfig]);
 ```

**Comment by lyzadanger** [comment]
Per verbal discussion today, we're going to tackle the client callback (so the client can query if side-by-side is being applied by the host) subsequently to these changes. Next steps are to clean up this work and add tests.

File: via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js [hunk 3]
```diff
@@ -0,0 +1,149 @@
...
+    wrappers.push(wrapper);
+    return wrapper;
+  }
+
+  beforeEach(() => {
+    document.body.style.width = '1024px';
+    document.body.style.paddingRight = '0px';
+    container = document.createElement('div');
+    container.setAttribute('id', 'container');
+    document.body.append(container);
```
robertknight, line 33: The `body` style adjustments could potentially interfere with other tests. If we can make these tests work with a temporary element (eg. by adding a test seam to the hook), that would be preferable, though it will still need to be mounted into the body for the duration of the test. The other option would be to just make sure the style adjustments are undone after the tests have executed.

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
```diff
@@ -0,0 +1,149 @@
...
+
+    wrappers = [];
+
+    fakeLayoutDetail = {
+      sideBySideActive: false,
+      sidebarLayout: {
+        expanded: true,
+        width: 461,
```
robertknight, line 43: Can you add a note about where these values come from. It looks like you captured the actual values from the client on your system?

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
```diff
@@ -0,0 +1,149 @@
...
+      const attr = document.body.getAttribute('data-side-by-side');
+      return attr && attr === 'false';
+    }, 50);
+  }
+
+  describe('initial render', () => {
+    it('does not lay out side-by-side', async () => {
+      renderComponent();
+
+      await sideBySideOff();
```
robertknight: This uses await - but here you want to check the state immediately after rendering, rather than waiting for it. See my notes about making the tests synchronous above.

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
```diff
@@ -0,0 +1,149 @@
...
+    });
+
+    it('provides room for sidebar controls', async () => {
+      renderComponent();
+
+      await sideBySideOff();
```
robertknight: This looks redundant to me, since the previous test checks the state after initial render.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,90 @@
+import { useCallback, useEffect, useState } from 'preact/hooks';
+
+type SidebarLayout = {
```
robertknight, line 6: Can you add a note about the source of these types. In future I think we could create a package, with types, for this purpose. In the YouTubePlayer component for example I added links to the relevant documentation pages.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,90 @@
...
+type LayoutChangeEventDetail = {
+  sideBySideActive: boolean;
+  sidebarLayout: SidebarLayout;
+};
+
+type LayoutChangeEvent = CustomEvent & { detail: LayoutChangeEventDetail };
+
+// Minimum body width at which laying out side-by-side should be considered
+export const SIDEBYSIDE_THRESHOLD = 875;
+// The sidebar reports its toolbar width as 33px. This is a custom padding
```
robertknight, line 26: How sensitive is this code to small changes in the toolbar width in future?

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,90 @@
...
+  const [sideBySideApplicable, setSideBySideApplicable] = useState(false);
+  const [sufficientSpace, setSufficientSpace] = useState(
+    document.body.clientWidth >= SIDEBYSIDE_THRESHOLD
+  );
+  const [sidebarWidth, setSidebarWidth] = useState(0);
+
+  const updateSidebarState = useCallback((e: Event) => {
+    const sidebarInfo: LayoutChangeEventDetail = (e as LayoutChangeEvent)
+      .detail;
+    const sideBySide = sidebarInfo.sideBySideActive;
```
robertknight: The `sideBySideActive` property should be ignored if the application is managing side-by-side itself. Currently it always returns `false` if the mode is set to `manual`. The way I have implemented the callback in https://github.com/hypothesis/client/pull/5623, it will web app's own callback to get this state.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,90 @@
...
+  }, []);
+
+  const onResize = () => {
+    setSufficientSpace(document.body.clientWidth >= SIDEBYSIDE_THRESHOLD);
+  };
+
+  // Make room for the sidebar at the right of the app layout when applicable.
+  // Otherwise, provide some room for the sidebar's controls. Padding is used
+  // here (versus margin) because it is part of an element's dimensions
+  // (`clientWidth`) and thus easier to compute with.
```
robertknight, line 61: Helpful comment :)

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,90 @@
...
+    const layout = sidebarInfo.sidebarLayout;
+    // Laying out side-by-side is applicable when the sidebar is open, and the
+    // client's own side-by-side mode is not active.
+    setSideBySideApplicable(layout.expanded && !sideBySide);
+    setSidebarWidth(layout.width);
+  }, []);
+
+  const onResize = () => {
+    setSufficientSpace(document.body.clientWidth >= SIDEBYSIDE_THRESHOLD);
+  };
```
robertknight: This helper can be pushed down into the only effect where it is used. This will avoid needing to add it as a dependency.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,90 @@
...
+    );
+    window.addEventListener('resize', onResize);
+    return () => {
+      document.body.removeEventListener(
+        'hypothesis:layoutchange',
+        updateSidebarState
+      );
+      window.removeEventListener('resize', onResize);
+    };
+  });
```
robertknight: This effect is missing dependencies, so it will re-run on every render, which I'm guessing was unintentional? I'm slightly surprised that the linter didn't say anything about that.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
robertknight: How are you suggesting `useReducer` would be used here? My immediate thought was that if too-frequent-updates were a problem, some debouncing could be added. There is a trade-off though as less frequent updates might cause more visually jumpy transitions when resizing.

In my testing the behavior seems fine as is though. We could also add debouncing of events on the client side.

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
```diff
@@ -0,0 +1,149 @@
+import { mount } from 'enzyme';
```
robertknight, line 1: These tests can actually be entirely synchronous, if the `act` function is used to wrap dispatches of events and make effects run synchronously. The upside is that this will make the tests run faster (by default effects run only on the next animation frame) and make them easier to reason about.

The downside is that if you added debouncing in future you'd need to add a way to make that happen synchronously in tests too.

```diff
diff --git a/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js b/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
index c7e5da6..834a9fd 100644
--- a/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
+++ b/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
@@ -1,6 +1,6 @@
 import { mount } from 'enzyme';
+import { act } from 'preact/test-utils';
 
-import { waitFor } from '../../../test-util/wait';
 import { useSideBySideLayout } from '../use-side-by-side-layout';
 import { TOOLBAR_WIDTH } from '../use-side-by-side-layout';
 
@@ -54,95 +54,78 @@ describe('useSideBySideLayout', () => {
     const event = new CustomEvent('hypothesis:layoutchange', {
       detail,
     });
-    document.body.dispatchEvent(event);
-  }
-
-  async function sideBySideOn() {
-    await waitFor(() => {
-      const attr = document.body.getAttribute('data-side-by-side');
-      return attr && attr === 'true';
-    }, 50);
+    act(() => {
+      document.body.dispatchEvent(event);
+    });
   }
 
-  async function sideBySideOff() {
-    await waitFor(() => {
-      const attr = document.body.getAttribute('data-side-by-side');
-      return attr && attr === 'false';
-    }, 50);
+  function sideBySideState() {
+    const attr = document.body.getAttribute('data-side-by-side');
+    if (attr === 'true') {
+      return true;
+    } else if (attr === 'false') {
+      return false;
+    } else {
+      return undefined;
+    }
   }
 
   describe('initial render', () => {
-    it('does not lay out side-by-side', async () => {
+    it('does not lay out side-by-side', () => {
       renderComponent();
-
-      await sideBySideOff();
+      assert.strictEqual(sideBySideState(), false);
     });
 
-    it('provides room for sidebar controls', async () => {
+    it('provides room for sidebar controls', () => {
       renderComponent();
-
-      await sideBySideOff();
-
+      assert.strictEqual(sideBySideState(), false);
       assert.equal(document.body.style.paddingRight, `${TOOLBAR_WIDTH}px`);
     });
   });
 
   describe('when sidebar opens', () => {
     [875, 876, 1024, 1268].forEach(bodyWidth => {
-      it('applies side-by-side layout when there is enough space', async () => {
+      it('applies side-by-side layout when there is enough space', () => {
         document.body.style.width = `${bodyWidth}px`;
         renderComponent();
         triggerEvent();
-
-        await sideBySideOn();
+        assert.isTrue(sideBySideState());
       });
     });
 
     [450, 600, 874].forEach(bodyWidth => {
-      it('allows sidebar to overlap content on narrow screens', async () => {
+      it('allows sidebar to overlap content on narrow screens', () => {
         document.body.style.width = `${bodyWidth}px`;
         renderComponent();
         triggerEvent();
-
-        await sideBySideOff();
+        assert.isFalse(sideBySideState());
       });
     });
-
-    it('does not apply side-by-side layout if client side-by-side layout already activated', async () => {
-      renderComponent();
-      triggerEvent();
-
-      await sideBySideOn();
-
-      // This indicates the client itself is handling side-by-side layout
-      fakeLayoutDetail.sideBySideActive = true;
-      triggerEvent();
-
-      return await sideBySideOff();
-    });
   });
 
   describe('when window is resized', () => {
-    it('recalculates side-by-side layout', async () => {
+    it('recalculates side-by-side layout', () => {
       document.body.style.width = '450px';
       renderComponent();
       triggerEvent();
 
-      await sideBySideOff();
+      assert.isFalse(sideBySideState());
 
-      document.body.style.width = '1450px';
-      window.dispatchEvent(new Event('resize'));
-      await sideBySideOn();
+      act(() => {
+        document.body.style.width = '1450px';
+        window.dispatchEvent(new Event('resize'));
+      });
+      assert.isTrue(sideBySideState());
     });
   });
 
   describe('when sidebar closes', () => {
-    it('does not provide room for sidebar at right', async () => {
+    it('does not provide room for sidebar at right', () => {
       renderComponent();
       fakeLayoutDetail.sidebarLayout.expanded = false;
       triggerEvent();
 
-      await sideBySideOff();
+      assert.isFalse(sideBySideState());
       assert.equal(document.body.style.paddingRight, `${TOOLBAR_WIDTH}px`);
     });
   });
```

robertknight (commented):
The user-facing behavior works well. I had a few comments on the implementation and tests.

- Can you add a link to the relevant docs (or other source) for types that refer to APIs defined outside the app (in the client, in this case)
- The `sideBySideActive` state on the `hypothesis:layoutchange` event should be ignored in apps using manual side-by-side
- There is an effect that is missing dependencies
- The tests can be made synchronous, which will make them run faster and a bit easier to reason about.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
lyzadanger: I was attempting to decouple this hook from the logic that sets the config (disables the side-by-side mode)—this hook can't be certain that that config has been set. Is there any _harm_ in having this check here? I could at least comment it better.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
lyzadanger: 🤦🏻‍♀️ Was so busy worrying about the other side-effect dependencies that I overlooked this one.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
lyzadanger: Mildly. I'm open to other suggestions on how to handle this. This adjustment makes it so the (custom) bucket bar is flush up against the sidebar's bucket channel/toolbar but allows the top tool buttons to overlap. If the width of the `[data-testid="sidebar-edge"]` element changed, we could end up with a small gap between it and the custom bucket bar.

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
lyzadanger: > If we can make these tests work with a temporary element (eg. by adding a test seam to the hook), 

I had already considered this and will do it now.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
robertknight: > Mildly. I'm open to other suggestions on how to handle this.

You've documented the assumptions the code is making here, so that's good at least. The client exposes a `toolbarWidth` property on the `hypothesis:layoutchange` event, so we could document that to make it official (in https://h.readthedocs.io/projects/client/en/latest/publishers/events.html) and then make use of it when we have it. Before we have that info, we can use a default.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
lyzadanger: I think this is fine as-is. I've tuned the state primitives to represent "meaningful state change": if any of these state values change, we legitimately do need to re-evaluate our side-by-side layout.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
robertknight: I think this could be handled in a follow-up PR though - it is very low-priority.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
robertknight: > Is there any harm in having this check here?

Yes. It will break when https://github.com/hypothesis/client/pull/5623 lands, unless we decide to change what this field in the event means (currently in that PR it means "does the client think side-by-side is active, regardless of who makes that decision" as opposed to "is the client's automatic side-by-side active") Some ways we could resolve that:

- Have the caller of this hook specify the side-by-side mode
- Just document the assumption that manual side-by-side mode is being used
- Add a field to the `hypothesis:layoutchange` event to indicate the mode

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
```diff
@@ -0,0 +1,136 @@
...
+import { act } from 'preact/test-utils';
+
+//import { waitFor } from '../../../test-util/wait';
```
robertknight: Commented-out code here.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Adjust the available space to the right of app content to make room for
+ * the sidebar when plausible. This mimics the client's "side-by-side" mode,
+ * tuned specifically for this application.
+ *
+ * Side-by-side layout will be re-computed whenever sidebar state via the
+ * `hypothesis:layoutchange` event changes meaningfully, or when the window
+ * is resized.
+ *
+ * @param [containerEl] test seam
```
robertknight:
```suggestion
 * @param [container] test seam
```

Parameter name is out of date

robertknight (approved):
LGTM, just two tiny nits.

PR #1067, opened 2023-07-11
Implement custom scroll anchoring for transcript container

### LINKED ISSUE #1021: Transcript scroll position lost when opening / closing sidebar in Safari
Opening and closing the sidebar in the video player can cause the displayed location of the transcript to shift significantly, if the browser does not support [scroll anchoring](https://caniuse.com/css-overflow-anchor). In my testing, Chrome's scroll anchoring seems to work well. Firefox's scroll anchoring did not always work the first time that the sidebar was opened, but did work after that. The main issue is in Safari which doesn't support scroll anchoring.

Sidebar closed, note visible timestamps:

Sidebar closed

After opening the sidebar, the visible range of timestamps changes significantly:

Sidebar opened

Tested with https://qa-via.hypothes.is/https://www.youtube.com/watch?v=MrORaCfcKxQ but should apply to any video.

acelaya:
This PR introduces a new `useScrollAnchor` hook that tries to mimic `overflow-anchor` for browsers that do not support it, with a combination of `ResizeObserver` and `scroll` event.

Since that can consume many resources, the logic is made in a way that it is skipped if the browser natively supports the feature.

### Todo

- [x] Add tests
- [x] Fix cases in which calculation is messed-up. It happens as you get closer to the bottom of the scroll.
- [x] Remove hardcoded `overflow-anchor: none` used to test this on any browser.
- [x] Un-comment check used to verify if `overflow-anchor` is natively supported.

### Testing steps

Using a browser which does not support overflow anchoring (Safair), go to http://localhost:9083/https://www.youtube.com/watch?v=MrORaCfcKxQ and resize your window to try to find a resolution in which the transcript has a different size when the sidebar is open and closed.

One good way to test it is in a resolution where the transcript is moved under the video when the sidebar is open, like this:

[Grabación de pantalla desde 2023-07-11 12-47-21.webm](https://github.com/hypothesis/via/assets/2719332/83e9e498-16f8-49cc-aa63-24a497942f3e)

Once you are there, follow these steps:

1. Scroll down in the transcript, and check what's the first visible segment.
2. Open the sidebar. The first visible segment should still be the same.
3. Scroll again, and then close the sidebar.
4. Again, the first segment should be the same as before closing it.

If you try the same on `main` branch, you'll see the segments can change significantly.

> This PR fixes https://github.com/hypothesis/via/issues/1021

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -9,6 +9,7 @@ import {
   useRef,
 } from 'preact/hooks';
 
+import { useScrollAnchor } from '../hooks/use-scroll-anchor';
 import { formatTimestamp } from '../utils/time';
 import type { MatchOffset, Segment, TranscriptData } from '../utils/transcript';
 import { filterTranscript } from '../utils/transcript';
@@ -346,6 +347,15 @@ export default function Transcript({
     }
     return filterTranscript(transcript.segments, filter);
   }, [filter, transcript]);
+
+  // Adjust the scroll position when the transcript container is resized, so the
+  // user doesn't lose their place. See
+  // https://github.com/hypothesis/via/issues/1021.
+  const getScrollChildren = useCallback(
+    (element: HTMLElement) => element.querySelectorAll('li'),
+    []
+  );
+  useScrollAnchor(scrollRef, getScrollChildren);
 
   return (
     <ScrollContainer borderless>
--- a/via/static/scripts/video_player/hooks/test/use-scroll-anchor-test.js
+++ b/via/static/scripts/video_player/hooks/test/use-scroll-anchor-test.js
@@ -1,0 +1,138 @@
+import { mount } from 'enzyme';
+import { useCallback, useRef } from 'preact/hooks';
+
+import { useScrollAnchor } from '../use-scroll-anchor';
+
+const loremIpsum =
+  'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.';
+
+describe('useScrollAnchor', () => {
+  // Scrollable test widget with children that have text wrapping onto
+  // multiple lines. As the width of the container is varied, the
+  // `useScrollAnchor` hook will need to adjust the scroll position to keep
+  // the same child visible.
+  function TestWidget({ width = 200, height = 200 }) {
+    const container = useRef(null);
+    const getChildren = useCallback(
+      () => container.current.querySelectorAll('.child'),
+      []
+    );
+
+    useScrollAnchor(container, getChildren);
+
+    return (
+      <div
+        data-testid="container"
+        ref={container}
+        style={{
+          width: `${width}px`,
+          height: `${height}px`,
+
+          // Hide scrollbars so the `clientWidth` of this element is the same
+          // on all platforms.
+          overflowY: 'hidden',
+        }}
+      >
+        <div className="child">{loremIpsum}</div>
+        <div className="child">{loremIpsum}</div>
+        <div className="child">{loremIpsum}</div>
+        <div className="child">{loremIpsum}</div>
+        <div className="child">{loremIpsum}</div>
+      </div>
+    );
+  }
+
+  let wrappers;
+
+  const renderWidget = props => {
+    const wrapper = mount(<TestWidget {...props} />, {
+      attachTo: document.body,
+    });
+    wrappers.push(wrapper);
+    return wrapper;
+  };
+
+  /** Return Y offset of top of `child` from top of `container`. */
+  function offsetFromTop(container, child) {
+    const containerRect = container.getBoundingClientRect();
+    const childRect = child.getBoundingClientRect();
+    return childRect.top - containerRect.top;
+  }
+
+  function waitForEvent(element, event) {
+    return new Promise(resolve => {
+      element.addEventListener(event, resolve, { once: true });
+    });
+  }
+
+  function waitForResize(element) {
+    return new Promise(resolve => {
+      const ro = new ResizeObserver(() => {
+        resolve();
+        ro.disconnect();
+      });
+      ro.observe(element);
+    });
+  }
+
+  beforeEach(() => {
+    wrappers = [];
+  });
+
+  afterEach(() => {
+    wrappers.forEach(w => w.unmount());
+  });
+
+  it('disables built-in scroll anchoring', () => {
+    const wrapper = renderWidget();
+    assert.equal(
+      wrapper.find('[data-testid="container"]').getDOMNode().style
+        .overflowAnchor,
+      'none'
+    );
+  });
+
+  it('adjusts scroll position when container is resized', async () => {
+    const wrapper = renderWidget({ width: 150 });
+    const container = wrapper.find('[data-testid="container"]').getDOMNode();
+
+    // Scroll down enough that, when we make the container wider below, the
+    // content will shift.
+    const scrolled = waitForEvent(container, 'scroll');
+    container.scrollBy(0, 50);
+    await scrolled;
+
+    // Pick the second child for testing. This is the first child that fits
+    // within the viewport and so the one which `useScrollAnchor` will pick as
+    // a scroll anchor.
+    const child = wrapper.find('.child').at(1).getDOMNode();
+    const topOffset = offsetFromTop(container, child);
+    const scrollOffset = container.scrollTop;
+
+    // Make the container wider. The item contents will now wrap onto fewer
+    // lines and so if the scroll position of the container is not updated by
+    // `useScrollAnchor`, the content would shift.
+    const resized = waitForResize(container);
+    wrapper.setProps({ width: 200 });
+    await resized;
+
+    // Check that `useScrollAnchor` did adjust the `scrollTop` of the container
+    // so as to keep the relative offset of the child the same.
+    const topOffset2 = offsetFromTop(container, child);
+    const scrollOffset2 = container.scrollTop;
+
+    assert.notEqual(scrollOffset, scrollOffset2);
+    assert.equal(topOffset, topOffset2);
+
+    // Wait for a repaint. Otherwise we'll get a ResizeObserver loop error
+    // when resizing a second time below.
+    await new Promise(resolve => requestAnimationFrame(resolve));
+
+    // Resize to zero width/height and trigger a scroll. This exercises code
+    // paths for the case where no scroll anchor can be chosen.
+    const resized2 = waitForResize(container);
+    wrapper.setProps({ width: 0, height: 0 });
+    container.dispatchEvent(new Event('scroll'));
+    await resized2;
+  });
+});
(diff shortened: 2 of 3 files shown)


via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -347,9 +348,12 @@ export default function Transcript({
     return filterTranscript(transcript.segments, filter);
   }, [filter, transcript]);
 
+  useOverflowAnchor(scrollRef, element => element.querySelectorAll('li'));
+
   return (
     <ScrollContainer borderless>
       <Scroll
+        style={{ overflowAnchor: 'none' }} // TODO Remove this!
```
acelaya: This is used just to test this on any browser. To be removed before this PR is ready.

via/static/scripts/video_player/hooks/use-overflow-anchor.ts
```diff
@@ -0,0 +1,94 @@
...
+) {
+  const firstChildInViewport = useRef<ChildInViewport>({
+    element: null,
+    offsetTop: 0,
+  });
+
+  useEffect(() => {
+    // if (CSS.supports('overflow-anchor: auto')) {
+    //   return () => {};
+    // }
```
acelaya: This will be un-commented before the PR is ready.

via/static/scripts/video_player/hooks/use-overflow-anchor.ts
```diff
@@ -0,0 +1,94 @@
...
+    return false;
+  }
+
+  return (
+    rectB.left >= rectA.left &&
+    rectB.right <= rectA.right &&
+    rectB.top >= rectA.top &&
+    rectB.bottom <= rectA.bottom
+  );
+}
```
acelaya: `rectContains` and `rectIsEmpty` are copied from `client`.

via/static/scripts/video_player/hooks/use-overflow-anchor.ts
```diff
@@ -0,0 +1,94 @@
...
+
+type ChildInViewport = {
+  element: HTMLLIElement | null;
+  offsetTop: number;
+};
+
+export function useOverflowAnchor(
+  scrollRef: MutableRef<HTMLElement | null>,
+  getChildren: (element: HTMLElement) => NodeListOf<HTMLElement>
```
acelaya: This callback let's consuming code decide how the children elements should be resolved, without making any assumptions at the hook level.

via/static/scripts/video_player/hooks/use-overflow-anchor.ts
```diff
@@ -0,0 +1,94 @@
...
+  useEffect(() => {
+    // if (CSS.supports('overflow-anchor: auto')) {
+    //   return () => {};
+    // }
+
+    const scrollElement = scrollRef.current!;
+    let scrollRect = scrollElement.getBoundingClientRect();
+
+    const observer = new ResizeObserver(() => {
+      scrollRect = scrollElement.getBoundingClientRect();
```
acelaya: We recalculate the scrollElement's DOMRect only when it is resized. That way we can use a precalculated value on the scroll listener, which is run more frequently.

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -347,9 +348,12 @@ export default function Transcript({
     return filterTranscript(transcript.segments, filter);
   }, [filter, transcript]);
 
+  useOverflowAnchor(scrollRef, element => element.querySelectorAll('li'));
```
robertknight: Don't forget to memo the callback, otherwise the effects internally will re-run on every render.

robertknight: The browser feature that we are emulating here is called [scroll anchoring](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring) rather than _overflow anchoring_.

robertknight: The CSS properties that controls the behavior all have an `overflow-` prefix, which is a bit confusing. I'm guessing that was done for consistency with the existing `overflow-` properties that relate to scrolling.

via/static/scripts/video_player/hooks/use-overflow-anchor.ts
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    rectB.left >= rectA.left &&
+    rectB.right <= rectA.right &&
+    rectB.top >= rectA.top &&
+    rectB.bottom <= rectA.bottom
+  );
+}
+
+type ChildInViewport = {
+  element: HTMLLIElement | null;
```
robertknight: `getChildren` returns a list of `HTMLElement`, which may not be an `li`, but the element is recorded as an `HTMLLIElement` here.

acelaya: Oh, this is an oversight from when I extracted everything to a hook. Good catch.

via/static/scripts/video_player/components/Transcript.tsx
acelaya: Oh, I see. Yeah, I got confused by the name of the CSS property.

robertknight:
This is mostly working for me in Safari. I did encounter an issue where it sometimes ends up at the wrong location after opening and closing the sidebar. The issue seems to relate to being scrolled nearer to the end. I can reproduce at http://localhost:9083/https://www.youtube.com/watch?v=x8TO-nrUtSI using these steps:

1. Close sidebar
2. Scroll all the way to the bottom of the transcript
3. Open the sidebar
4. Close the sidebar

After step 4 the scroll offset has changed. The exact location I end up at is different in Chrome and Firefox compared to Safari, but in all three the view is scrolled up.

acelaya:
> This is mostly working for me in Safari. I did encounter an issue where it sometimes ends up at the wrong location after opening and closing the sidebar. The issue seems to relate to being scrolled nearer to the end

I'll do some extra testing of edge cases, like when you are at the beginning or end, when there's an active filter and therefore no scroll, etc.

acelaya:
After a lot of debugging, these are my findings:

* When opening/closing the sidebar, the ResizeObserver is sometimes invoked twice. I have seen cases in which, during one of those, the scroll element is computed as 0x0 size, specially if the resize involves transitioning from 1 column to 2 columns.
* Resizing the scroll element also triggers scroll on it, which might require re-computing the first visible child. When that happens while the scroll element is 0x0, no children are considered visible, messing up follow-up calculations.
* When there's enough space to keep 2 columns while opening/closing the sidebar, the overall logic works much better, except when scroll is at the very bottom, where the logic also doesn't work. I'm still not sure if this is unrelated from the issue above, or also somewhat related.

I think some/all of those are somehow related with the flex layout, but haven't been able to find a working combination yet.

robertknight:
> Resizing the scroll element also triggers scroll on it

Does the scroll happen before or after the resize? I expect there are ways we could filter scroll events generated by resizes. For example we could ignore scroll events if the widget has recently been resized (this assumes we get the resize notification first).

A more specialized approach we could take would be to capture the scroll position only when the content layout is about to change. Once @lyzadanger's work in https://github.com/hypothesis/via/pull/1063 is completed, the video player app will be applying these content layout changes itself, and therefore know when the transcript container's layout is about to change.

acelaya:
> Does the scroll happen before or after the resize? I expect there are ways we could filter scroll events generated by resizes. For example we could ignore scroll events if the widget has recently been resized (this assumes we get the resize notification first).

The thing is that I'm not completely sure we actually want to ignore them.

If you start to resize the window, for example, it will start triggering scroll and resize events "in a loop", so that it can properly retain the scroll position. This is the desired behavior, so I think it's correct.

The only problem happens when there's a "weird" size change in between. For example, when it goes to `0px` between two "real" sizes.

I have tried ignoring these, but you need to make sure both the resize and the scroll events are ignored when this happens, so that this "step" is "skipped" (so to speak), but since they are different callbacks, it requires tracking it somehow.

It also doesn't cover the third bullet point, unless I find a way to detect a common pattern which is not "it goes to zero size" but "it transitions to an impossible size due to external layout".

robertknight:
> If you start to resize the window, for example, it will start triggering scroll and resize events "in a loop", so that it can properly retain the scroll position. This is the desired behavior, so I think it's correct.

In that case though the final scroll position we want, expressed in terms of which segment is at the top of the transcript, should be the same as before the user initiated the sequence of resize events.

If it simplifies things at all, I think it would be acceptable if resizing only preserved which segment was at the top of the viewport, but not the exact offset of that segment from the top of the screen. The main problem here is when the transcript scrolls by say, 30+ segments after the sidebar is opened/closed.

acelaya:
> If it simplifies things at all, I think it would be acceptable if resizing only preserved which segment was at the top of the viewport, but not the exact offset of that segment from the top of the screen.

I think I tried this. Just save the first segment in the viewport, and on resize, get its new `offsetTop` and just scroll to it instead of calculating the difference, but in the `0x0` case, there's none inside of the viewport, so it doesn't make a difference.

I'm thinking though, we could just "skip" this case and keep the previous top segment when no segment can be "found" in the viewport after scroll, but I believe I tried this and there was something else... Let me try again.

acelaya:
> > If it simplifies things at all, I think it would be acceptable if resizing only preserved which segment was at the top of the viewport, but not the exact offset of that segment from the top of the screen.
> 
> I think I tried this. Just save the first segment in the viewport, and on resize, get its new `offsetTop` and just scroll to it instead of calculating the difference, but in the `0x0` case, there's none inside of the viewport, so it doesn't make a difference.
> 
> I'm thinking though, we could just "skip" this case and keep the previous top segment when no segment can be "found" in the viewport after scroll, but I believe I tried this and there was something else... Let me try again.

Ok, ☝🏼 this does actually fix it. A combination of: 

* Just keep the upper-most visible segment and scroll to its new offsetTop on resize.
* Do nothing on scroll if an element could not be computed inside the viewport.

It covers all the bullet points. What a pain 😅

robertknight:
I have re-worked this PR and added tests so that we have an MVP we can land. I retained the overall structure of the original, but changed it as follows:

1. The implementation now avoids doing any scroll adjustments if no child elements fit wholly within the scroll container's viewport. This makes the code simpler and easier to test, at the cost of losing the scroll position in some cases, but none that are IMO critical here.
2. The implementation now compares the viewport rect (`getBoundingClientRect`) of the container and child elements to compute the offset for the scroll anchor element, so it isn't reliant on the container being the `offsetParent` of the child.
3. We now _always_ use the custom scroll anchoring for the transcript container. This ensures consistency across browsers, and also Firefox's didn't always work perfectly in my testing.

The main scenario where issue (1) comes up in the current layout is if the window is short and just narrow enough to trigger a single column layout. In that case the transcript container can end up being zero height. We should really fix this in the app layout so that it doesn't happen, as the app is not usable in that state.

robertknight:
I filed an issue at https://github.com/hypothesis/via/issues/1085 about the app layout in wide-but-short windows.

acelaya:
> The main scenario where issue (1) comes up in the current layout is if the window is short and just narrow enough to trigger a single column layout. In that case the transcript container can end up being zero height. We should really fix this in the app layout so that it doesn't happen, as the app is not usable in that state.

This was one of my main headaches. It does make sense we figure out what is making it have that in-between zero-height state, when transitioning from one to two columns.

acelaya:
@robertknight consider this PR approved from my side 👍🏼

robertknight:
Thanks. Will get this merged on Monday!

PR #1070, opened 2023-07-11
Use `hypothesisConfig` global to pass configuration to the client

robertknight:
This enables us to pass configuration to the client which is not JSON-serializable, such as callbacks.

The immediate use case I'm thinking of is allowing the video player app to pass an "is side-by-side active?" callback to the client ([slack thread](https://hypothes-is.slack.com/archives/C1M8NH76X/p1689065768678309?thread_ts=1689065493.060469&cid=C1M8NH76X)) but it has come up in other contexts that it may be useful to pass non-JSON serializable config. If in future we support [loading Hypothesis via ESM import](https://github.com/hypothesis/client/issues/5569) that will provide a much more direct way to pass arbitrary config to the client.

### DIFF
--- a/via/static/scripts/video_player/components/HypothesisClient.tsx
+++ b/via/static/scripts/video_player/components/HypothesisClient.tsx
@@ -1,7 +1,18 @@
-import { useEffect } from 'preact/hooks';
+import { useLayoutEffect } from 'preact/hooks';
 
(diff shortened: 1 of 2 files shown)


acelaya (approved):
Looks good

robertknight:
FYI @lyzadanger

PR #1072, opened 2023-07-11
Cache YouTube transcripts in the DB

### LINKED ISSUE #927: Save transcripts in the DB
**Problem:** Via currently [calls the YouTube API to get the video transript](https://github.com/hypothesis/via/blob/a3179570411b9c83ca2329f37e11e816d9958186/via/services/youtube.py#L78-L85) every time someone loads Via's YouTube video player page. This risks getting us blocked by YouTube for making too many requests to their API; it makes page loads slower than necessary (because every page load must wait for network requests to the YouTube API); and it makes page loads more unreliable than necessary (whenever we make network requests to a third-party service there's a chance this request can fail). It also means that if our code for getting transcripts from the YouTube API breaks (say YouTube change or remove their API) then all existing video assignments will be broken as each assignment load will crash when trying to get the transcript.

**Solution:** Via's `get_transcript()` method should save the video title in Via's DB after retrieving it from the YouTube API. The next time `get_transcript()` is called, if the transcript for this particular video is already in the DB, it should return the title from the DB and not make a YouTube API request.

### Note on DB keys

A single YouTube video can have multiple transcripts, for example transcripts in different languages, so we can't just use the video ID as the key when saving transcripts in the DB. Instead we need to use `(video_id, transcript_id)` where `transcript_id` is some sort of transcript identifier.

The way the code currently works (based on the [youtube-transcript-api](https://github.com/jdepoix/youtube-transcript-api) library), is that transcript IDs are simple language codes like `"en"`, for example:

```python
def get_transcript(self, video_id):
    return YouTubeTranscriptApi.get_transcript(video_id, languages=("en",))
```

So DB keys will be a combination of a YouTube video ID and a language code, e.g. `{"video_id": "Bw8b9YV0EPA", "transcript_id": "en"}` (and the transcript will be whatever a successful call to `YouTubeTranscriptApi.get_transcript("Bw8b9YV0EPA", languages=("en",))` returns.

### Known Issues

* We're planning to replace the third-party `youtube-transcript-api` library with our own code and our code will probably use a different form of transcript ID than the language codes that `youtube-transcript-api`'s uses (e.g. `"en"`): this is a flaw in `youtube-transcript-api`, a single YouTube video can have multiple transcripts with the same language code (see https://github.com/jdepoix/youtube-transcript-api/issues/150). When we change the transcript IDs we'll end up invalidating any existing cached transcripts in Via's DB as these will use the old-style IDs and so the lookup code will no longer find them.

  This is fine: the transcripts for these videos will simply be re-fetched once.

  We could do extra work to migrate the IDs of these transcripts in the DB but I doubt it's worth the effort: we can just let Via re-fetch them.

  We could also do extra work to delete the invalidated transcripts in the DB but again I don't think it's worth the effort.

* When we replace `youtube-transcript-api` we may also inadvertently change the transcripts of existing assignments from one English transcript to another, if the assignment is for a YouTube video that has multiple English transcripts. This is because our new code will pick a different English transcript by default than what `youtube-transcript-api` picks. This is fine: we discussed it in a project meeting (Thu 20th July 2023, Sean, Marcos, Rob, Alejandro, Nairi, Lee, Dan) and decided to accept the one-off risk of changing the transcripts of some existing assignments.

seanh:
Depends on https://github.com/hypothesis/via/pull/1019. Fixes https://github.com/hypothesis/via/issues/927.

Save YouTube transcripts in the DB and re-serve them from the DB rather than requesting them from the YouTube API again.

Once a transcript has been saved once it'll always be re-served from the DB and never fetched again. We might want to periodically re-fetch transcripts, but we can consider that separately.

### Test `models.Transcript` in the shell

```python
$ make shell
>>> transcript = models.Transcript(video_id="VIDEO_ID", transcript_id="TRANSCRIPT_ID", transcript=[])
>>> db.add(transcript)
>>> tm.commit()
>>> tm.begin()
>>> db.scalars(select(models.Transcript)).all()
[Transcript(id=1, video_id='VIDEO_ID', transcript_id='TRANSCRIPT_ID')] 
```

### Test `TranscriptFactory` in the shell

```python
$ make shell
>>> # You could actually add this to the DB and commit it if you wanted to.
>>> factories.TranscriptFactory()
Transcript(id=None, video_id='video_id_0', transcript_id='transcript_id_0')
>>> factories.TranscriptFactory()
Transcript(id=None, video_id='video_id_1', transcript_id='transcript_id_1')
```

### Test the transcript API

First get a bearer token for calling the API:

```python
$ make shell
>>> from via.security import ViaSecurityPolicy
>>> ViaSecurityPolicy.encode_jwt(request)
'eyJ***Mt4'
```

Now run `make dev` and in another terminal use [HTTPie](https://httpie.io/) to make requests to the transcript API using the bearer token:

```terminal
$ http GET 'http://localhost:9083/api/youtube/transcript/TbzZIMQC6vk' Authorization:"Bearer eyJ***Mt4"
...
```

You should notice that if you run this command again it's much faster the second time because it's returning the cached transcript from Postgres. If you run `make sql` you can inspect the cached transcript in the `transcript` table:

```terminal
$ make sql
postgres=# select * from transcript;
...
```

### Test the Web UI

<http://localhost:9083/https://www.youtube.com/watch?v=byN4S8WDPt8>

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,14 +3,19 @@ from os import environ
 import httpretty
 import pytest
 from h_matchers import Any
+from pytest_factoryboy import register
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
+from tests.factories import TranscriptFactory
 from tests.factories.factoryboy_sqlalchemy_session import (
     clear_factoryboy_sqlalchemy_session,
     set_factoryboy_sqlalchemy_session,
 )
 from via.db import Base
+
+# Each factory has to be registered with pytest_factoryboy.
+register(TranscriptFactory)
 
 
 @pytest.fixture
--- a/tests/factories/transcript.py
+++ b/tests/factories/transcript.py
@@ -1,0 +1,29 @@
+from factory import Sequence
+from factory.alchemy import SQLAlchemyModelFactory
+
+from via.models import Transcript
+
+
+class TranscriptFactory(SQLAlchemyModelFactory):
+    class Meta:
+        model = Transcript
+
+    video_id = Sequence(lambda n: f"video_id_{n}")
+    transcript_id = Sequence(lambda n: f"transcript_id_{n}")
+    transcript = [
+        {
+            "text": "[Music]",
+            "start": 0.0,
+            "duration": 7.52,
+        },
+        {
+            "text": "how many of you remember the first time",
+            "start": 5.6,
+            "duration": 4.72,
+        },
+        {
+            "text": "you saw a playstation 1 game if you were",
+            "start": 7.52,
+            "duration": 4.72,
+        },
+    ]
--- a/via/models/_mixins.py
+++ b/via/models/_mixins.py
@@ -1,0 +1,26 @@
+from datetime import datetime
+
+from sqlalchemy import func
+from sqlalchemy.orm import Mapped, MappedAsDataclass, mapped_column
+
+
+class AutoincrementingIntegerIDMixin(MappedAsDataclass):
+    id: Mapped[int] = mapped_column(
+        init=False, primary_key=True, autoincrement=True, sort_order=-100
+    )
+
+
+class CreatedUpdatedMixin(MappedAsDataclass):
+    created: Mapped[datetime] = mapped_column(
+        init=False,
+        repr=False,
+        server_default=func.now(),  # pylint:disable=not-callable
+        sort_order=-10,
+    )
+    updated: Mapped[datetime] = mapped_column(
+        init=False,
+        repr=False,
+        server_default=func.now(),  # pylint:disable=not-callable
+        onupdate=func.now(),  # pylint:disable=not-callable
+        sort_order=-10,
+    )
--- a/via/models/transcript.py
+++ b/via/models/transcript.py
@@ -1,0 +1,15 @@
+from sqlalchemy import UniqueConstraint
+from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy.orm import Mapped, mapped_column
+
+from via.db import Base
+from via.models._mixins import AutoincrementingIntegerIDMixin, CreatedUpdatedMixin
+
+
+class Transcript(AutoincrementingIntegerIDMixin, CreatedUpdatedMixin, Base):
+    __tablename__ = "transcript"
+    __table_args__ = (UniqueConstraint("video_id", "transcript_id"),)
+
+    video_id: Mapped[str]
+    transcript_id: Mapped[str]
+    transcript: Mapped[list] = mapped_column(JSONB, repr=False)
--- a/via/services/youtube.py
+++ b/via/services/youtube.py
@@ -1,7 +1,10 @@
 from urllib.parse import parse_qs, quote_plus, urlparse
 
+from sqlalchemy import select
+from sqlalchemy.exc import NoResultFound
 from youtube_transcript_api import YouTubeTranscriptApi
 
+from via.models import Transcript
 from via.services.http import HTTPService
 
 
@@ -10,7 +13,10 @@ class YouTubeDataAPIError(Exception):
 
 
 class YouTubeService:
-    def __init__(self, enabled: bool, api_key: str, http_service: HTTPService):
+    def __init__(
+        self, db_session, enabled: bool, api_key: str, http_service: HTTPService
+    ):
+        self._db = db_session
         self._enabled = enabled
         self._api_key = api_key
         self._http_service = http_service
@@ -82,11 +88,37 @@ class YouTubeService:
         :raise Exception: this method might raise any type of exception that
             YouTubeTranscriptApi raises
         """
-        return YouTubeTranscriptApi.get_transcript(video_id)
+        transcript_id = language_code = "en"
+
+        try:
+            transcript = (
+                self._db.scalars(
+                    select(Transcript).where(
+                        Transcript.video_id == video_id,
+                        Transcript.transcript_id == transcript_id,
+                    )
+                )
+                .one()
+                .transcript
+            )
+        except NoResultFound:
+            transcript = YouTubeTranscriptApi.get_transcript(
+                video_id, languages=(language_code,)
+            )
+            self._db.add(
+                Transcript(
+                    video_id=video_id,
+                    transcript_id=transcript_id,
+                    transcript=transcript,
+                )
+            )
+
+        return transcript
 
 
 def factory(_context, request):
     return YouTubeService(
+        db_session=request.db,
         enabled=request.registry.settings["youtube_transcripts"],
         api_key=request.registry.settings["youtube_api_key"],
         http_service=request.find_service(HTTPService),
--- a/tests/factories/__init__.py
+++ b/tests/factories/__init__.py
@@ -1,0 +1,1 @@
+from tests.factories.transcript import TranscriptFactory
--- a/tests/unit/via/services/youtube_test.py
+++ b/tests/unit/via/services/youtube_test.py
@@ -2,8 +2,11 @@ from io import BytesIO
 from unittest.mock import sentinel
 
 import pytest
+from h_matchers import Any
 from requests import Response
+from sqlalchemy import select
 
+from via.models import Transcript
 from via.services.youtube import YouTubeDataAPIError, YouTubeService, factory
 
 
@@ -17,9 +20,10 @@ class TestYouTubeService:
             (True, sentinel.api_key, True),
         ],
     )
-    def test_enabled(self, enabled, api_key, expected):
+    def test_enabled(self, db_session, enabled, api_key, expected):
         assert (
             YouTubeService(
+                db_session=db_session,
                 enabled=enabled,
                 api_key=api_key,
                 http_service=sentinel.http_service,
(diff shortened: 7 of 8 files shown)


tests/conftest.py
```diff
@@ -1,6 +1,12 @@
 import httpretty
 import pytest
 from h_matchers import Any
+from pytest_factoryboy import register
+
+from tests.factories import TranscriptFactory
+
+# Each factory has to be registered with pytest_factoryboy.
+register(TranscriptFactory)
```
seanh, line 9: No way around this AFAICT. `pytest_factoryboy` comes with an `@register` decorator that you can just put on the factory classes where you define them, but it doesn't seem to work.

tests/factories/transcript.py
```diff
@@ -0,0 +1,28 @@
...
+            "text": "how many of you remember the first time",
+            "start": 5.6,
+            "duration": 4.72,
+        },
+        {
+            "text": "you saw a playstation 1 game if you were",
+            "start": 7.52,
+            "duration": 4.72,
+        },
+    ]
```
seanh, line 28: This factory is already useful as-is but in future we'll probably be adding more fields to transcripts (e.g. maybe a language code) so it'll become even more useful to have a factory for generating realistic instances

via/models/transcript.py
```diff
@@ -0,0 +1,12 @@
...
+
+from via.db import Base
+
+
+class Transcript(Base):
+    __tablename__ = "transcript"
+
+    id: Mapped[int] = mapped_column(primary_key=True, autoincrement=True, init=False)
+    video_id: Mapped[str] = mapped_column(unique=True)
+    transcript: Mapped[list] = mapped_column(JSONB, repr=False)
```
seanh: We should probably add `created` and `updated` columns to this.

jon-betts: We could grab the mixin for that.

Looking ahead, do we want to namespace the transcripts in anyway to account for the possibility of non-Youtube videos which might clash?

We could have a Python managed enum for this (which we also have a mixin for I think?). If we are storing huge transcripts saving a few bytes on the type is probably not worth a properly normalized table.

via/services/youtube.py
```diff
@@ -58,8 +63,22 @@ def get_transcript(self, video_id):
...
+        try:
+            transcript = (
+                self._db.scalars(
+                    select(Transcript).where(Transcript.video_id == video_id)
+                )
+                .one()
+                .transcript
+            )
+        except NoResultFound:
+            transcript = YouTubeTranscriptApi.get_transcript(video_id)
```
jon-betts: I think we might want to process the transcript we get from this API before we store it for two reasons. The first is to vet we are getting the shape we expect, and the second is to try and keep the size down.

Looking at the docs we get back something like this:

```python
[
    {
        'text': 'Hey there',
        'start': 7.58,
        'duration': 6.13
    },
    {
        'text': 'how are you',
        'start': 14.08,
        'duration': 7.58
    },
    ...
]
```

There's quite a lot of repetition in the keys there. You can automatically deal with that with something like [ZSON](https://github.com/postgrespro/zson), but I've not used it and I don't think we can put custom plugins in our RDS Postgres instances. In any case it's potentially annoying and fragile.

Much simpler would be to convert this to a more compact tuple format like:

```python
[
    [7.58, 6.13, 'Hey there'],
    [14.08, 7.58, 'how are you'],
    ...
]
```

I think this will save us quite a decent amount of storage space. If we also used this as our API format it would also save us and our users bandwidth when transmitting it.

I looked at whether it might make sense to use something like zlib to compress the string before we send it, but Postgres will store big strings like this in TOAST storage, which is already compressed. It's not great compression, but it's totally transparent to us and already priced in.

jon-betts: I tried this out on a random hour long YouTube video and the results of the JSON string length were:

 * Dict format: 136,664
 * Tuple format: 90,235
 * Zlib: 33,186 (tuple) / 35,696 (dict)
 
This is about 2/3 of the original size for the tuple. Zlibing it is even better at about 25% of the original size. The code isn't much, but will take a little CPU time:

```python
import zlib

zlib.compress(json.dumps(transcript).encode('utf-8'), level=9)
```

It's possible we could serve this content directly by declaring it's content encoding as zlib.

These are decent savings, but it's pre-postgres. To really know you'd want to store both in the DB 1000 times and see what happens to the TOAST table size in Postgres is done messing with it. In either case, it would mean a bandwidth reduction for sure if we adopted it as the API format.

via/models/transcript.py
seanh: > We could grab the mixin for that.

I wonder if SQLAlchemy 2 has a different way but yeah probably.

> Looking ahead, do we want to namespace the transcripts in anyway to account for the possibility of non-Youtube videos which might clash?

Yeah, maybe we should add a `service` column or something whose value would be `"youtube"`, and the unique constraint would be on `(service, video_id)`. I think we can always do this later if/when we need it, we don't need to do it now.

> We could have a Python managed enum for this (which we also have a mixin for I think?). If we are storing huge transcripts saving a few bytes on the type is probably not worth a properly normalized table.

Again I wonder what the SQLAlchemy 2 way might be here

via/services/youtube.py
seanh: > I think we might want to process the transcript we get from this API before we store it for two reasons. The first is to vet we are getting the shape we expect, and the second is to try and keep the size down.

I don't think there's much point in the backend validating the transcripts that we get from the YouTube API. All the backend does with these transcripts is blindly pass them to the frontend. If a transcript isn't in the shape that the frontend expects then I think it'd be up to the frontend to detect that and show an error. After all there's not much the backend could do if it detected an invalid transcript: it could just send an error to the frontend which would then have to detect that API error and show an error anyway.

It might have the advantage of potentially preventing us from storing an invalid transcript in the DB? But then again that'd have the downside of disabling our caching and suddenly causing a spike in how many requests we're sending to YouTube if the format of transcripts that YouTube is returning changes from what our backend expects and so we stop caching them.

Regarding compression and saving on bandwidth: isn't that applied at the HTTP level? Like by NGINX? I guess we could potentially save on Postgres disk space but again that seems like something that Postgres itself might already do.

seanh: Oh I see, you're actually talking about reformatting the transcripts from dicts into tuples. Hmm. It's more complicated because currently the transcripts are just opaque strings to the backend, and then just the frontend needs to understand YouTube's transcript format. With reformatting the backend would need to understand YouTube's transcript format and then the frontend would need to understand the backend's reformatted transcript format. On the other hand, it sounds like it could save us storage space and bandwidth even though we already have generic compression.

We'd have to coordinate such a change with the already-existing frontend code that expects the current transcript format.

tests/conftest.py
```diff
@@ -3,15 +3,20 @@
...
 
+from tests.factories import TranscriptFactory
 from tests.factories.factoryboy_sqlalchemy_session import (
     clear_factoryboy_sqlalchemy_session,
     set_factoryboy_sqlalchemy_session,
 )
 from via.db import Base
 
+# Each factory has to be registered with pytest_factoryboy.
+register(TranscriptFactory)
```
seanh, line 18: No way around this AFAICT. `pytest_factoryboy` comes with an `@register` decorator that you can just put on the factory classes where you define them, but it doesn't seem to work.

via/models/transcript.py
seanh: I've added a `CreatedUpdatedMixin` for the `created` and `updated` columns.

I'm not gonna bother namespacing the transcripts yet (e.g. with a `type="YOUTUBE"` column): we can always add that column in future if we add a second transcript source, but that may never happen. This avoids having to address the question of enums and how best to do them for now.

via/services/youtube.py
seanh: We [decided](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1690279983515129) to postpone any pre-processing or optimisation of transcripts for maybe later

tests/factories/transcript.py
```diff
@@ -0,0 +1,29 @@
...
+
+from via.models import Transcript
+
+
+class TranscriptFactory(SQLAlchemyModelFactory):
+    class Meta:
+        model = Transcript
+
+    video_id = Sequence(lambda n: f"video_id_{n}")
+    transcript_id = Sequence(lambda n: f"transcript_id_{n}")
```
seanh, line 12: Faker can actually [generate random langauge codes](https://faker.readthedocs.io/en/master/providers/baseprovider.html?highlight=language_code#faker.providers.BaseProvider.language_code) but don't think our transcript IDs are going to remain language codes for long so I didn't bother

via/models/_mixins.py
```diff
@@ -0,0 +1,20 @@
...
+        server_default=func.now(),  # pylint:disable=not-callable
+        sort_order=-10,
+    )
+    updated: Mapped[datetime] = mapped_column(
+        init=False,
+        repr=False,
+        server_default=func.now(),  # pylint:disable=not-callable
+        onupdate=func.now(),  # pylint:disable=not-callable
+        sort_order=-10,
+    )
```
seanh, line 26: This is based on the [`TimestampMixin` example from the SQLAlchemy docs](https://docs.sqlalchemy.org/en/20/orm/declarative_mixins.html#mixing-in-columns) and also taking some ideas from [our own `CreatedUpdatedMixin` in LMS](https://github.com/hypothesis/lms/blob/e9f4491ffdc07cf586dc7a36303dd1de165aa0cd/lms/models/_mixins/created_updated.py#L4-L11) (using `server_default`, `onupdate` and `func.now()` for the timestamp fields so they get created and updated automatically).

I removed the timestamps from the instance's string representation (`repr=False`) because they're quite noisy when printed out.

I also removed the timestamps from the model's `__init__()` method (`init=False`) because I didn't want users to have to provide `created` and `updated` arguments every time you create a `Transcript`. You can add a `default=func.now()` (rather than `server_default`) so that `created` and `updated` become optional `__init__()` but because these columns are in a mixin class this then results in an error when SQLAlchemy tries to create the Python dataclass:

    sqlalchemy.exc.InvalidRequestError: Python dataclasses error encountered when creating dataclass for 'Transcript': TypeError("non-default argument 'video_id' follows default argument"). Please refer to Python dataclasses documentation for additional information. (Background on this error at: https://sqlalche.me/e/20/dcte)

A downside of the current implementation is that if you actually want to create a `Transcript` and provide your own `created` and `updated` columns you can't pass `created` and `updated` as arguments to `Transcript.__init__()`. Instead you'll have to do this:

```python
transcript = Transcript(...)
transcript.created = ...
transcript.updated = ...
```

I think this is the best compromise. In practice you never do want to set `created` and `updated` yourself anyway

seanh: The `sort_order=-10` is to make sure that the `created` and `updated` columns come first in the SQL schema. Otherwise I think we could have problems with the order of the columns if any new columns are added to `Transcript` in future

via/services/youtube.py
```diff
@@ -82,11 +88,37 @@ def get_transcript(self, video_id):
         :raise Exception: this method might raise any type of exception that
             YouTubeTranscriptApi raises
         """
-        return YouTubeTranscriptApi.get_transcript(video_id)
+        transcript_id = language_code = "en"
```
seanh, line 91: Making it explicit that we're using the hard-coded language code `"en"` as the transcript ID for caching.

Language codes are what the third-party `youtube-transcript-api` library that we're using uses for transcript IDs, so I think we're stuck with this for now.

If/when we replace `youtube-transcript-api` with our own code our code will probably use a different form of transcript ID since the language code alone isn't unique. If/when that happens we'll effectively invalidate existing cached transcripts in the DB as they'll still have the transcript ID `"en"` so lookups won't find them anymore. We've decided to accept this: Via will just re-download each of those transcripts once and store them with the new ID. We'll have duplicate transcripts in the DB but this is unimportant and we could easily delete them if we want.

via/services/youtube.py
```diff
@@ -82,11 +88,37 @@ def get_transcript(self, video_id):
...
+                        Transcript.video_id == video_id,
+                        Transcript.transcript_id == transcript_id,
+                    )
+                )
+                .one()
+                .transcript
+            )
+        except NoResultFound:
+            transcript = YouTubeTranscriptApi.get_transcript(
+                video_id, languages=(language_code,)
```
seanh, line 106: `languages=("en",)` is the default, but I wanted to be explicit here

via/models/transcript.py
```diff
@@ -0,0 +1,16 @@
...
+
+from via.db import Base
+from via.models._mixins import CreatedUpdatedMixin
+
+
+class Transcript(CreatedUpdatedMixin, Base):
+    __tablename__ = "transcript"
+    __table_args__ = (UniqueConstraint("video_id", "transcript_id"),)
+
+    id: Mapped[int] = mapped_column(primary_key=True, autoincrement=True, init=False)
```
seanh: Discussed in Slack: we might want to add `sort_order=-100` to `id` so that it sorts before `created` and `updated`, and we might want to move `id` into a mixin.

seanh: Sent a separate PR to show this change, should we want to make it: https://github.com/hypothesis/via/pull/1138

marcospri (approved):
Looks good, not much to add.

We have to deploy the migration before this, added a `do not merge` tag just in case.

PR #1073, opened 2023-07-11
Add a DB migration to add the `transcript` table

seanh:


marcospri:
Tested with

```docker stop via3-postgres-1; docker rm via3-postgres-1; make services; tox -e dev --run-command 'alembic upgrade head'```

marcospri:
To apply the effort for https://github.com/hypothesis/via/pull/1138

We should modify the migration to reflect that order

seanh:
> To apply the effort for #1138
> 
> We should modify the migration to reflect that order

Done

PR #1074, opened 2023-07-12
Add clear filter to video app search box

### LINKED ISSUE #1050: Add "clear search" icon to transcript search
Add a way to clear the search with an icon in the transcript search

acelaya:
This PR closes #1050 by adding a dynamic button that allows clearing the active search/filter in the video app.

> Take the styling as provisional.

> **Note**
> This PR is easier to review ignoring whitespaces.

[Grabación de pantalla desde 2023-07-12 10-56-48.webm](https://github.com/hypothesis/via/assets/2719332/94863ebf-7369-45a4-a815-c710d3988d99)

### Testing steps:

1. Write some content in the search box. The clear button should appear to the right
2. Remove the content from the search box. The button should disappear.
3. Write content again, and click the button. The content box should be cleared, and the button removed.
4. The button should be part of the focus sequence, after the input, and have a focus ring when focused.
5. Screen readers should announce the button's title when focused.

### DIFF
--- a/via/static/scripts/video_player/components/FilterInput.tsx
+++ b/via/static/scripts/video_player/components/FilterInput.tsx
@@ -1,0 +1,49 @@
+import { CancelIcon, IconButton, Input } from '@hypothesis/frontend-shared';
+import classnames from 'classnames';
+import type { Ref } from 'preact';
+
+export type FilterInputProps = {
+  elementRef: Ref<HTMLElement | undefined>;
+  setFilter: (filter: string) => void;
+  filter: string;
+};
+
+export default function FilterInput({
+  elementRef,
+  setFilter,
+  filter,
+}: FilterInputProps) {
+  return (
+    <div className="relative">
+      <Input
+        data-testid="filter-input"
+        aria-label="Transcript filter"
+        classes={classnames(
+          // Match height of search input in sidebar
+          'h-[32px]',
+          // Extra padding right to prevent text and "clear" button overlapping
+          'pr-8'
+        )}
+        elementRef={elementRef}
+        onKeyUp={e => {
+          // Allow user to easily remove focus from search input.
+          if (e.key === 'Escape') {
+            (e.target as HTMLElement).blur();
+          }
+        }}
+        onInput={e => setFilter((e.target as HTMLInputElement).value)}
+        placeholder="Search..."
+        value={filter}
+      />
+      {filter && (
+        <IconButton
+          icon={CancelIcon}
+          data-testid="clear-filter-button"
+          title="Clear filter"
+          onClick={() => setFilter('')}
+          classes="absolute right-0 top-[1px]"
+        />
+      )}
+    </div>
+  );
+}
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -1,15 +1,12 @@
 import {
   Button,
-  CancelIcon,
   Checkbox,
   CopyIcon,
   IconButton,
-  Input,
   LogoIcon,
   Spinner,
 } from '@hypothesis/frontend-shared';
 import classnames from 'classnames';
-import type { Ref } from 'preact';
 import {
   useCallback,
   useEffect,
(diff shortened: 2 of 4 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -71,24 +72,43 @@ type FilterInputProps = {
...
+          unstyled
+          classes={classnames(
+            'text-grey-7 hover:text-grey-9 font-semibold',
+            'focus-visible-ring',
+            'absolute right-2 top-[6px] w-5 h-5'
+          )}
+        >
+          <CancelIcon className="w-5 h-5" />
+        </Button>
       )}
```
robertknight: I expect we'll want to incorporate some variation of this into the pattern library in future. I would definitely extract a separate component here though. The `VideoPlayerApp` component is becoming decently large, so we need to keep an eye out for opportunities to extract pieces into their own components.

lyzadanger: I agree with Rob that it is probably time to extract this as a component (locally).

acelaya: I guess you guys mean the whole "search input"?

lyzadanger: Yep, what is currently a component local to `VideoPlayerApp`.

acelaya: Gotcha

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
```diff
@@ -69,6 +69,7 @@ describe('VideoPlayerApp', () => {
     // without dedicated tests.
     $imports.$restore({
       './icons': true,
+      './FilterInput': true,
```
acelaya: At this point it was simpler to not mock this component and just extend this test with a couple new cases.

At some point we may want to create a `FilterInput-test` though.

lyzadanger (changes requested):
As discussed together in Slack, let's go ahead and add some `FilterQuery` tests for the extracted component. Thanks!

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
```diff
@@ -506,35 +508,44 @@ describe('VideoPlayerApp', () => {
...
-        .find('input[data-testid="filter-input"]')
-        .getDOMNode();
-      inputField.focus();
+    context('when interacting with input', () => {
+      beforeEach(() => {
+        // These tests need the actual input
+        $imports.$restore({
+          './FilterInput': true,
+        });
+      });
```
acelaya, line 517: I have actually left these two tests here, where we do not mock the `FilterInput`, because I think they are more useful and test something closer to reality like this.

We could try to keep the mock, but then I have the feeling we would end up just testing that some mocks behave as we configured them here.

robertknight: If the test really depends on the combined behavior of both components that that's reasonable. In this case the "clear" button is really an implementation detail of the `FilterInput` - from the point of view of `VideoPlayerApp` whether you click the "clear" button or edit the input, it looks the same. This means that it is a good abstraction! But it also means that the tests for the clear button are better placed in dedicated tests for `FilterInput`.

acelaya: The tests for the clear button are all in `FilterInput-test`.

The tests I was mentioning here, that for me make more sense to keep, are all related with keyboard shortcuts and how they behave combined with the input being focused, not focused, etc.

via/static/scripts/video_player/components/FilterInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+        value={filter}
+      />
+      {filter && (
+        <IconButton
+          icon={CancelIcon}
+          data-testid="clear-filter-button"
+          title="Clear filter"
+          onClick={() => setFilter('')}
+          classes={classnames(
+            'focus-visible-ring',
```
lyzadanger: Shouldn't need this class (you get it for free).

acelaya: Done

lyzadanger (approved):
Looks good! We might tweak the clear-button styling at some point, but it's a solid start.

PR #1077, opened 2023-07-13
Add a Postgres DB to Via

seanh:
This adds all the DB boilerplate but doesn't actually make the DB usable by the code or tests yet. This is just the result of setting `postgres` to `yes` in `cookiecutter.json` and running `make template` (using the cookiecutter's current `main` branch). The only other things I did were:

* Add `__postgres_version` to `cookiecutter.json`
* Remove `docker-compose.yml` from the `ignore` list in `cookiecutter.json` so that the cookiecutter now templates `docker-compose.yml` because we want it do add the Postgres service
* Move Via's nginx service into `.cookiecutter/includes/docker-compose/services.yml` so that `make template` includes it in `docker-compose.yml`
* Add this to Via's nginx service:

      profiles:
      - make_dev

  This is so that the `make services` command (which is newly added by the cookiecutter in this commit) doesn't start nginx: Via's nginx is started by `make dev` (it's in `supervisord-dev.conf`, this is existing behaviour on Via `main`, the reason for this is so that after editing the nginx config files you only need to restart `make dev`, you don't need to restart `make services` in order for your changes to take effect).

After that just ran `make template` and did not make any manual changes in this PR.

## Testing

There's not much to test here since there aren't any models or tables yet. You can try out the various `make` commands. https://github.com/hypothesis/via/pull/1019 and https://github.com/hypothesis/via/pull/1072 actually add some code that you can test.


marcospri (approved):
I can't see anything here that is unfamiliar.

I tested the basic functionality over https://github.com/hypothesis/via/pull/1093#issuecomment-1641701870


We might find some issue while deploying this if we missed anything that affects only the EBS environments. I reckon we can rely on QA to catch those.

PR #1079, opened 2023-07-13
Report canonical URL from video player to top frame

robertknight:
Make sure the video player reports the same document URL to the top frame as the Hypothesis client uses to search for annotations.

This doesn't make any functional difference, since the actual values are the same before/after this PR. It just ensures consistency in the future.

See the PR linked in the removed "TODO" comment for more context.


acelaya (approved):
👍🏼

Issue #999, opened 2023-06-19
Include video title in video player page title


robertknight:
The Hypothesis video player title is just "Hypothesis Video Player". It would be more useful if the video title was included, ie. "{Video title} - Hypothesis Video Player".

robertknight:
It appears there are some [undocumented APIs](https://stackoverflow.com/questions/47282202/youtube-iframe-player-api-getvideodata-is-removed-how-to-get-title) that would allow the frontend to fetch this from the YouTube player. However I think we'd be in a safer position if the backend retrieved this using the same API we use in the LMS's assignment picker.

Edit: I confirmed that the `player.getVideoData()` API (where `player` is a `YT.Player` instance) does exist in the current YouTube embed and does return the title. We _could_ use this, but shouldn't rely on it. Getting the backend to provide the data would still be preferable.

robertknight:
This is now done. The tab title reads "Via: {video_title}", where `{video_title}` is the title of the video from the YouTube API, set via a `<title>` tag in the video player iframe.

PR #1080, opened 2023-07-13
Fix the `<title>` of `/video/youtube` pages

### LINKED ISSUE #1071: Provide video title to video player frontend
To support https://github.com/hypothesis/via/issues/941 and https://github.com/hypothesis/via/issues/999, the backend should provide the YouTube video title to the frontend. For this purpose it should use the same API that the LMS app uses when a YouTube assignment is configured.

The backend could provide the title information to the frontend in any of these ways:

1. Set the `<title>` tag of the HTML
2. A piece of configuration in the HTML
3. An API which the frontend calls to get the title

Approach (1) has the advantage of setting a sensible title if a tool which doesn't run JavaScript fetches the page. Approaches (2) and (3) are more flexible if we want to add in other metadata in future (eg. channel name, description). (3) allows the API call to YouTube to potentially be made concurrently with the client loading and the transcript being fetched.

seanh:
Fixes https://github.com/hypothesis/via/issues/1071. Depends on https://github.com/hypothesis/devdata/pull/92 (or set `YOUTUBE_API_KEY` in your shell).

This will call the YouTube API to get the video title every time someone loads the `/video/youtube` page. Once we've added a DB to Via we should save video titles in the DB and only call the API the first time.

## Testing

Visit a page like <http://localhost:9083/https://www.youtube.com/watch?v=sWNTsKX5H5M> and you should see that the browser's tab title gets set to the title of the YouTube video.

## Error handling

There are various ways that you can break the code to trigger an error from the YouTube API, for example: break the code to use an invalid API key, break the code to use an invalid API endpoint URL, request a YouTube video that doesn't exist (e.g. <http://localhost:9082/video/youtube?url=https://www.youtube.com/watch?v=foo>), break the code to send invalid request params, turn off your internet connection so the YouTube API can't be reached, etc. They all produce pretty much the same result.

For example let's try breaking the API key:

```diff
diff --git a/via/services/youtube.py b/via/services/youtube.py
index 3b3fcb1..30f10ac 100644
--- a/via/services/youtube.py
+++ b/via/services/youtube.py
@@ -81,6 +81,6 @@ class YouTubeService:
 def factory(_context, request):
     return YouTubeService(
         enabled=request.registry.settings["youtube_transcripts"],
-        api_key=request.registry.settings["youtube_api_key"],
+        api_key="foo",
         http_service=request.find_service(HTTPService),
     )
```

This results in a **400 Bad Request** response from Google with a JSON error body. Via turns this into a **417 Expectation Failed**:

image

The error also gets reported to Sentry, including the full details of the error message that we received from the YouTube API (including the full error response body). Here's an example from my local dev env: https://sean-hammond.sentry.io/share/issue/ad40b67f46d443768aa2a0553ecfebb9/

The error also gets logged to Papertrail and you'll see the full error response body from the YouTube API there as well, you can see this in your terminal in dev.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,6 +21,7 @@ def pyramid_settings():
         "dev": False,
         "youtube_transcripts": True,
         "api_jwt_secret": "secret",
+        "youtube_api_key": "test_youtube_api_key",
     }
 
 
--- a/via/app.py
+++ b/via/app.py
@@ -29,6 +29,7 @@ PARAMETERS = {
     "enable_front_page": {"formatter": asbool},
     "youtube_transcripts": {"formatter": asbool},
     "api_jwt_secret": {"required": True},
+    "youtube_api_key": {},
 }
 
 
--- a/via/services/youtube.py
+++ b/via/services/youtube.py
@@ -2,14 +2,22 @@ from urllib.parse import parse_qs, quote_plus, urlparse
 
 from youtube_transcript_api import YouTubeTranscriptApi
 
+from via.services.http import HTTPService
+
+
+class YouTubeDataAPIError(Exception):
+    """A problem with calling the YouTube Data API."""
+
 
 class YouTubeService:
-    def __init__(self, enabled: bool):
+    def __init__(self, enabled: bool, api_key: str, http_service: HTTPService):
         self._enabled = enabled
+        self._api_key = api_key
+        self._http_service = http_service
 
     @property
     def enabled(self):
-        return self._enabled
+        return bool(self._enabled and self._api_key)
 
     def canonical_video_url(self, video_id: str) -> str:
         """
@@ -51,6 +59,22 @@ class YouTubeService:
 
         return None
 
+    def get_video_title(self, video_id):
+        """Call the YouTube API and return the title for the given video_id."""
+        # https://developers.google.com/youtube/v3/docs/videos/list
+        try:
+            return self._http_service.get(
+                "https://www.googleapis.com/youtube/v3/videos",
+                params={
+                    "id": video_id,
+                    "key": self._api_key,
+                    "part": "snippet",
+                    "maxResults": "1",
+                },
+            ).json()["items"][0]["snippet"]["title"]
+        except Exception as exc:
+            raise YouTubeDataAPIError("getting the video title failed") from exc
+
     def get_transcript(self, video_id):
         """
         Call the YouTube API and return the transcript for the given video_id.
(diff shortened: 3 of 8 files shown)


tests/conftest.py
```diff
@@ -21,6 +21,7 @@ def pyramid_settings():
         "dev": False,
         "youtube_transcripts": True,
         "api_jwt_secret": "secret",
+        "youtube_api_key": "test_youtube_api_key",
```
seanh, line 24: I'm not a huge fan of this great big centralised settings fixture to be honest. `"youtube_api_key"` is only read by `YouTubeService` so it should only need to be set in `YouTubeService`'s tests. But this is the pattern used so far, might send a separate PR to change it

via/app.py
```diff
@@ -29,6 +29,7 @@
     "enable_front_page": {"formatter": asbool},
     "youtube_transcripts": {"formatter": asbool},
     "api_jwt_secret": {"required": True},
+    "youtube_api_key": {"required": True},
```
seanh: Making this envvar required. I don't think we want to have to worry about two different behaviours/configurations of Via: with and without a YouTube API key. We can just set the envvar in every deployment that we have

via/services/youtube.py
```diff
@@ -51,6 +55,19 @@ def get_video_id(self, url):
...
+        # https://developers.google.com/youtube/v3/docs/videos/list
+        return self._http_service.get(
+            "https://www.googleapis.com/youtube/v3/videos",
+            params={
+                "id": video_id,
+                "key": self._api_key,
+                "part": "snippet",
+                "maxResults": "1",
+            },
+        ).json()["items"][0]["snippet"]["title"]
```
seanh: :man_shrugging: Pretty simple, no point in complicating things

via/app.py
robertknight: If we are going to add a lot of integrations to Via over time then I am concerned about stacking up the number of API keys etc. that have to be configured before you can run it. I think it would be fine to make this API key a prerequisite of using the video player app though, as you can still use it for proxying PDFs etc. without that.

robertknight (commented):
Functionally this works as desired from the frontend / client's perspective. When this lands we can also close https://github.com/hypothesis/via/issues/999 and https://github.com/hypothesis/via/issues/941 (there is a part in #941 about capturing timestamps, but I think that can be extracted into a separate issue).

I would prefer to make YouTube API keys optional to run Via, but required to use the video player app.

marcospri (approved):
Works as expected :+1: 

My only concern/thing to be aware of is that the although it's a small cost the /list endpoint does counts towards the quota usage of the official API.

https://developers.google.com/youtube/v3/docs/videos/list


https://developers.google.com/youtube/v3/getting-started#quota

10.000 video views in the LMS context **per day** should be plenty in the LMS context.

PR #1081, opened 2023-07-13
Encapsulate layout logic in `useAppLayout` hook; update breakpoints

lyzadanger:
This PR:

* Encapsulates computed layout values within the `useAppLayout` hook and removes computation from `VideoPlayerApp`. Instead of returning a single `appSize` string, `useAppLayout` now returns a layout info object. This object should only be updated if the relative container size changes.
* Reshuffles breakpoints and removes `2xl` size.

There should be very minimal visible changes (very small adjustments to padding around the video in some widths).

Depends on #1063

Fixes https://github.com/orgs/hypothesis/projects/118/views/1?pane=issue&itemId=33204585

### DIFF
--- a/via/static/scripts/video_player/hooks/use-app-layout.ts
+++ b/via/static/scripts/video_player/hooks/use-app-layout.ts
@@ -1,24 +1,41 @@
 import type { RefObject } from 'preact';
-import { useCallback, useLayoutEffect, useState } from 'preact/hooks';
+import { useCallback, useLayoutEffect, useRef, useState } from 'preact/hooks';
 
 import { SIDEBYSIDE_THRESHOLD, TOOLBAR_WIDTH } from './use-side-by-side-layout';
 
 /* Relative application container size */
 export type AppSize = 'sm' | 'md' | 'lg' | 'xl' | '2xl';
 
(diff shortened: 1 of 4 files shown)


via/static/scripts/video_player/hooks/use-app-layout.ts
```diff
@@ -1,24 +1,41 @@
...
 export function useAppLayout(
   appContainer: RefObject<HTMLDivElement | null>
-): AppSize {
-  const [appSize, setAppSize] = useState<AppSize>('sm');
+): AppLayoutInfo {
+  const lastContainerSize = useRef<AppSize>('sm');
+  const [layoutInfo, setLayoutInfo] = useState<AppLayoutInfo>({
+    appSize: 'sm',
+    multicolumn: false,
+    transcriptWidth: transcriptWidths.sm,
```
robertknight, line 37: `multicolumn` and `transcriptWidth` can be computed from the `appSize` value, right? In that case you could also store just the `appSize` in state and compute the other properties from that.

robertknight (approved):
LGTM, though I left a note about potentially storing only the `appSize` in state if the other properties can be computed from it.

Issue #1062, opened 2023-07-10
Spike: how can we select the transcript for a YouTube video?


seanh:
A YouTube video can have multiple transcripts. For example there can be transcripts in different languages such as **"English"**, **"English (US)"**, **"Spanish"**, etc. There can also be multiple transcripts for the same language, e.g. **"English"**, **"English - DTVCC1"**, **"English - CC1"**, etc. There can also be manually-created and machine-generated transcripts, e.g. **"English"** and **"English (auto-generated)"**.

The YouTube UI lists all the available transcripts and lets the user select exactly which transcript they want, including selecting between multiple transcripts for the same language:

![image](https://github.com/hypothesis/via/assets/22498/8374ebc4-7bbc-4cd9-bd56-3f96458ebfbe)

![image](https://github.com/hypothesis/via/assets/22498/37238479-98de-4dae-83b8-eafaadabd4a3)

[Ideally](https://hypothes-is.slack.com/archives/C4K6M7P5E/p1689000118683689?thread_ts=1688993296.282669&cid=C4K6M7P5E) we'd like Via (and LMS) to let the user select between the same list of transcripts that YouTube's own UI does, including letting the user select the exact transcript that they want even when there are multiple transcripts for the same language or manually-created and auto-generated transcripts for the same language.

How can we let the user select the transcript to use, given that we're using https://github.com/jdepoix/youtube-transcript-api to get the transcripts (and we probably don't want to replace that library right now as that would take some time)?

## Some examples

 * Only has an auto-generated English transcript: https://www.youtube.com/watch?v=EU6TDnV5osM
 * Has transcripts in multiple languages, including both manually-created and machine generated English ones: https://www.youtube.com/watch?v=qQ6a0iOzyHE
 * Has no transcript at all: https://www.youtube.com/watch?v=rw_scW4D3ik
 * Has four different English transcripts: https://www.youtube.com/watch?v=rSTqpRYzJbo

jon-betts:
I think this pretty easy. It's well supported by the API exposed by the library. Here is some example code:

```python
from youtube_transcript_api import YouTubeTranscriptApi


def list_transcripts(video_id):
    for transcript in YouTubeTranscriptApi.list_transcripts(video_id):
        # This is just to demonstrate the interface
        yield {
            "name": transcript.language,
            "language_code": transcript.language_code,
            "is_autogenerated": transcript.is_generated,
            "translatable_to": transcript.translation_languages
        }


[...]

PR #1082, opened 2023-07-13
First party YouTube API client - with video info

jon-betts:
For:

 * https://github.com/hypothesis/via/issues/1062

This is a modification of the approach taken by https://pypi.org/project/youtube-transcript-api/

The basic structure of that approach is still here:

 * Get the HTML page for the video (dealing with cookie popups)
 * Parse parts of the video config JSON
 * Marshal parts of the parsed JSON into structured info about the video centered on transcripts
 * Check other parts of the video config JSON **_as raw HTML strings_** to detect situations and errors
 * Pick a transcript from those listed
 * Use the URL provided by YouTube to get the XML transcript

How this approach differs:

 * We parse the whole JSON config instead of just parts
 * We marshal more information into structured info
 * We use that structured info to detect more error situations (which should be less error prone than string searching)
 * We make no assumptions about the uniqueness of language codes
 * We use a discovered `vssId` for a transcript instead

Side effects:

 * While we are marshalling the transcript info, we can also gather lots of video info
 * This include things like name, thumbnails, description etc.

## This is a demo! We don't want to use it like this!

At the time of writing this is integrated into the view which loads up a YouTube video, which means we are doing the look up twice:

 * Once during the initial page load
 * Once during the transcript lookup

As it's currently written, even with caching for transcripts we'd continue to hit YouTube all the time. 

**We don't want this. I'm not proposing this. This is just a demo showing we can get all of the info.**

As / when we get an official API we could replace the first lookup however:

* It's important to realise that parsing this info from scratch is how this approach works
* The transcript id is not enough, we need a short lived URL from YouTube to get it
* The only extra parsing that's done here (for show) is the video details part
* Even if we get an official thing which replicates a lot of this, we'll still need to do almost all of this anyway


via/services/youtube_api/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
jon-betts, line 21: I don't know at what point we've transformed this out of being a derivative work?

via/services/youtube_api/client.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def canonical_video_url(self, video_id: str) -> str:
+        """
+        Return the canonical URL for a YouTube video.
+
+        This is used as the URL which YouTube transcript annotations are
+        associated with.
+        """
+        escaped_id = quote_plus(video_id)
+        return f"https://www.youtube.com/watch?v={escaped_id}"
```
jon-betts, line 37: Robbed from the other service. Maybe the YT parsing thing should be here too?

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @classmethod
+    def from_json(cls, data):
+        return VideoDetails(
+            id=data["videoId"],
+            title=data["title"],
+            short_description=data["shortDescription"],
+            author=data["author"],
+            thumbnails=data["thumbnail"]["thumbnails"],
+        )
```
jon-betts, line 31: This part is the only part which isn't strictly necessary for transcripts. I put it here to show we can grab this at the same time easily.

We could use this type of info to give a very rich interface in the file picker.

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,118 @@
...
+    @classmethod
+    def from_json(cls, data):
+        return Captions(
+            tracks=[
+                CaptionTrack.from_json(track) for track in data.get("captionTracks", [])
+            ],
+            translation_languages=[
+                {"code": language["languageCode"], "name": language["languageName"]}
+                for language in data.get("translationLanguages", [])
+            ],
```
jon-betts, line 170: We're not using it, but if:

 * A transcript `is_translatable`
 * And you have languages here
 * Then you can request a transcript be translated by appending `&tlang={language_code}` on the end

This list of languages does not belong to any specific caption track.

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,118 @@
+from dataclasses import dataclass
+from typing import List, Optional
+
+
+@dataclass
+class VideoDetails:
```
jon-betts, line 12: All of the names of these things are chosen for minimum mapping from the YouTube JSON names. I followed the format very closely.

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,118 @@
...
+    id: str
+    title: str
+    short_description: str
+    author: str
+    thumbnails: List[dict]
```
jon-betts: There are other things in here which might be of interest to us. For example:

 * `isPrivate` - Might this be an indicator a user has messed up? We might want to catch that
 * `isLiveContent` - Doesn't seem annotatable?

via/services/youtube_api/LICENSE.txt
jon-betts: Once we aren't scraping HTML we have to be pretty close...

via/services/youtube_api/models.py
seanh: FYI we already have code for fetching these kinds of video details (name, thumbnail) from the offical V3 API, the file picker in LMS already does this and presents a rich picker interface. But that's in LMS, it's separate from the code in Via that needs to get the transcripts themselves.

Via does also need to get the video title as well as the transcript but that's quite easily done from the v3 API: https://github.com/hypothesis/via/blob/a04feaa64ec2b4530d0c89191b06373f6b3a44c2/via/services/youtube.py#L58-L69

seanh:
This is no longer needed: it's been replaced by a series of non-WIP PRs

PR #1083, opened 2023-07-14
Side by side active

### LINKED ISSUE #1084: Update side-by-side handling to provide `isActive` callback in client config
Logic introduced in https://github.com/hypothesis/client/pull/5623 allows the video app to fully handle side-by-side.

However, the client needs to know if it's enabled or not, for things like automatically closing the sidebar when overlapping content.

Extend the logic to make use of https://github.com/hypothesis/client/pull/5623 and notify the client when side-by-side is enabled.

acelaya:
https://github.com/hypothesis/via/pull/1063 introduced logic to manually handle side-by-side directly from the video app. This PR extends that logic to let the client know when `side-by-side` is enabled by the app, ensuring the sidebar is not automatically closed if not needed.

### Testing steps

1. Open http://localhost:9083/https://www.youtube.com/watch?v=MrORaCfcKxQ on a resolution big enough so that the sidebar can coexist with the rest of the content.
2. Clicking anywhere outside the sidebar should not close it.
3. Change resolution to something smaller, where the sidebar overlaps with the content when it is open.
4. Clicking anywhere outside the sidebar should close it in this case.

In `main`, the sidebar always gets closed when clicking outside.

> This PR closes https://github.com/hypothesis/via/issues/1084

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -111,7 +111,7 @@ export default function VideoPlayerApp({
 
   const { appSize, multicolumn, transcriptWidth } =
     useAppLayout(appContainerRef);
-  useSideBySideLayout();
+  const sideBySideActive = useSideBySideLayout();
 
   // Fetch transcript when app loads.
   const [transcript, setTranscript] = useState<
@@ -229,12 +229,17 @@ export default function VideoPlayerApp({
   };
 
   const bucketContainerId = 'bucket-container';
+  const prevSideBySideActive = useRef(sideBySideActive);
+  prevSideBySideActive.current = sideBySideActive;
   const clientConfig = useMemo(() => {
     return {
       ...baseClientConfig,
       bucketContainerSelector: '#' + bucketContainerId,
       sideBySide: {
         mode: 'manual',
+        // Using a ref here to make sure the `isActive` callback reference is
+        // kept, and only its return value changes
+        isActive: () => prevSideBySideActive.current,
       },
     };
   }, [baseClientConfig]);
--- a/via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
+++ b/via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
@@ -53,6 +53,7 @@ describe('VideoPlayerApp', () => {
   let wrappers;
   let fakeLayoutInfo;
   let fakeUseAppLayout;
+  let fakeUseSideBySideLayout;
 
   function createVideoPlayer(props = {}) {
     const wrapper = mount(
@@ -78,6 +79,7 @@ describe('VideoPlayerApp', () => {
       transcriptWidth: '100%',
     };
     fakeUseAppLayout = sinon.stub().returns(fakeLayoutInfo);
+    fakeUseSideBySideLayout = sinon.stub().returns(false);
     fakeCallAPI = sinon.stub().rejects(new Error('Dummy API error'));
 
     $imports.$mock(mockImportedComponents());
@@ -91,6 +93,9 @@ describe('VideoPlayerApp', () => {
     $imports.$mock({
       './Transcript': FakeTranscript,
       '../hooks/use-app-layout': { useAppLayout: fakeUseAppLayout },
+      '../hooks/use-side-by-side-layout': {
+        useSideBySideLayout: fakeUseSideBySideLayout,
+      },
       '../utils/api': {
         callAPI: fakeCallAPI,
       },
@@ -463,12 +468,21 @@ describe('VideoPlayerApp', () => {
     const clientConfig = { openSidebar: true };
     const wrapper = createVideoPlayer({ clientConfig });
     const mergedConfig = wrapper.find('HypothesisClient').prop('config');
-    assert.deepEqual(mergedConfig, {
-      openSidebar: true,
-      bucketContainerSelector: '#bucket-container',
-      sideBySide: {
-        mode: 'manual',
-      },
+
+    assert.isTrue(mergedConfig.openSidebar);
+    assert.equal(mergedConfig.bucketContainerSelector, '#bucket-container');
+    assert.equal(mergedConfig.sideBySide.mode, 'manual');
+    assert.isDefined(mergedConfig.sideBySide.isActive);
+  });
+
+  [true, false].forEach(sideBySideEnabled => {
+    it('tells client if side-by-side is active', () => {
+      fakeUseSideBySideLayout.returns(sideBySideEnabled);
+
+      const wrapper = createVideoPlayer();
+      const { sideBySide } = wrapper.find('HypothesisClient').prop('config');
+
+      assert.equal(sideBySide.isActive(), sideBySideEnabled);
     });
   });
 
--- a/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
+++ b/via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
@@ -11,10 +11,11 @@ describe('useSideBySideLayout', () => {
 
   // Create a fake component to mount in tests that uses the hook
   function FakeComponent() {
-    useSideBySideLayout(container);
+    const sideBySideEnabled = useSideBySideLayout(container);
     return (
       <div style="width:100%">
         <button>Hi</button>
+        {sideBySideEnabled && <span data-testid="side-by-side-enabled" />}
       </div>
     );
   }
@@ -62,14 +63,9 @@ describe('useSideBySideLayout', () => {
   }
 
   function sideBySideState() {
-    const attr = container.getAttribute('data-side-by-side');
-    if (attr === 'true') {
-      return true;
-    } else if (attr === 'false') {
-      return false;
-    } else {
-      return undefined;
-    }
+    return (
+      container.querySelector('[data-testid="side-by-side-enabled"]') !== null
+    );
   }
 
   describe('initial render', () => {
--- a/via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
+++ b/via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
@@ -38,13 +38,15 @@ export const TOOLBAR_WIDTH = 22;
  * is resized.
  *
  * @param [container] test seam
+ * @return Whether side-by-side is currently active or not
  */
-export function useSideBySideLayout(container = document.body) {
+export function useSideBySideLayout(container = document.body): boolean {
   const [sidebarOpen, setSidebarOpen] = useState(false);
   const [sufficientSpace, setSufficientSpace] = useState(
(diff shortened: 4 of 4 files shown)


via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
```diff
@@ -62,14 +63,9 @@ describe('useSideBySideLayout', () => {
...
-    if (attr === 'true') {
-      return true;
-    } else if (attr === 'false') {
-      return false;
-    } else {
-      return undefined;
-    }
+    return (
+      container.querySelector('[data-testid="side-by-side-enabled"]') !== null
+    );
```
acelaya, line 68: This helps test the value which is now returned by the hook.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -235,8 +236,14 @@ export default function VideoPlayerApp({
...
         mode: 'manual',
+        // Using a ref here to make sure the `isActive` callback reference is
+        // kept, and only its return value changes
+        isActive: () => sideBySideActiveRef.current,
       },
     };
+    // The sideBySideActiveRef is a ref, but since it is returned by a custom
+    // hook, ESLint doesn't seem to properly detect it's a stable value
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [baseClientConfig]);
```
robertknight, line 245: Disabling `exhaustive-deps` should be avoided if possible because it will end up ignoring any future missing dependencies you add, as well as the current one. If you have a stable value from a custom hook, you can just pass it as a dependency. What I would prefer to do here however is to separate out obtaining the current side-by-side state, which may _also_ be used in rendering or effects that need to re-run (though not currently), from storing the value in a ref to bypass this effect:

```diff
diff --git a/via/static/scripts/video_player/components/VideoPlayerApp.tsx b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
index d53d8ec..8f76579 100644
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -112,7 +112,7 @@ export default function VideoPlayerApp({
 
   const { appSize, multicolumn, transcriptWidth } =
     useAppLayout(appContainerRef);
-  const sideBySideActiveRef = useSelfUpdatedRef(useSideBySideLayout());
+  const sideBySideActive = useSideBySideLayout();
 
   // Fetch transcript when app loads.
   const [transcript, setTranscript] = useState<
@@ -230,6 +230,8 @@ export default function VideoPlayerApp({
   };
 
   const bucketContainerId = 'bucket-container';
+  const prevSideBySideActive = useRef(sideBySideActive);
+  prevSideBySideActive.current = sideBySideActive;
   const clientConfig = useMemo(() => {
     return {
       ...baseClientConfig,
@@ -238,12 +240,9 @@ export default function VideoPlayerApp({
         mode: 'manual',
         // Using a ref here to make sure the `isActive` callback reference is
         // kept, and only its return value changes
-        isActive: () => sideBySideActiveRef.current,
+        isActive: () => prevSideBySideActive.current,
       },
     };
-    // The sideBySideActiveRef is a ref, but since it is returned by a custom
-    // hook, ESLint doesn't seem to properly detect it's a stable value
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [baseClientConfig]);
 
   return (
```

via/static/scripts/video_player/components/HypothesisClient.tsx
```diff
@@ -23,11 +25,19 @@ export default function HypothesisClient({
...
+  // `hypothesisConfig` callback is set only once, retaining the reference, and
+  // only its return value changes
+  const configRef = useSelfUpdatedRef(config);
+
   // nb. Use a layout effect here so that variable is definitely set before
   // client's boot script executes.
   useLayoutEffect(() => {
-    (window as any).hypothesisConfig = () => config;
-  }, [config]);
+    (window as any).hypothesisConfig = () => configRef.current;
```
robertknight: The video player app can only rely on `hypothesisConfig` being called once, after the initial render. The client then parses the results and no longer calls it again after that. For that reason I'd like to avoid the changes in this component. Maintaining the identity of `window.hypothesisConfig` isn't important, what is important is that it is set correctly on the first render and any callbacks in the config object return current values, like you've done with `isActive` here.

via/static/scripts/video_player/components/HypothesisClient.tsx
```diff
@@ -23,11 +25,19 @@ export default function HypothesisClient({
...
+    // The configRef is a ref, but since it is returned by a custom hook, ESLint
+    // doesn't seem to properly detect it's a stable value
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
robertknight: I don't think we need the changes in this component, but if we did need something similar in future, the right approach is just to pass the result from the custom hook as a dependency. If that result happens to be stable, it will just never cause the effect/memo etc. to be invalidated.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
robertknight: Related to this, at one point React were going to introduce a [`useEvent` hook](https://blog.logrocket.com/what-you-need-know-react-useevent-hook-rfc/) which would basically be a callback with a stable reference.

```js
// Here `doSomething` never changes, but always invokes the most recent `someCallback` when called
const doSomething = useEvent(someCallback);
```

I think plans changed since, we have a number of use cases for something like this in the video player app, and we could use it for `isActive` here (`const isActive = useEvent(() => sideBySideActive)`). However, I think that abstraction is best left for a separate PR.

robertknight (changes requested):
Functionally, this works as expected.

I do think we have several use cases now in the video player app for a systematic approach to getting stable callbacks which can be used inside effects / memo-d functions etc. without invalidating them. However, I'd prefer not to introduce that here since it needs a bit of care in the documentation to make it clear when / when not to use it. I suggested an alternative which matches a pattern that the `YouTubeVideoPlayer` component already uses.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
acelaya: > However, I think that abstraction is best left for a separate PR.

Fair enough.

via/static/scripts/video_player/components/HypothesisClient.tsx
acelaya: Now that you mention it, it's possible the change for `isActive` is enough.

I first thought this was the culprit for this value not getting updated, just to find out changing this did not make a difference and had to apply a similar logic to `isActive`.

However, with the Monday-morning eyes, what you say now makes a lot of sense. This callback's reference should not affect the reference to `isActive`, even if it later changes for whatever reason.

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
robertknight: This approach is good. The `useSideBySideLayout` hook still sets a `data-side-by-side` attribute on the container which I think is now redundant?

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
```diff
@@ -463,12 +468,21 @@ describe('VideoPlayerApp', () => {
...
+  });
+
+  [true, false].forEach(sideBySideEnabled => {
+    it('tells client if side-by-side is active', () => {
+      fakeUseSideBySideLayout.returns(sideBySideEnabled);
+
+      const wrapper = createVideoPlayer();
+      const { sideBySide } = wrapper.find('HypothesisClient').prop('config');
+
+      assert.equal(sideBySideEnabled, sideBySide.isActive());
```
robertknight: The syntax is `assert.equal(actual, expected)`, so the order of arguments should be swapped here, otherwise you'll get confusing error messages if the test fails.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -39,12 +39,16 @@ export const TOOLBAR_WIDTH = 22;
  *
  * @param [container] test seam
```
robertknight, line 40: Can you add a JSDoc comment here about what the return value means.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -39,12 +39,16 @@ export const TOOLBAR_WIDTH = 22;
...
+export function useSideBySideLayout(container = document.body): boolean {
   const [sidebarOpen, setSidebarOpen] = useState(false);
   const [sufficientSpace, setSufficientSpace] = useState(
     container.clientWidth >= SIDEBYSIDE_THRESHOLD
   );
   const [sidebarWidth, setSidebarWidth] = useState(0);
+  const sideBySideActive = useMemo(
+    () => sufficientSpace && sidebarOpen,
+    [sufficientSpace, sidebarOpen]
+  );
```
robertknight: Since this function is really simple and just returns a boolean, so it would be fine not to memo it here.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
```diff
@@ -60,17 +64,16 @@ export function useSideBySideLayout(container = document.body) {
...
       // prettier-ignore
       container.style.paddingRight = `${
         sidebarWidth - (TOOLBAR_WIDTH / 2)
       }px`;
     } else {
       container.style.paddingRight = `${TOOLBAR_WIDTH}px`;
     }
-    container.setAttribute('data-side-by-side', `${applySideBySide}`);
-  }, [container, sidebarOpen, sidebarWidth, sufficientSpace]);
+    container.setAttribute('data-side-by-side', `${sideBySideActive}`);
```
robertknight: I think this attribute may now be redundant?

robertknight (approved):
LGTM with some minor comments.

via/static/scripts/video_player/components/test/VideoPlayerApp-test.js
acelaya: If only all testing frameworks got this standardized 😂 

Good catch.

via/static/scripts/video_player/hooks/test/use-side-by-side-layout-test.js
acelaya: I considered removing it, but then I thought it might be useful for debugging. But I'm open to remove it, don't have a strong opinion.

via/static/scripts/video_player/hooks/use-side-by-side-layout.ts
acelaya: I left it in case it could be useful for debugging, but probably redundant in most of the cases. I'll remove it and we can always get it back if needed.

PR #1089, opened 2023-07-17
Add logic to handle toast messages (useReducer)

### LINKED ISSUE #1132: Add a toast message component for use in video-annotations application

acelaya:
> **Note**
> This PR is easier to review by ignoring whitespaces.

This PR introduces a proof of concept on how we could handle toast messages for the video app, in a way that we can reuse part of the logic in other apps and have a consistent component stack and UI.

This PR introduces three main pieces.

* A controlled `ToastMessages` component: it is basically the same as `BaseToastMessages` from client, but slightly simplified (the `message` prop is not a string, but `ComponentChildren`, which allows to remove `moreInfoURL` and let consumers determine the full content of a toast message and reduce coupling).
* A context and a hook that can be used to expose the toast messages to components across the app, with mechanisms to append and dismiss messages.
  Any component can do something in these lines:
  ```tsx
  const { toastMessages, appendToastMessage, dismissToastMessage } = useToastMessages();

  // toastMessages is the list of ToastMessage objects.
  // appendToastMessage({ message: 'foo', type: 'error' })
  // dismissToastMessage('abc123')
  ```
* A `ToastMessagesProvider` wrapper, that can be used to expose the context down the component tree, and also handles the toast messages reducer.

In order to verify the logic works and makes sense, it also enhances the copy-to-clipboard button, so that it displays a toast message after content has been copied or an error has occurred.

### TODO / To think

- [ ] Consider extracting `ToastMessages` to frontend-shared, as it can be potentially used in client as well, with no changes.
- [ ] Think if there's a better way to define the toast message handling logic. In client we have the `ToastMessengerService`. Here the logic has been delegated to a simple `useReducer` and exposed via context, but that means only other components can append or dismiss toast messages at the moment.
- [ ] Fix toast message transitions (same problem as in client).
- [ ] Fix toast messages styling and positioning.
- [ ] Think if `ToastMessagesProvider` should wrap `VideoPlayerApp` instead of being "in" it.
  Right now it is in it, as it seems reasonable that, for proper styling, the `ToastMessages` container is inside the app container, to benefit from multicolumn logic.
  However, this means the `VideoPlayerApp` itself cannot use the toast messages context.
  For now the simple workaround is to "extract" `CopyToClipboard` to its own component, as it makes sense anyway.

### DIFF
--- a/via/static/scripts/video_player/components/ToastMessages.tsx
+++ b/via/static/scripts/video_player/components/ToastMessages.tsx
@@ -1,0 +1,89 @@
+import { Callout } from '@hypothesis/frontend-shared';
+import classnames from 'classnames';
+import type { ComponentChildren } from 'preact';
+
+export type ToastMessage = {
+  id: string;
+  type: 'error' | 'success' | 'notice';
+  message: ComponentChildren;
+  visuallyHidden?: boolean;
+  isDismissed?: boolean;
+};
+
+type ToastMessageItemProps = {
+  message: ToastMessage;
+  onDismiss: (id: string) => void;
+};
+
+/**
+ * An individual toast message: a brief and transient success or error message.
+ * The message may be dismissed by clicking on it. `visuallyHidden` toast
+ * messages will not be visible but are still available to screen readers.
+ */
+function ToastMessageItem({ message, onDismiss }: ToastMessageItemProps) {
+  // Capitalize the message type for prepending; Don't prepend a message
+  // type for "notice" messages
+  const prefix =
+    message.type !== 'notice'
+      ? `${message.type.charAt(0).toUpperCase() + message.type.slice(1)}: `
+      : '';
+
+  return (
+    <Callout
+      classes={classnames({
+        'sr-only': message.visuallyHidden,
+      })}
+      status={message.type}
+      onClick={() => onDismiss(message.id)}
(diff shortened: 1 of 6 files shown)


via/static/scripts/video_player/components/ToastMessages.tsx
acelaya, line 1: This is the same as [`BaseToastMessage`](https://github.com/hypothesis/client/blob/57fa3b296f151e1bcc04a1a78ceb1ba6e28819c6/src/shared/components/BaseToastMessages.tsx) from client, with the difference that `message.message` is a `ComponentChildren` instead of a string, and `message.moreInfoURL` was removed.

via/static/scripts/video_player/components/ToastMessagesProvider.tsx
acelaya, line 1: This component is meant to be a wrapper of other components that should have access to toast messages via context.

It also renders the `ToastMessages` container transparently.

via/static/scripts/video_player/utils/generate-hex-string.ts
acelaya, line 1: > Copied from client

acelaya:
Closing in favor of https://github.com/hypothesis/via/pull/1095

PR #1090, opened 2023-07-17
Scroll to current segment after clearing filter

### LINKED ISSUE #1048: Clicking timestamp when filter is active should scroll to that segment, even if auto-scroll is disabled
**Steps to reproduce:**

1. Load a YouTube video in Via
2. Turn off auto-scroll
3. Filter the transcript by typing in the filter bar
4. Click on a timestamp for one of the matches

**Expected result:**

Filter is cleared and transcript scrolls to the match.

**Actual result:**

Filter is cleared but transcript does not scroll to that match.

robertknight:
When the user clicks a segment while a filter is active, the filter should be cleared, the segment made current and the transcript scrolled to the location of that segment in the un-filtered view.

Previously this worked if auto-scroll was enabled, where the transcript was scrolled as a by-product of the current timestamp changing, but not if auto-scroll was disabled. This fixes that case.

Fixes https://github.com/hypothesis/via/issues/1048

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -138,11 +138,14 @@ export default function VideoPlayerApp({
   // Listen for the event the Hypothesis client dispatches before it scrolls
   // a highlight into view.
   const nextRender = useNextRender();
+  const pendingScrollToHighlight = useRef(false);
   useEffect(() => {
     const listener = (e: Event) => {
       if (!isScrollToRangeEvent(e)) {
         return;
       }
+
+      pendingScrollToHighlight.current = true;
 
       // If a filter is currently active, clear it first to ensure the highlight
       // is visible.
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -169,10 +172,26 @@ export default function VideoPlayerApp({
   }, [nextRender]);
 
   const syncTranscript = useCallback(
-    () => transcriptControls.current!.scrollToCurrentSegment(),
+    () => transcriptControls.current?.scrollToCurrentSegment(),
```
robertknight, line 175: This avoids an error if the user manages to set and then clear a filter before the transcript is loaded. We could try to disable the filter input if the transcript has not yet loaded, but given that there are a number of triggers that clear the filter, I think it is going to be more reliable to check when this function is called.

acelaya (approved):
LGTM

PR #1091, opened 2023-07-17
Load Hypothesis client concurrently with transcript

### LINKED ISSUE #1088: Load Hypothesis client concurrently with transcript
The video player currently waits for the transcript to load, before loading the Hypothesis client. It does this to ensure that the content is ready by the time that the client tries to anchor annotations. This is inefficient however and adds a delay of at least hundreds of milliseconds to the time taken to load the client and display annotations.

Instead it would provide a better user experience if the client was loaded as soon as possible, but deferred anchoring annotations until the content was loaded. This will require some client support. See https://github.com/hypothesis/client/issues/5568.

robertknight:
~~**Depends on https://github.com/hypothesis/client/pull/5636**~~

Pass a `contentReady` promise [1] to the client to signal when the transcript is loaded. This enables the client to be loaded concurrently with the transcript, while still making sure that the client waits for the transcript to be rendered before it anchors annotations.

Additionally if the transcript fails to load, the user will still be able to see and interact with any annotations that were already created on the transcript.

[1] https://github.com/hypothesis/client/pull/5636

Fixes https://github.com/hypothesis/via/issues/1088

**Testing:**

1. Check out this branch and https://github.com/hypothesis/client/pull/5636.
2. Load a video. You should see that the client starts to load earlier, compared to `main`. Depending on how fast the transcript loads, you may see the loading spinner at the same time as parts of the client UI.
3. Edit `get_transcript` in `via/views/api/youtube.py` to raise an exception, then reload the page. You will see an error message in the page and the browser console, but annotations will still show up in the sidebar.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -118,6 +118,24 @@ export default function VideoPlayerApp({
     TranscriptData | APIError | null
   >(isTranscript(transcriptSource) ? transcriptSource : null);
 
+  // Notify client once the transcript is rendered, and ready to anchor
+  // annotations.
+  const [contentReady, resolveContentReady, rejectContentReady] =
+    useMemo(() => {
+      let reject: (err: Error) => void;
+      let resolve: () => void;
+      const promise = new Promise<void>((resolve_, reject_) => {
+        resolve = resolve_;
+        reject = reject_;
+      });
+      return [promise, resolve!, reject!];
+    }, []);
+  if (isTranscript(transcript)) {
+    resolveContentReady();
+  } else if (transcript instanceof Error) {
+    rejectContentReady(new Error('Transcript failed to load'));
+  }
+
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -118,6 +118,24 @@ export default function VideoPlayerApp({
...
+  // annotations.
+  const [contentReady, resolveContentReady, rejectContentReady] =
+    useMemo(() => {
+      let reject: (err: Error) => void;
+      let resolve: () => void;
+      const promise = new Promise<void>((resolve_, reject_) => {
+        resolve = resolve_;
+        reject = reject_;
+      });
+      return [promise, resolve!, reject!];
```
robertknight, line 131: In future, dances like this can be simplified using [Promise.withResolvers](https://github.com/tc39/proposal-promise-with-resolvers).

acelaya (approved):
LGTM, and works as expected.

PR #1093, opened 2023-07-17
Cache YouTube video titles in the DB

### LINKED ISSUE #1118: Save video titles in the DB
**Problem:** Via currently [calls the YouTube API to get the video title](https://github.com/hypothesis/via/blob/a3179570411b9c83ca2329f37e11e816d9958186/via/services/youtube.py#L62-L76) every time someone loads Via's YouTube video player page. This consumes are YouTube API key's quota faster than necessary (although I think the quota is very high so this may not be a an issue); it makes page loads slower than necessary (because every page load must wait for a network request to the YouTube API); and it makes page loads more unreliable than necessary (whenever we make a network request to a third-party service there's a chance this request can fail). It also means that if our code for getting video titles from the YouTube API breaks (say YouTube change or remove their API) then all existing video assignments will be broken as each assignment load will crash when trying to get the video title.

**Solution:** Via's `get_video_title()` method should save the video title in Via's DB after retrieving it from the YouTube API. The next time `get_video_title()` is called, if the title for this particular video is already in the DB, it should return the title from the DB and not make a YouTube API request.

### Known Issues

* This will mean that if the title of a YouTube video changes our video player will continue showing the old title from our DB. We could fix this by periodically invalidating the titles in our DB (saying after 24hrs) and re-fetching them but this probably isn't necessary or worthwhile so it's **out of scope** for this issue: at least for the first version of this we'll just continue using the same title forever.

seanh:
Fixes https://github.com/hypothesis/via/issues/1118.

### Test `models.Video`

```python
$ make shell
>>> video = models.Video(video_id="VIDEO_ID", title="TITLE")
>>> video
Video(id=None, video_id='VIDEO_ID')
>>> db.add(video)
>>> tm.commit()
>>> tm.begin()
<transaction._transaction.Transaction at 0x7f5235b13040>
>>> db.scalars(select(models.Video)).all()
[Video(id=1, video_id='VIDEO_ID')]
```

### Test `VideoFactory`

```python
>>> factories.VideoFactory()
Video(id=None, video_id='video_id_0')
>>> factories.VideoFactory()
Video(id=None, video_id='video_id_1')
```

### Test the web UI

<http://localhost:9083/https://www.youtube.com/watch?v=byN4S8WDPt8> (the video title is used for the tab's title and for the document title if you create an annotation)

### DIFF
--- a/via/models/video.py
+++ b/via/models/video.py
@@ -1,0 +1,11 @@
+from sqlalchemy.orm import Mapped, mapped_column
+
+from via.db import Base
+from via.models._mixins import AutoincrementingIntegerIDMixin, CreatedUpdatedMixin
+
+
+class Video(AutoincrementingIntegerIDMixin, CreatedUpdatedMixin, Base):
+    __tablename__ = "video"
+
+    video_id: Mapped[str] = mapped_column(unique=True)
+    title: Mapped[str] = mapped_column(repr=False)
--- a/via/services/youtube.py
+++ b/via/services/youtube.py
@@ -4,7 +4,7 @@ from sqlalchemy import select
 from sqlalchemy.exc import NoResultFound
 from youtube_transcript_api import YouTubeTranscriptApi
 
-from via.models import Transcript
+from via.models import Transcript, Video
 from via.services.http import HTTPService
 
 
@@ -67,9 +67,12 @@ class YouTubeService:
 
     def get_video_title(self, video_id):
         """Call the YouTube API and return the title for the given video_id."""
-        # https://developers.google.com/youtube/v3/docs/videos/list
+        if video := self._db.scalar(select(Video).where(Video.video_id == video_id)):
+            return video.title
+
         try:
-            return self._http_service.get(
+            # https://developers.google.com/youtube/v3/docs/videos/list
+            title = self._http_service.get(
                 "https://www.googleapis.com/youtube/v3/videos",
                 params={
                     "id": video_id,
@@ -80,6 +83,10 @@ class YouTubeService:
             ).json()["items"][0]["snippet"]["title"]
         except Exception as exc:
             raise YouTubeDataAPIError("getting the video title failed") from exc
+
+        self._db.add(Video(video_id=video_id, title=title))
+
+        return title
 
     def get_transcript(self, video_id):
         """
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,7 +7,7 @@ from pytest_factoryboy import register
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
-from tests.factories import TranscriptFactory
+from tests.factories import TranscriptFactory, VideoFactory
 from tests.factories.factoryboy_sqlalchemy_session import (
     clear_factoryboy_sqlalchemy_session,
     set_factoryboy_sqlalchemy_session,
@@ -16,6 +16,7 @@ from via.db import Base
 
 # Each factory has to be registered with pytest_factoryboy.
 register(TranscriptFactory)
+register(VideoFactory)
 
 
 @pytest.fixture
--- a/tests/factories/__init__.py
+++ b/tests/factories/__init__.py
@@ -1,1 +1,2 @@
 from tests.factories.transcript import TranscriptFactory
+from tests.factories.video import VideoFactory
--- a/tests/factories/video.py
+++ b/tests/factories/video.py
@@ -1,0 +1,12 @@
+from factory import Sequence
+from factory.alchemy import SQLAlchemyModelFactory
+
+from via.models import Video
+
+
+class VideoFactory(SQLAlchemyModelFactory):
+    class Meta:
+        model = Video
+
+    video_id = Sequence(lambda n: f"video_id_{n}")
+    title = Sequence(lambda n: f"video_title_{n}")
--- a/tests/unit/via/services/youtube_test.py
+++ b/tests/unit/via/services/youtube_test.py
@@ -6,7 +6,7 @@ from h_matchers import Any
 from requests import Response
 from sqlalchemy import select
 
-from via.models import Transcript
+from via.models import Transcript, Video
 from via.services.youtube import YouTubeDataAPIError, YouTubeService, factory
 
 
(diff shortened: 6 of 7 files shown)


via/models/_mixins.py
```diff
@@ -0,0 +1,16 @@
+from datetime import datetime
+
+from sqlalchemy import func
+from sqlalchemy.orm import Mapped, MappedAsDataclass, mapped_column
+
+
+class CreatedUpdatedMixin(MappedAsDataclass):
```
marcospri: These get added as the last columns, I don't think that the case in SQL1.4, not an issue.

via/models/video.py
```diff
@@ -0,0 +1,21 @@
...
+class VideoType(enum.Enum):
+    YOUTUBE = 1
+
+
+class Video(CreatedUpdatedMixin, Base):
+    __tablename__ = "video"
+    __table_args__ = (UniqueConstraint("type", "video_id"),)
+
+    id: Mapped[int] = mapped_column(primary_key=True, autoincrement=True, init=False)
+    type: Mapped[int] = mapped_column(Enum(VideoType))
```
marcospri: Over in LMS we settled in using a enum in the python side and leaving the postgres side as a varchar.

This maintains the python side benefits and makes migrations easier (well, trivial) at the expense of the enforcement on the DB side:

https://github.com/hypothesis/lms/blob/main/lms/db/_columns.py


With the custom sql type migrations need to be manually written which adds some extra work.

I prefer the approach over in LMS but this of course also works.

marcospri:
Testing this a a way to test the boilerplate in other PRs, doing:

- make services
- make db
- This PR doesn't include the migration, that needs to happen before this. Tested that also
- mkdir via/migrations/versions
- tox -e dev --run-command 'alembic revision --autogenerate -m "video"'
- tox -e dev --run-command 'alembic history'
- tox -e dev --run-command 'alembic upgrade head'
- tox -e dev --run-command 'alembic downgrade -1'
- tox -e dev --run-command 'alembic upgrade head'
- Got around https://github.com/hypothesis/via/pull/1019#discussion_r1267778787 comenting create_all.
- make sql works
- The actual purpose of this PR also works (ie new videos appear on the table)

via/models/_mixins.py
seanh: Yeah. There is actually a [`sort_order` argument to `mapped_column()`](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.mapped_column.params.sort_order) that we can potentially use to control the column order.

Actually, I wonder if we do need to fix this? Say in future we add a new `foo` column to the bottom of this model. After running the DB migration to add `foo` the order of the columns will be `id`, `type`, `video_id`, `title`, `created`, `updated`, `foo`. But if you re-created the DB and SQLAlchemy re-created the table from scratch then the order of the columns would be `id`, `type`, `video_id`, `title`, `foo`, `created`, `updated` :worried:

via/models/_mixins.py
```diff
@@ -0,0 +1,16 @@
...
+    created: Mapped[datetime] = mapped_column(
+        init=False,
+        server_default=func.now(),  # pylint:disable=not-callable
```
marcospri: I was holding the SQLA alchemy PR over LMS because this is annoying (not-callable)

I guess I'll have to bite the bullet and spirnkle a few of these there 
https://github.com/hypothesis/lms/pull/5283

via/models/_mixins.py
seanh: This got moved into https://github.com/hypothesis/via/pull/1072 (which this PR is now based on) and changed to put the `created` and `updated` columns at the start of the table

via/models/video.py
seanh: I ended up removing the `type` column: in https://github.com/hypothesis/via/pull/1072 we're not namespacing transcripts by type, so we don't need to do it for titles either. We can always add a `type` column to both tables in future if we ever add a second type of video. This enables us to avoid the whole enum thing for now

via/models/video.py
```diff
@@ -0,0 +1,12 @@
...
+from via.db import Base
+from via.models._mixins import CreatedUpdatedMixin
+
+
+class Video(CreatedUpdatedMixin, Base):
+    __tablename__ = "video"
+
+    id: Mapped[int] = mapped_column(primary_key=True, autoincrement=True, init=False)
+    video_id: Mapped[str] = mapped_column(unique=True)
+    title: Mapped[str] = mapped_column(repr=False)
```
seanh: As we've done with transcripts (over in https://github.com/hypothesis/via/pull/1072) I've decided not to namespace videos by type/source yet (e.g. by adding a `type="youtube"` column). We can add this later if and when we ever add support for a second type of video

via/services/youtube.py
```diff
@@ -67,9 +67,14 @@ def get_video_id(self, url):
 
     def get_video_title(self, video_id):
         """Call the YouTube API and return the title for the given video_id."""
-        # https://developers.google.com/youtube/v3/docs/videos/list
+        video = self._db.scalar(select(Video).where(Video.video_id == video_id))
+
+        if video:
```
marcospri: I guess is now idiomatic python to use `if video :=` here

via/models/video.py
```diff
@@ -0,0 +1,11 @@
+from sqlalchemy.orm import Mapped, mapped_column
+
+from via.db import Base
+from via.models._mixins import AutoincrementingIntegerIDMixin, CreatedUpdatedMixin
+
+
+class Video(AutoincrementingIntegerIDMixin, CreatedUpdatedMixin, Base):
+    __tablename__ = "video"
+
+    video_id: Mapped[str] = mapped_column(unique=True)
```
marcospri, line 10: Now that we already have transcripts in the DB it's probably too complicated to be worth it but we could have a FK in transcripts pointing to this table.

We can join now by video_id if we need to, it has an index in both tables.

marcospri (approved):
Looks good :+1: 

Added the `do-not-merge` tag here just as remainder to deploy and run the migration first

via/services/youtube.py
seanh: :+1: Nice! Done

via/models/video.py
seanh: Yeah, I thought about this but I didn't really think it was worth the complexity. We don't need an fkey in order to look up a `Video` given a `Transcript`, as you said. So I'm not sure an fkey would really buy us anything. In fact, I think we might actually _not_ want to enforce that every `Video` in the DB must have a matching `Transcript` or vice-versa, the order of operations (API calls etc) may call for the `Video` alone to first be saved and then the transcript not downloaded until a later request or vice-versa. I suppose this does mean that any code that has a `Video` cannot assume that a matching `Transcript` exists or vice-versa, which may lead to a little more code complexity at some point.

PR #1094, opened 2023-07-18
Add logic to handle toast messages (smarter `ToastMessages` container).

### LINKED ISSUE #1132: Add a toast message component for use in video-annotations application

acelaya:
This is PR introduces a different approach to handle toast messages (first one is https://github.com/hypothesis/via/pull/1089).

[Grabación de pantalla desde 2023-07-18 11-19-08.webm](https://github.com/hypothesis/via/assets/2719332/5ed84644-1362-4759-a60e-25ef076ad75d)

The main pieces are:

* A controlled `ToastMessages` component: it is similar to the one in client, but with the addition that it internally handles the logic to dismiss toast messages with a delay that allows for CSS animations to work.
  In some way it makes sense for this component to handle that, as the animations are an implementation detail of the component itself. From external consumers point of view, adding and dismissing toast messages is a matter of adding and removing items from a list.
* A new `AppProvider` component wrapping the entire video app, that we can use as a way to share a store across all components.
  It currently only handles one entry for toast messages, but it can be extended or replaced with something else, like redux.
* A `useToastMessages` hook, which reads the `toastMessages` from the app context, enhances it with a helper function to append and dismiss messages, and returns them together with the list of toast messages.
* A `VideoPlaterAppToastMessagess` component, which wraps the controlled `ToastMessages` components and feeds it with the result from the `useToastMessages` hook.
  This component can then be rendered wherever needed within the video app.

In order to verify the logic works and makes sense, this PR also enhances the copy-to-clipboard button, so that it displays a toast message after content has been copied, or an error has occurred.

### TODO / To think

- [ ] Consider extracting `ToastMessages` to frontend-shared, as it can be potentially used in client as well, with a couple of changes.
- [ ] Fix toast messages styling and positioning (currently rendered over the transcript, but spacing needs to be adjusted).

### DIFF
--- a/tailwind.config.mjs
+++ b/tailwind.config.mjs
@@ -8,6 +8,11 @@ export default {
   ],
   theme: {
     extend: {
+      animation: {
+        'fade-in': 'fade-in 0.3s forwards',
+        'fade-out': 'fade-out 0.3s forwards',
+        'slide-in-from-right': 'slide-in-from-right 0.3s forwards ease-in-out',
+      },
       borderRadius: {
         // Equivalent to tailwind default `rounded-sm` size
         DEFAULT: '0.125rem',
@@ -25,6 +30,37 @@ export default {
         // Similar to tailwind's default `shadow-inner` but coming from the
         // right edge instead of the left
         'r-inner': 'inset -2px 0 4px 0 rgb(0,0,0,.05)',
+      },
+      keyframes: {
+        'fade-in': {
+          '0%': {
+            opacity: '0',
+          },
+          '100%': {
+            opacity: '1',
+          },
+        },
+        'fade-out': {
+          '0%': {
+            opacity: '1',
+          },
+          '100%': {
+            opacity: '0',
+          },
+        },
+        'slide-in-from-right': {
+          '0%': {
+            opacity: '0',
+            left: '100%',
+          },
+          '80%': {
+            left: '-10px',
+          },
+          '100%': {
+            left: '0',
+            opacity: '1',
+          },
+        },
       },
     },
   },
--- a/via/static/scripts/video_player/components/AppProvider.tsx
+++ b/via/static/scripts/video_player/components/AppProvider.tsx
@@ -1,0 +1,43 @@
+import type { ComponentChildren } from 'preact';
+import { createContext } from 'preact';
+import type { StateUpdater } from 'preact/hooks';
+import { useState } from 'preact/hooks';
+
+import type { ToastMessage } from './ToastMessages';
+
+export type ToastMessagesState = {
+  toastMessages: ToastMessage[];
+  setToastMessages: StateUpdater<ToastMessage[]>;
+};
+
+export type AppStore = {
+  toastMessages: ToastMessagesState;
+};
+
+export const AppContext = createContext<AppStore>({
+  toastMessages: {
+    toastMessages: [],
+    setToastMessages: () => {},
+  },
+});
+
+export type AppProviderProps = {
+  children: ComponentChildren;
+};
+
+export default function AppProvider({ children }: AppProviderProps) {
+  const [toastMessages, setToastMessages] = useState<ToastMessage[]>([]);
+
+  return (
+    <AppContext.Provider
+      value={{
+        toastMessages: {
+          toastMessages,
+          setToastMessages,
+        },
+      }}
+    >
+      {children}
+    </AppContext.Provider>
+  );
+}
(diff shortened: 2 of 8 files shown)


via/static/scripts/video_player/components/AppProvider.tsx
```diff
@@ -0,0 +1,43 @@
...
+    setToastMessages: () => {},
+  },
+});
+
+export type AppProviderProps = {
+  children: ComponentChildren;
+};
+
+export default function AppProvider({ children }: AppProviderProps) {
+  const [toastMessages, setToastMessages] = useState<ToastMessage[]>([]);
```
acelaya, line 29: Using `useState` as the simplest way to handle the list of toast messages.

It can be later enhanced with adding/dismissing capabilities via helper functions.

via/static/scripts/video_player/hooks/use-toast-messages.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function useToastMessages(): ToastMessages {
+  const { toastMessages, setToastMessages } =
+    useContext(AppContext).toastMessages;
+  const appendToastMessage = (toastMessageData: ToastMessageData) => {
+    const id = generateHexString(10);
+    setToastMessages(messages => [...messages, { ...toastMessageData, id }]);
+  };
+  const dismissToastMessage = (id: string) =>
+    setToastMessages(messages => messages.filter(message => message.id !== id));
```
acelaya, line 23: We discussed having these helpers somewhere else, which perhaps would drive this hook useless. I left it here for convenience, but I'm open to discuss it.

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -0,0 +1,148 @@
...
+    [dismissMessage]
+  );
+
+  useEffect(() => {
+    messages.forEach(({ autoDismiss = true, id }) => {
+      if (autoDismiss) {
+        scheduleMessageDismiss(id);
+      }
+    });
+  }, [messages, scheduleMessageDismiss]);
```
acelaya, line 109: All these hooks/helpers allow to track which messages have been scheduled for dismissal (because they have `autoDismiss: true`) and which have already been dismissed and need to wait for the CSS animation before actually removing.

It is perhaps be too much logic for this component, but it's the only way to not leak implementation details. I'm mainly unhappy with the handling of auto-dismiss, but doing it from the outside of the component means the use of CSS animations needs to be known from the outside.

tailwind.config.mjs
acelaya, line 1: > Taken from client

via/static/scripts/video_player/utils/generate-hex-string.ts
acelaya, line 1: > Taken from client

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const dismissMessage = useCallback(
+    (id: string) => {
+      if (dismissedMessages.includes(id)) {
+        return;
+      }
+
+      // Track that this message has been dismissed, and once animation ends,
+      // actually remove it
+      setDismissedMessages(ids => [...ids, id]);
+      setTimeout(() => onMessageDismiss(id), 500);
```
lyzadanger, line 82: IIRC, the entire purpose of a delay on dismiss was to ensure the transition had time to complete before unmounting the message component. I think we have better tools in our toolbox now (e.g. listening for a transition-end event).

What would you think of initially landing a version of `ToastMessages` that doesn't handle dismiss delays or transitions at all, but adds them in a follow-on PR just focused on that? Could we reuse the `TransitionComponent` concept? Perhaps it would be helpful to get the core functionality of these toast messages sorted out without tripping over transitions?

via/static/scripts/video_player/components/AppProvider.tsx
lyzadanger, line 1: I think it's more consistent with our conventions to have a `Context` that is used by a component, and the component is responsible for setting the `value` on the `Provider`. Thus this would be `AppContext` and then the `index` module would be responsible for initializing the value of the context when wrapping the app with it.

i.e. I think it's the responsibility of the code that uses the context to set its value, versus the context itself.

via/static/scripts/video_player/components/ToastMessages.tsx
acelaya: Yeah, that sounds reasonable. It would let us land something usable earlier

acelaya:
Closing this in favor of https://github.com/hypothesis/via/pull/1095

PR #1095, opened 2023-07-19
Add logic to handle toast messages (preact/signals)

### LINKED ISSUE #1132: Add a toast message component for use in video-annotations application

acelaya:
> **Note**
> This PR is easier to review by ignoring whitespaces.

This PR is a slightly different possible approach to handle toast messages, which differs from https://github.com/hypothesis/via/pull/1089 and https://github.com/hypothesis/via/pull/1094 in:

* The controlled `ToastMessages` component introduced in the two previous PRs is still there. However, anything related with animations has been removed, as it simplifies the logic to dismiss toast mesages.
  The concept of auto-dismiss is still there, but considering we only have a single possible toast message which will almost never happen, we can even get rid of that.
* The toast messages state is handled via `@preact/signals`. This becomes the first use case in any of our projects, so let's review carefully and provide feedback.
  > The first commit uses `useState`, and the second commit introduces signals. In case you want to check the diff.
* The `AppContext` and `VideoPlayerAppToastMessages` introduced in https://github.com/hypothesis/via/pull/1094 are still there.

### Todo

- [ ] Fix tests
- [ ] Decide proper positioning and styling for `VideoPlayerAppToastMessages`.

### Testing steps

1. Open `VideoPlayerApp.tsx`.
2. Right before `await navigator.clipboard.writeText(formattedTranscript);`, add this line.
    ```diff
    + throw new Error('Hi!`);
    await navigator.clipboard.writeText(formattedTranscript);
    ```
3. Open http://localhost:9083/https://www.youtube.com/watch?v=MrORaCfcKxQ and click the copy-to-clipboard button. It should display a toast message every time you click it.
4. Toast messages should be dismissed after 5 seconds.
5. Clicking on a toast message should dismiss it immediatelly.

[Grabación de pantalla desde 2023-07-19 08-37-55.webm](https://github.com/hypothesis/via/assets/2719332/a998ef17-31c9-45a9-b9ce-9b1d77c108e3)

### DIFF
--- a/via/static/scripts/video_player/components/AppContext.tsx
+++ b/via/static/scripts/video_player/components/AppContext.tsx
@@ -1,20 +1,12 @@
+import type { Signal } from '@preact/signals';
 import { createContext } from 'preact';
-import type { StateUpdater } from 'preact/hooks';
 
 import type { ToastMessage } from './ToastMessages';
 
-export type ToastMessagesState = {
-  toastMessages: ToastMessage[];
-  setToastMessages: StateUpdater<ToastMessage[]>;
-};
-
 export type AppStore = {
-  toastMessages: ToastMessagesState;
+  toastMessages: Signal<ToastMessage[]>;
 };
 
 export const AppContext = createContext<AppStore>({
-  toastMessages: {
-    toastMessages: [],
-    setToastMessages: () => {},
-  },
-});
+  toastMessages: {},
+} as AppStore);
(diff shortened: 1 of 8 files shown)


via/static/scripts/video_player/components/AppContext.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import type { ToastMessage } from './ToastMessages';
+
+export type AppStore = {
+  toastMessages: Signal<ToastMessage[]>;
+};
+
+export const AppContext = createContext<AppStore>({
+  toastMessages: {},
+} as AppStore);
```
acelaya, line 12: Just noticed this is wrong. My intention was to initialize the context as `{} as AppStore`.

Not saying that's the right thing to do, as perhaps we should set `toastMessages` optional, but definitely the actual code I pushed is not the intended one.

acelaya:
Closing in favor of something simpler.

PR #1096, opened 2023-07-19
Set a default minimum font size of 16px for inputs on touch devices

### LINKED ISSUE #1061: Increase text size of transcript search field to prevent zoom on iOS
The transcript search field has a font size < 16px, which causes iOS to zoom into it when focused. We have a solution for this in the client for the annotation edit field. I guess we'll want to do something similar here, or add a more general solution?

In testing this, I noticed that the search bar at the top of the client's UI has the same issue.

robertknight:
Previously the font size for inputs was taken from the `text-sm` default for `body`, which evaluates to 14px. This is below the threshold on iOS that causes the browser to zoom into the field when focused.

Fixes #1061

See also https://tailwindcss.com/docs/preflight#extending-preflight

**Testing:**

1. Open a video
2. Toggle the mobile device mode in the Chrome devtools toolbar. The search input should be 16px when in mobile mode and 14px otherwise

If testing with the local dev server on an actual mobile device, you will need to edit the [`client_embed_url` setting](https://github.com/hypothesis/via/blob/a89f260c8aa908dd6ef9a8f9c3bca6f4dbab35e8/conf/development.ini#L6) to point to the IP of your dev system, in order for the client to load.

### DIFF
--- a/via/static/styles/video_player.css
+++ b/via/static/styles/video_player.css
@@ -3,14 +3,25 @@ CSS entry point for the video player app.
 */
 
 @tailwind base;
-
-body {
-  @apply bg-grey-0;
-  @apply font-sans text-sm text-color-text;
-}
-
 @tailwind components;
 @tailwind utilities;
+
+@layer base {
+  body {
+    @apply bg-grey-0;
+    @apply font-sans text-sm text-color-text;
+  }
+
+  /*
+  Input fields must have a minimum size on iOS to avoid the browser zooming in
+  when they are focused.
+  */
+  @media(pointer: coarse) {
+    input, textarea {
+      font-size: max(16px, 100%);
+    }
+  }
+}
 
 p::highlight(transcript-filter-match) {
   text-decoration: underline;


via/static/styles/video_player.css
```diff
@@ -3,15 +3,26 @@ CSS entry point for the video player app.
...
-
-body {
-  @apply bg-grey-0;
-  @apply font-sans text-sm text-color-text;
-}
-
 @tailwind components;
 @tailwind utilities;
 
+@layer base {
```
robertknight, line 9: This follows Tailwind's recommendation for extending base styles: https://tailwindcss.com/docs/preflight#extending-preflight. If there is a reason we shouldn't be doing that, please let me know!

via/static/styles/video_player.css
```diff
@@ -3,15 +3,26 @@ CSS entry point for the video player app.
...
+
+  /*
+  Input fields must have a minimum size on iOS to avoid the browser zooming in
+  when they are focused.
+  */
+  @media(pointer: coarse) {
+    input, textarea {
+      font-size: max(16px, 100%);
+    }
+  }
```
robertknight, line 23: Should we upstream this into the frontend-shared package?

via/static/styles/video_player.css
lyzadanger: Yep, this is fine and conventional.

via/static/styles/video_player.css
lyzadanger: Actually, I want to think about that. I've found so far that it's easier to wrangle base styles per-app, as that's where our variation tends to be. Not sure. Perhaps we could track in a separate ticket?

lyzadanger (approved):
👍🏻

via/static/styles/video_player.css
robertknight: > Perhaps we could track in a separate ticket?

Will do 👍

robertknight: https://github.com/hypothesis/frontend-shared/issues/1159

PR #1099, opened 2023-07-19
Add a model for parsing data from YouTube V1 video info API

jon-betts:
# :warning: READ ME FIRST

I was blindly copying things from the way they were done in the library we used in certain places. I don't think we need to check for translation languages at all. As a result there is a shorter PR here without that as an alternative to this:

 * https://github.com/hypothesis/via/pull/1107

So we can review that instead of this PR.

----

The new YouTube stuff will come in a few different major parts:

 * A client which retrieves JSON data and deals with transcript XML
 * A model (this PR) which interprets that JSON data
 * All the plumbing with the rest of the system to use this

This PR doesn't do anything on it's own, but marks a sensible chunk of code worth reviewing. 

## Review notes

 * The model follows along with the names and concepts of the underlying JSON very closely.
 * This looks like a lot, but it's bloated out with a lot of docstrings and comments

## Testing notes

If you want to go through a kind of functional test you can create the root object and have a play about with it. You could do that like this:

 * Get a blob of JSON data for a video ([some help with that here](https://hypothes-is.slack.com/archives/C1MA4E9B9/p1689769029071919)) 
```python
from h_matchers import Any

from via.services.youtube_api import Video, CaptionTrack

my_json = {
    # ... Your JSON here
}

video = Video.from_json(url="http://example.com", data=my_json)

# Basic info
print(video.id)
print(video.details)
print(video.caption.tracks)
print(video.caption.translation_languages)

# Try the matching system
caption_track = video.caption.find_matching_track([
    CaptionTrack("en-gb"),
    CaptionTrack("en", name=Any(), kind=Any()),
])
print(caption_track)
```

You can also instantiate individual parts of it to play around with.


seanh:
Closing in favour of https://github.com/hypothesis/via/pull/1107

PR #1102, opened 2023-07-19
Revert "Revert "Add a Postgres DB to Via""

seanh:
Reverts hypothesis/via#1101 and adds one new commit that fixes the issue with the Docker build.

https://github.com/hypothesis/via/pull/1077 broke the Docker build (`make docker` was broken), see https://github.com/hypothesis/via/pull/1101. This PR re-applies https://github.com/hypothesis/via/pull/1077 and adds a new commit that fixes the `Dockerfile` so that `make docker` should work.

Only the second commit in this PR needs to be reviewed, the rest was already reviewed with https://github.com/hypothesis/via/pull/1077.


Dockerfile
```diff
@@ -18,8 +18,7 @@ RUN yarn build
 FROM python:3.8.17-slim-bullseye
 LABEL maintainer="Hypothes.is Project and contributors"
 
-# Install nginx & supervisor
-RUN apt-get update && apt-get install --yes nginx nginx-extras gettext-base supervisor \
+RUN apt-get update && apt-get install --yes nginx nginx-extras gettext-base supervisor libpq-dev \
```
seanh, line 21: I _think_ `libpq-dev` (which contains the `pg_config` command) needs to be left in the container, it's not only needed for the `pip install` command.

Dockerfile
```diff
@@ -33,8 +32,11 @@ RUN chown -R hypothesis:hypothesis /etc/nginx/conf.d /var/log/nginx /var/lib/ngi
 COPY requirements/prod.txt ./
 
 # Install build deps, build, and then clean up.
-RUN pip install --no-cache-dir -U pip \
-  && pip install --no-cache-dir -r prod.txt
+RUN apt-get install --yes gcc \
+  && pip install --no-cache-dir -U pip \
+  && pip install --no-cache-dir -r prod.txt \
+  && apt-get remove --yes gcc \
+  && apt-get autoremove --yes
```
seanh, line 39: We only need `gcc` during the `pip install` command, we don't need to leave it in the container, so I installed it just before it's needed and then removed it again.

Why do we need `gcc` though? It's because our `Dockerfile` is building `psycopg2` from source. I wonder if there's a different fix here where we could get `pip` to install a pre-built `psycopg2` binary?

marcospri: There's https://pypi.org/project/psycopg2-binary/

but the docs there say:

```
You can also obtain a stand-alone package, not requiring a compiler or
external libraries, by installing the `psycopg2-binary`_ package from PyPI::

    $ pip install psycopg2-binary

The binary package is a practical choice for development and testing but in
production it is advised to use the package built from sources.
```

marcospri (approved):

PR #1100, opened 2023-07-19
Add a client for getting data from YouTube

jon-betts:
Requires:

 * https://github.com/hypothesis/via/pull/1107

## Review notes

This PR uses an undocumented API for getting the JSON information. It seems the same data as the HTML scraping method, but avoids:

 * Getting the JSON from the HTML
 * Getting the HTML in the first place
 * Dealing with cookie consent issues in the HTML

This is all being done a bit skin of the teeth. I'd ideally like to try more error scenarios, but I guess we'll catch them later.
The one I've spotted so far is that asking for a translated 

## Testing notes

You can play about as you'd like, but here is something to get you started.

```python
from dataclasses import asdict
import json

from via.services.youtube_api import YouTubeAPIClient

client = YouTubeAPIClient()

# Get the video info
video = client.get_video_info("1v8u3ua6BPk")
print("Video details:")
print("\t", video.details.title)
print("\t", video.details.author)
print("\t", video.url)

# Try the matching system
caption_track = video.caption.tracks[0]
print("\nSelecting caption track:")
print("\t", caption_track)

# Get the transcript
transcript = client.get_transcript(caption_track)
print("\nTranscript:")
print(json.dumps(asdict(transcript), indent=4))
```

PR #1103, opened 2023-07-19
Integrate the first party YouTube client into Via

jon-betts:
Requires:

 * https://github.com/hypothesis/via/pull/1100
 * https://github.com/hypothesis/via/pull/1104

## Review notes

 * This includes an enormous commit which is mostly just format changes in the requirements
 * This is looking up a matching transcript based on a list of preferences
    * I don't think we should do this for LMS
    * I think the LMS should either do this for us, or make the user pick

## Testing notes

 * `make dev`
 * Visit: http://localhost:9083/https://www.youtube.com/watch?v=Xzv84ZdtlE0
 * You should see a transcript (it should be `en`, though not sure how you'd tell without a break point)
 * http://localhost:9083/https://www.youtube.com/watch?v=Xzv84ZdtlE0&via.video.lang=en.a
 * Should work again
 * http://localhost:9083/https://www.youtube.com/watch?v=Xzv84ZdtlE0&via.video.lang=en...fr
 * Should take a while, and potentially timeout. On a repeat attempt you should see it in French


tests/unit/services.py
```diff
@@ -62,10 +62,4 @@ def pdf_url_builder_service(mock_service):
 
 @pytest.fixture
 def youtube_service(mock_service):
-    youtube_service = mock_service(YouTubeService)
-    youtube_service.enabled = True
-
-    # By default, the mock reports that URLs are not YouTube URLs.
-    youtube_service.get_video_id.return_value = None
-
-    return youtube_service
```
jon-betts, line 71: This violated the principal of least surprise for me. This is only required in one test, so I moved it there.

tests/unit/via/views/view_video_test.py
```diff
@@ -59,17 +60,55 @@ def test_it(
...
+        if not has_matches:
+            video.caption.find_matching_track.return_value = None
+
+        response = view_youtube_video(pyramid_request)
+
+        assert response["api"]["transcript"]["url"] == pyramid_request.route_url(
+            "api.youtube.transcript",
+            video_id=youtube_service.get_video_id.return_value,
+            transcript_id="en.a",
+        )
```
jon-betts, line 99: These tests are very intense. I really don't think we should be doing this anyway here which might reduce this effect

via/views/view_video.py
```diff
@@ -8,26 +9,44 @@
...
-    youtube_service = request.find_service(YouTubeService)
+def view_youtube_video(request, url, **kwargs):
+    youtube_service: YouTubeService = request.find_service(YouTubeService)
 
     if not youtube_service.enabled:
         raise HTTPUnauthorized()
 
     video_id = youtube_service.get_video_id(url)
-    video_url = youtube_service.canonical_video_url(video_id)
-    video_title = youtube_service.get_video_title(video_id)
```
jon-betts, line 25: I think part of the reason these tests aren't as nice as they could be is we have too many atomic calls.

The new method returns a single object with all these items inside it. It would be good to unify these.

via/views/view_video.py
```diff
@@ -8,26 +9,44 @@
...
-
     if not video_id:
         raise BadURL(f"Unsupported video URL: {url}", url=url)
 
-    _, client_config = Configuration.extract_from_params(kwargs)
+    video_url = youtube_service.canonical_video_url(video_id)
+    video_title = youtube_service.get_video_title(video_id)
```
jon-betts, line 47: I moved these down, because we can early return if the video id isn't valid. Rather than doing lots of strange stuff.

via/views/view_video.py
```diff
@@ -38,11 +57,34 @@ def youtube(request, url, **kwargs):
...
     }
+
+
+def _get_transcript_id(youtube_service, url, via_config):
+    # Try the `via.video.lang` param first
+    if transcript_id := via_config.get("video", {}).get("lang"):
+        return transcript_id
+
+    # Then look up a transcript matching our preferences
+    video = youtube_service.get_video_info(video_url=url)
```
jon-betts, line 82: We could look this up in the parent, but I chose to keep it separate for now.

We could look to unify the look up with a call like:

```
youtube_service.get_video_info(video_url=url, get_captions=True)
```

Where we could use V3 stuff if `get_captions=False` or the V1 stuff if it's true etc.

seanh:
Labeling this as blocked just because it's waiting for me to get some more urgent PRs done before I can review this. I promise I'm working my way round to reviewing this ASAP!

via/services/youtube.py
```diff
@@ -75,19 +87,32 @@ def get_video_title(self, video_id):
         except Exception as exc:
             raise YouTubeDataAPIError("getting the video title failed") from exc
 
-    def get_transcript(self, video_id):
-        """
-        Call the YouTube API and return the transcript for the given video_id.
+    def get_video_info(self, video_id=None, video_url=None) -> Video:
+        if video_url:
+            video_id = self.get_video_id(video_url)
```
seanh, line 92:
```suggestion
    def get_video_info(self, video_id) -> Video:
```

The `video_url` argument isn't needed: this is only called by `get_transcript()` below and it always passes `video_id` never `video_url`

via/views/view_video.py
```diff
@@ -38,11 +57,34 @@ def youtube(request, url, **kwargs):
...
                 },
             }
         },
     }
+
+
+def _get_transcript_id(youtube_service, url, via_config):
+    # Try the `via.video.lang` param first
+    if transcript_id := via_config.get("video", {}).get("lang"):
+        return transcript_id
```
seanh, line 79:
```suggestion
```

We should remove this as it'll be dead code: nothing is ever going to be passing the `via.video.lang` param. We can add support for this query param back in a later PR when we need it

seanh (changes requested):
Not a proper review yet but just as a first-pass: I think the `video_url` argument to `get_video_info()` can be removed as it's unused, and the `via.video.lang` query param should be removed as nothing's going to be using it

seanh:
Closing this, I think it was superseded by https://github.com/hypothesis/via/pull/1160

PR #1105, opened 2023-07-20
Limit video height to half window height in single-column layouts

### LINKED ISSUE #1085: Wide-but-short video player window can lead to zero height transcript
If the video player window is wide but short, the transcript container can end up getting resized to zero height when the sidebar is open. In this starting layout:

Wide player - sidebar closed

Opening the sidebar leads to:

Wide player - sidebar open

The main situation where this has come up so far for me is when the browser devtools are open, since that squashes the viewport. However this could happen for anyone.

robertknight:
In single column layouts with "short" window heights, limit the height of the video to a maximum of half the window height, to ensure there is a good amount of space for the transcript. Since videos have a 16:9 aspect ratio, this means not using the full width and centering the video instead.

To make it easier to keep `videoWidth` in sync with the other properties returned by `useAppLayout`, the internals were refactored to separate the external state (the container width and height) from the properties that can be computed from that state (app size enum value, multicolumn, max video width).

Fixes https://github.com/hypothesis/via/issues/1085

----

**Before:**

A single column layout where giving the video full width makes the transcript too short:

Too-short transcript

(See also https://github.com/hypothesis/via/issues/1085 for more extreme examples)

**After:**

Better-sized transcript

Note that in multi-column layouts, and single-column layouts where a full-width video previously resulted in a transcript having >= 50% of the viewport height, the layout should look the same as before.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -109,7 +109,7 @@ export default function VideoPlayerApp({
   const filterInputRef = useRef<HTMLInputElement>();
   const appContainerRef = useRef<HTMLDivElement | null>(null);
 
-  const { appSize, multicolumn, transcriptWidth } =
+  const { appSize, multicolumn, transcriptWidth, videoWidth } =
     useAppLayout(appContainerRef);
   const sideBySideActive = useSideBySideLayout();
 
@@ -315,6 +315,10 @@ export default function VideoPlayerApp({
               // multicolumn layouts (NB: It will take up full width in
               // single-column)
               grow: multicolumn,
+
+              // In single-column layouts, the video may be narrower than the
+              // column width. Center the video in the column.
+              'ml-auto mr-auto': typeof videoWidth !== 'undefined',
             },
             {
               // Adapt spacing around video for different app sizes
@@ -324,6 +328,7 @@ export default function VideoPlayerApp({
               'py-2 px-4': appSize === 'xl',
             }
           )}
+          style={{ width: videoWidth ? `${videoWidth}px` : undefined }}
         >
           <YouTubeVideoPlayer
             videoId={videoId}
--- a/via/static/scripts/video_player/hooks/test/use-app-layout-test.js
+++ b/via/static/scripts/video_player/hooks/test/use-app-layout-test.js
@@ -6,20 +6,15 @@ import { useAppLayout } from '../use-app-layout';
 
 describe('useAppLayout', () => {
   let container;
+  let lastAppLayout;
   let wrappers;
 
   // Create a fake component to mount in tests that uses the hook
   function FakeComponent() {
     const myRef = useRef();
-    const { appSize, multicolumn } = useAppLayout(myRef);
+    lastAppLayout = useAppLayout(myRef);
     return (
-      <div
-        ref={myRef}
-        data-app-size={appSize}
-        data-app-multicolumn={multicolumn}
-        style="width:100%"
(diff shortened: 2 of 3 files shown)


robertknight:
There is one state where the new layout is arguably suboptimal, which is when all of the following apply:

- Single-column layout
- Side-by-side not active
- Sidebar open
- Window is narrow, so video doesn't use full width

In this case because the video is centered, it is partly hidden behind the sidebar:

Single-column no side-by-side

As soon as you close the sidebar however, the centered layout becomes optimal again.

We could perhaps improve upon this by left-aligning the video when the sidebar is open. I think I'd like to leave that as a separate task though.

lyzadanger:
@robertknight I'll take a look at this later today, time allowing. I have a sense there might be a simpler way to approach this, but obviously I need to prove that :).

lyzadanger:
@robertknight I did indeed find what appears to be a CSS-only approach here. We can't set a max-height on the video container because that can potentially cut off the bottom of the video because of the way `AspectRatio` is implemented. We _can_, however, define a max width. If we set that max width based on our desired max height (taking the 16:9 aspect ratio into account), we can make this work.

My example here sets a maximum height of `40vh` (versus your 50%), but you can plug any value in that you prefer.

Note that the video cannot be inside a flex container for this to work (specifically the auto margins for centering), but that was easy enough to adjust.

This works even more simply on my branch where I've converted the layout to grid instead of a bunch of nested flex-es, but I don't want to introduce that much change right now.

NB: This patch is against `main`, not this branch.

```diff
diff --git a/via/static/scripts/video_player/components/VideoPlayerApp.tsx b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
index df9f843..4d8ad50 100644
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -328,12 +328,11 @@ export default function VideoPlayerApp({
         <div
           data-testid="embedded-video-container"
           className={classnames(
-            'flex flex-col',
             {
               // Allow video to grow with available horizontal space in
               // multicolumn layouts (NB: It will take up full width in
               // single-column)
-              grow: multicolumn,
+              'flex flex-col grow': multicolumn,
             },
             {
               // Adapt spacing around video for different app sizes
@@ -344,13 +343,23 @@ export default function VideoPlayerApp({
             }
           )}
         >
-          <YouTubeVideoPlayer
-            videoId={videoId}
-            play={playing}
-            time={timestamp}
-            onPlayingChanged={setPlaying}
-            onTimeChanged={setTimestamp}
-          />
+          <div
+            className={classnames({
+              // Constrain maximum height of video to 40vh in single-column
+              // views by constraining the width of the embedded video based on
+              // a 16:9 aspect ratio. Center the video if it doesn't fill the
+              // full width.
+              'max-w-[calc(40vh*16/9)] mx-auto': !multicolumn,
+            })}
+          >
+            <YouTubeVideoPlayer
+              videoId={videoId}
+              play={playing}
+              time={timestamp}
+              onPlayingChanged={setPlaying}
+              onTimeChanged={setTimestamp}
+            />
+          </div>
         </div>
         <div
           data-testid="transcript-and-controls-container"
```

robertknight:
Thanks @lyzadanger, that's great! I created a new PR based on that patch in https://github.com/hypothesis/via/pull/1115. I kept the 40vh height from the patch. I suspect we get users asking for more control over the amount of space given to the video vs transcript, but we can address that later.

robertknight:
I might still extract some of the internal refactoring in the `useAppLayout` hook from this PR, to separate computed and non-computed state.

PR #1104, opened 2023-07-19
Run `make requirements` with no changes to requirements

jon-betts:
No idea what's going on here, but this causes a total mess in subsequent PRs.

PR #1107, opened 2023-07-20
A simpler YouTube V1 model, without translation information

jon-betts:
Slimer alternative to:

 * https://github.com/hypothesis/via/pull/1099

Requires:

 * https://github.com/hypothesis/via/pull/1104

The V3 API doesn't seem to have the concept of a track being non-translatable. It also seems to accept all languages as options. We can therefore simplify our code by not storing and comparing these values.

This version also strips out all unnecessary metadata info not related to captions, but retains the structure to add them back in.


seanh:
Labeling this as blocked just because it's waiting for me to get some more urgent PRs done before I can review this. I promise I'm working my way round to reviewing this ASAP!

via/services/youtube_api/__init__.py
```diff
@@ -0,0 +1 @@
+from via.services.youtube_api.models import *
```
seanh: I don't think you want to use `import *` here as this will import everything from `models.py` including for example things that `models.py` imports (`base64`, `deepcopy`, ...), any private methods in `models.py`, etc.

You could add an `__all__` to `models.py` but I think it'd be better to follow our usual pattern and explicitly import each individual name that you want in the `__init__.py`.

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+
+    name: Optional[str] = None
+    """Human set name for the track."""
+
+    kind: str = None
+    """Is this track automatically generated by audio to text AI?"""
+
+    translated_language_code: Optional[str] = None
+    """Language to machine translate this into. Set this manually."""
+
```
seanh:
```suggestion
```

Discussed in Slack: I'd like to move the machine translation support into a separate PR.

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+    """A source of transcription data, in a particular language."""
+
+    # Items which form the unique part
+    language_code: str
+    """Original language of the track."""
+
+    name: Optional[str] = None
+    """Human set name for the track."""
+
+    kind: str = None
```
seanh:
```suggestion
    kind: Optional[str] = None
```

Hazards of not running a type checker I guess

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+from typing import List, Optional
+
+from via.services.youtube_api._nested_data import safe_get
+
+
+@dataclass
+class CaptionTrack:
+    """A source of transcription data, in a particular language."""
+
+    # Items which form the unique part
```
seanh:
```suggestion
```

It's unclear what this comment refers to. Items that form the unique part of what? But I think this form of comments is usually not a good idea:

```python
# Items that form the unique part of the `id`.

...

# Other items.

...
```

What tends to happen over time is that some non-unique items get added under the `# Items that form the unique part` comment and/or some unique items get added under the `# Other items` comment. See https://github.com/hypothesis/via/pull/1078 for a recent example of this. It's going to be better to put these kinds of comments in the `id()` method itself since that's what they're about

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+    name: Optional[str] = None
+    """Human set name for the track."""
+
+    kind: str = None
+    """Is this track automatically generated by audio to text AI?"""
+
+    translated_language_code: Optional[str] = None
+    """Language to machine translate this into. Set this manually."""
+
+    # Other items which we cannot determine from the id
```
seanh:
```suggestion
```

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+            )
+        )
+
+        if name := data.get("name"):
+            data["name"] = base64.b64decode(name.encode("utf-8")).decode("utf-8")
+
+        if data.pop("auto_generated", None):
+            data["kind"] = "asr"
+
+        return cls(**data)
```
seanh:
```suggestion
```

This method isn't needed: what Via needs to do is get the list of transcript dicts from the JSON API, make a `CaptionTrack` object from each transcript dict, and then choose one of those `CaptionTrack` objects to be the default choice. It doesn't need to make `CaptionTrack` objects from IDs.

If in a later PR a need for this `from_id()` method comes up, then can we add this method then in that PR?

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+        else:
+            name = None
+
+        return ".".join(
+            part or ""
+            for part in [
+                self.language_code,
+                "a" if self.is_auto_generated else None,
+                name,
+                self.translated_language_code,
```
seanh:
```suggestion
```

(We don't want machine translation support for now, as discussed in an earlier comment.)

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+
+    @property
+    def is_auto_generated(self) -> bool:
+        """Get whether this caption track auto generated."""
+
+        return self.kind == "asr"
+
+    @is_auto_generated.setter
+    def is_auto_generated(self, value: bool):
+        self.kind = "asr" if value else None
```
seanh:
```suggestion
```

I don't think we ever need to set `is_auto_generated`?

via/services/youtube_api/models.py
```diff
@@ -0,0 +1,219 @@
...
+        """Get the URL to download a transcript of this caption track."""
+        if not self.base_url:
+            return None
+
+        url = self.base_url
+
+        if self.translated_language_code:
+            url += f"&tlang={self.translated_language_code}"
+
+        return url
```
seanh:
```suggestion
```

If we're not adding the machine translation support yet then this method isn't needed: code can just ise `caption_track.base_url` directly.

seanh (changes requested):
Not a full review (yet) but I'm just pointing out some unneeded/out-of-scope code that we should remove

seanh:
Closing in favour of https://github.com/hypothesis/via/pull/1162

PR #1111, opened 2023-07-20
Add extra details required by Via and LMS to the YouTube model

jon-betts:
Based on:

 * https://github.com/hypothesis/via/pull/1107

Before we were grabbing some random indicators of playablity and various metadata. This version is targeted at the values required in LMS. This is in preparation for this code to go into a library.

This also covers all metadata requirements of Via as well.

All of the objects here are intended to be V3 data compatible.


seanh:
Closing this: we don't have any plans to move this code into a library or to reuse it in LMS, so we don't need this change

PR #1112, opened 2023-07-20
Allow building a video object from V3 format data

jon-betts:
Based on:

 * https://github.com/hypothesis/via/pull/1111

This allows the construction of the video object from V3 formatted data.

This means we can have a method like:

```python
def get_video_details(self, video_id, with_captions=False):
    if with_captions:
         return Video.from_v1_json(...)

   return Video.from_v3_json(...)
```

And get identical results out (minus the caption information).


seanh:
Closing this: we don't have any plans to move this code into a library or to re-use it in LMS, so we don't need this change

PR #1113, opened 2023-07-20
Unified access to YouTube data via the client

jon-betts:
Based on:

 * https://github.com/hypothesis/via/pull/1112
 * https://github.com/hypothesis/via/pull/1100
 * https://github.com/hypothesis/via/pull/1103

This PR:

 * Moves video data getting responsibility to the YouTube client object
 * Unifies the V1 and V3 paths
 * Simplifes the service a lot as a result

## Review notes

This isn't ready yet, we need a lot of other stuff first.

If you look at this, nearly every commit is from other branches I've picked in.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -61,6 +61,8 @@ gunicorn==20.1.0
     # via -r requirements/prod.txt
 h-assets==1.0.5
     # via -r requirements/prod.txt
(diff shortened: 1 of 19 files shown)


seanh:
Closing this: we don't have any plans to move this code into a library or to re-use it in LMS, so we don't need this change

PR #1114, opened 2023-07-21
Extract "Copy transcript" button into a separate component

robertknight:
Extract this out of VideoPlayerApp since we expect to add some extra complexity here soon (see notes in https://github.com/hypothesis/lms/pull/5586 about iframed assignments). There are no functional or visual changes.


acelaya (approved):
👍🏼

PR #1115, opened 2023-07-24
Limit video height in single column view to leave space for transcript

### LINKED ISSUE #1085: Wide-but-short video player window can lead to zero height transcript

robertknight:
In single column layouts with "short" window heights, limit the height of the video to a maximum of ~~half~~ 40% the window height, to ensure there is a good amount of space for the transcript. Since videos have a 16:9 aspect ratio, this means not using the full width and centering the video instead.

This is based on @lyzadanger's patch in https://github.com/hypothesis/via/pull/1105#issuecomment-1645814269.

Fixes https://github.com/hypothesis/via/issues/1085

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -324,12 +324,11 @@ export default function VideoPlayerApp({
         <div
           data-testid="embedded-video-container"
           className={classnames(
-            'flex flex-col',
             {
               // Allow video to grow with available horizontal space in
               // multicolumn layouts (NB: It will take up full width in
               // single-column)
-              grow: multicolumn,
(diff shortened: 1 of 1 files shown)


lyzadanger:
> maximum of half the window height

Pedantically, it's 40% :)

lyzadanger (approved):
LGTM!

robertknight:
> Pedantically, it's 40% :)

My bad. I copied the description from the previous PR and forgot to update this.

PR #1116, opened 2023-07-24
Add `data-time-{start, end}` attributes to segments

robertknight:
This enables the Hypothesis client to capture media time selectors with annotations, and later use them when anchoring, or for other purposes.

On this branch, the `<p>` elements containing transcript text now have attributes to indicate the corresponding time range of the video. I had a look around for possible existing ARIA or schema.org metadata, but I didn't find anything that looked like an ideal fit, so these are custom attrs. See the client PR for more details.

date-time-attrs

There are a couple of caveats about the precision of timestamps and the end timestamp for the last segment. See notes in the code.

See https://github.com/hypothesis/client/pull/5655.

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -415,7 +415,7 @@ export default function Transcript({
                   startTime={segment.start}
                   // We are currently missing an end time for the last segment
                   // because we don't know the duration of the video here.
-                  endTime={nextSegment ? nextSegment.start : undefined}
+                  endTime={nextSegment?.start}
                   text={segment.text}
                 />
               );
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -87,7 +87,8 @@ type TranscriptSegmentProps = {
 
   isCurrent: boolean;
   onSelect: () => void;
-  time: number;
+  startTime: number;
+  endTime?: number;
   text: string;
 };
 
@@ -101,7 +102,8 @@ function TranscriptSegment({
   matches,
   isCurrent,
   onSelect,
-  time,
+  startTime,
+  endTime,
   text,
 }: TranscriptSegmentProps) {
   const contentRef = useRef<HTMLParagraphElement>(null);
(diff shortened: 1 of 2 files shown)


via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -385,22 +396,30 @@ export default function Transcript({
...
-                isCurrent={index === currentIndex}
-                matches={filterMatches?.get(index)}
-                onSelect={() => onSelectSegment?.(segment)}
-                time={segment.start}
-                text={segment.text}
-              />
-            ))}
+            {transcript.segments.map((segment, index, segments) => {
+              const nextSegment =
+                index < segments.length - 1 ? segments[index + 1] : undefined;
```
acelaya, line 401: I think `const nextSegment = segments[index + 1]` should have the same result, as `segments[index + 1]` will be undefined when `index + 1 >= segments.length`.

```suggestion
              const nextSegment = segments[index + 1];
```

via/static/scripts/video_player/components/Transcript.tsx
```diff
@@ -385,22 +396,30 @@ export default function Transcript({
...
+                      : false
+                  }
+                  highlight={highlight}
+                  isCurrent={index === currentIndex}
+                  matches={filterMatches?.get(index)}
+                  onSelect={() => onSelectSegment?.(segment)}
+                  startTime={segment.start}
+                  // We are currently missing an end time for the last segment
+                  // because we don't know the duration of the video here.
+                  endTime={nextSegment ? nextSegment.start : undefined}
```
acelaya:
```suggestion
                  endTime={nextSegment?.start}
```

acelaya (approved):
Looking good

via/static/scripts/video_player/components/Transcript.tsx
robertknight: It does at runtime, but not in TS's type system, at least with our current TS config. Doing an explicit check allows TS to know that `nextSegment` might be undefined.

via/static/scripts/video_player/components/Transcript.tsx
robertknight: Will do

PR #1117, opened 2023-07-24
Dispatch dummy scroll event when filter changes

robertknight:
This triggers an update of the bucket bar, so that it reflects changes to the visibility of highlights.

**Testing:**

1. Go to http://localhost:9083/https://www.youtube.com/watch?v=TbzZIMQC6vk
2. Scroll down most of the way and make an annotation
3. Scroll back up to the top and select the first segment
4. Enter a query in the filter bar which matches quite a few segments, but not the one that was annotated (I used "test")

On `main`, the bucket bar indicator for the highlight will remain visible in step 4, even if the highlight is no longer visible. On this branch the bucket bar should update whenever the filter changes.

Step (3) in the above is important because when the video is not scrolled to the top, changing the filter will often cause the content to scroll, which will in turn update the bucket bar anyway.

### DIFF
--- a/via/static/scripts/video_player/components/VideoPlayerApp.tsx
+++ b/via/static/scripts/video_player/components/VideoPlayerApp.tsx
@@ -218,6 +218,12 @@ export default function VideoPlayerApp({
     pendingScrollToHighlight.current = false;
   }, [isFilterEmpty, syncTranscript]);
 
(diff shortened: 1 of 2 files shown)


acelaya (approved):
👍🏼

PR #1120, opened 2023-07-25
Use `duration` data from API to set `data-time-end` attribute for last segment

robertknight:
Use the `duration` property of segments to calculate the end time, rather than the `start` time of the next segment. This allows us to compute the `data-time-end` attribute value for the last segment.

The `duration` data from YouTube can be "too long" such that segments overlap (see comments for explanation). To deal with this, add a normalization step that clips duration values received from the API. We could do this normalization on the backend. For now it is just as easy to do it on the frontend since it is the only consumer.

**Testing:**

1. Open a video in Via (eg. http://localhost:9083/https://www.youtube.com/watch?v=TbzZIMQC6vk)
2. Inspect several of the `data-time-start` and `data-time-end` attributes for the `<p>` elements containing segment text, including the last segment. The segments should not overlap, the `data-time-start` values should match the displayed timestamps (but represented as seconds), and `data-time-end` should be set for the last segment.

### DIFF
--- a/via/static/scripts/test-util/video-player-fixtures.ts
+++ b/via/static/scripts/test-util/video-player-fixtures.ts
@@ -28,10 +28,12 @@ export const transcriptsAPIResponse = {
         {
           text: '[Music]',
           start: 0.0,
+          duration: 7.52,
         },
         {
           text: 'how many of you remember the first time',
           start: 5.6,
+          duration: 4.72,
         },
       ],
     },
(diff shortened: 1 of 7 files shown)


via/static/scripts/test-util/video-player-fixtures.ts
```diff
@@ -28,10 +28,12 @@ export const transcriptsAPIResponse = {
         {
           text: '[Music]',
           start: 0.0,
+          duration: 7.52,
         },
         {
           text: 'how many of you remember the first time',
           start: 5.6,
+          duration: 4.72,
```
robertknight, line 36: These are actual values from the API, taken from the `sample-transcript.ts` data. Note how they overlap.

acelaya (approved):
👍🏼

PR #1123, opened 2023-07-26
Move dev DB schema creation and stamping to a separate process

seanh:
This will make Via out of sync with the cookiecutter. There's a couple of follow-up cookiecutter PRs to fix this: https://github.com/hypothesis/cookiecutters/pull/146, https://github.com/hypothesis/cookiecutters/pull/147

Context
-------

At startup (when you run `make dev`) Via's Gunicorn processes attempt to
create the DB tables if they don't already exist. This is done in dev
only, not in production:

    # via/db.py

    def includeme(config):
        ...

        if config.registry.settings.get("dev"):
            # Create the database tables if they don't already exist.
            Base.metadata.create_all(engine)

Problem
-------

Via actually has four Gunicorn workers in dev (see the `workers = 4` in
`conf/gunicorn/dev.conf.py`). All 4 try to create the DB tables at the
same time and 3 of them crash, printing out error messages.

The issue is kind of difficult to produce:

1. You need to actually have a model, so you won't be able to produce
   the issue on the `db` branch without changes. See
   https://github.com/hypothesis/via/pull/1072 and
   https://github.com/hypothesis/via/pull/1093 for branches on which you
   can reproduce the issue.

2. The model's table needs to actually be missing from the DB table. You
   can do that by running this command:

       make services args=down && make services && make dev

3. Since the issue is a race condition you won't be able to reproduce it
   reliably, but if you run the above command several times you should
   see it.

Solution
--------

This commit moves the `sqlalchemy.create_all()` call out of the Gunicorn
worker process and instead adds a separate `make_db` process to the list
of processes that `supervisord-dev.conf` runs when you run `make dev`.
This new `make_db` process starts up, calls `create_all()`, and then
exits, and since it's a singular process there's no collision.

### DIFF
--- a/.cookiecutter/includes/conf/supervisord-dev.conf.json
+++ b/.cookiecutter/includes/conf/supervisord-dev.conf.json
@@ -1,5 +1,9 @@
 {
   "programs": {
+    "make_db": {
+      "command": "bin/make_db",
+      "startsecs": "0"
+    },
     "web": {
       "command": "gunicorn -c conf/gunicorn/dev.conf.py --paste conf/development.ini"
     },
--- a/bin/make_db
+++ b/bin/make_db
@@ -1,5 +1,35 @@
-"""Initialize the dev environment's DB."""
 #!/usr/bin/env python3
-import sys
+"""Initialize the dev environment's DB."""
+import alembic.command
+import alembic.config
 from pyramid.paster import bootstrap
-bootstrap("conf/development.ini")
+from sqlalchemy import text
+from sqlalchemy.exc import ProgrammingError
+
+from via.db import Base, create_engine
+
+
+def is_stamped(engine) -> bool:
+    """Return True if the DB is stamped with an Alembic revision ID."""
+    with engine.connect() as connection:
+        try:
+            if connection.execute(text("select * from alembic_version")).first():
+                return True
+        except ProgrammingError:
+            pass
+
+    return False
+
+
+def main():
+    with bootstrap("conf/development.ini") as env:
+        settings = env["registry"].settings
+        engine = create_engine(settings)
+
+        if not is_stamped(engine):
+            Base.metadata.create_all(engine)
+            alembic.command.stamp(alembic.config.Config("conf/alembic.ini"), "head")
+
+
+if __name__ == "__main__":
+    main()
--- a/via/db.py
+++ b/via/db.py
@@ -1,8 +1,6 @@
-import alembic.command
-import alembic.config
+import sqlalchemy
 import zope.sqlalchemy
-from sqlalchemy import MetaData, create_engine, text
-from sqlalchemy.exc import ProgrammingError
+from sqlalchemy import MetaData
 from sqlalchemy.orm import DeclarativeBase, MappedAsDataclass, Session
 
 
@@ -18,23 +16,13 @@ class Base(MappedAsDataclass, DeclarativeBase):
     )
 
 
-def stamp(engine):  # pragma: no cover
-    """If the DB isn't already stamped then stamp it with the latest revision."""
-    stamped = False
-
-    with engine.connect() as connection:
-        try:
-            if connection.execute(text("select * from alembic_version")).first():
-                stamped = True
-        except ProgrammingError:
-            pass
-
-    if not stamped:
-        alembic.command.stamp(alembic.config.Config("conf/alembic.ini"), "head")
+def create_engine(settings):  # pragma: no cover
+    """Return a SQLAlchemy engine."""
+    return sqlalchemy.create_engine(settings["database_url"])
 
 
 def includeme(config):  # pragma: no cover
-    engine = create_engine(config.registry.settings["database_url"])
+    engine = create_engine(config.registry.settings)
 
     def db_session(request):
         """Return the SQLAlchemy session for the given request."""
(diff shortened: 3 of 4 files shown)


bin/make_db
```diff
@@ -1,5 +1,12 @@
-"""Initialize the dev environment's DB."""
 #!/usr/bin/env python3
-import sys
+"""Initialize the dev environment's DB."""
```
seanh, line 2: The `#!` needs to be above the docstring or it doesn't work! :facepalm: 

`bin/make_db` actually comes from the cookiecutter, so I'll need to move these changes into the cookiecutter

bin/make_db
```diff
@@ -1,5 +1,12 @@
...
-bootstrap("conf/development.ini")
+
+from via.db import create_all, create_engine
+
+
+with bootstrap("conf/development.ini") as env:
+    settings = env["registry"].settings
+
+    engine = create_engine(settings)
+    create_all(engine)
```
seanh, line 12: Made a couple of changes to `bin/make_db` (which I'll have to move into the cookiecutter):

1. You're supposed to use `with bootstrap(...)`, see [Pyramid's docs](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/commandline.html#writing-a-script)
2. Previously `create_all()` and `stamp()` were [called while bootstrapping the web process](https://github.com/hypothesis/via/blob/35c123986311aa43f6f8e78f43cb31a8f36a4f34/via/db.py#L50-L53). This is why `create_all()` and `stamp()` would be called by the normal web processes when you ran `make dev`. It's also why `bin/make_db` could simply call `bootstrap()` and nothing else. Since I've now changed the normal web process bootstrapping to *not* call `create_all()` or `stamp()`, `bin/make_db` now needs to call them explicitly.

Of course this new code doesn't call `stamp()` yet, it only calls `create_all()`, so I've disabled auto-stamping here. I need to fix that.

Back-porting this into the cookiecutter might be tricky since `bin/make_db` is part of the cookiecutter but `db.py` (where `create_engine()`, `create_all()` and `stamp()` are) is not part of the cookiecutter.

via/db.py
```diff
@@ -33,8 +34,18 @@ def stamp(engine):  # pragma: no cover
...
 
 
+def create_engine(settings):
+    """Return a SQLAlchemy engine."""
+    return sqlalchemy.create_engine(settings["database_url"])
+
+
+def create_all(engine):
+    """Create all DB tables that don't already exist."""
+    Base.metadata.create_all(engine)
```
seanh: Pulled `create_all()` out into a separate process that's *not* called by the `includeme()` below anymore, and is imported and called by `bin/make_db` instead.

`create_engine()` also had to be pulled out because `make_db` needs to call it in order to get an engine to pass to `create_all(engine)`.

jon-betts: I think we could probably call "Base.metadata.create_all(engine)" directly in the script? Save ourselves a test. I'm not sure wrapping it here provides benefit.

via/db.py
```diff
@@ -47,11 +58,6 @@ def db_session(request):
 
         return session
 
-    if config.registry.settings.get("dev"):
-        # Create the database tables if they don't already exist.
-        Base.metadata.create_all(engine)
-        stamp(engine)
```
jon-betts, line 53: Honestly not having dev specific stuff in here seems like a win to me.

seanh: :+1:

via/db.py
seanh: I've inlined the `create_all()` and also the stamping, but left `create_engine()` in `db.py` because it's actually called in two places

.cookiecutter/includes/conf/supervisord-dev.conf.json
```diff
@@ -1,5 +1,9 @@
 {
   "programs": {
+    "make_db": {
+      "command": "bin/make_db",
+      "startsecs": "0"
```
seanh, line 5: By default if `bin/make_db` exits in under a second (which it easily can do) Supervisor will consider this a crash regardless of the exit code and will restart the process. Setting `startsecs` to 0 tells Supervisor not to consider this a crash and not to restart the process (unless the exit code is non-zero)

bin/make_db
```diff
@@ -1,5 +1,35 @@
...
+            if connection.execute(text("select * from alembic_version")).first():
+                return True
+        except ProgrammingError:
+            pass
+
+    return False
+
+
+def main():
+    with bootstrap("conf/development.ini") as env:
```
seanh, line 25: `with` is the recommended way to use `bootstrap()` in the Pyramid docs, `make_db` wasn't doing this before

bin/make_db
```diff
@@ -1,5 +1,35 @@
...
+        settings = env["registry"].settings
+        engine = create_engine(settings)
+
+        if not is_stamped(engine):
+            Base.metadata.create_all(engine)
+            alembic.command.stamp(alembic.config.Config("conf/alembic.ini"), "head")
```
marcospri, line 31: As I mentioned on the other PR I reckon either here or in the Makefile
we should run the migrations as part of the `make db` process 

https://github.com/hypothesis/lms/blob/main/Makefile#L18

I reckon with that we'll get a "it just works" experience for dev that are not making changes here often and just want to have the service running.

marcospri (approved):
I've seen Jon already took a look here.

Made a comment about the migration in the `make db` process. I reckon that's the right solution but if you prefer we can chat about that once we have actual migrations in this project.

bin/make_db
seanh: Via [does run the migrations in `Makefile`](https://github.com/hypothesis/via/blob/c97bceaf640c91493f394ab0340e6c6c9cc68f32/Makefile#L18). I think `make db` has always done this for all our apps?

It may be a bit confusing that there's a `bin/make_db` script, but calling that script is only one of two commands that `make db` runs. May want to move the `alembic uprade` command into the `make_db` script somehow

PR #1125, opened 2023-07-27
Add logic to handle toast messages (TransitionComponent)

### LINKED ISSUE #1132: Add a toast message component for use in video-annotations application

acelaya:
This PR takes the best ideas from https://github.com/hypothesis/via/pull/1089, https://github.com/hypothesis/via/pull/1094 and https://github.com/hypothesis/via/pull/1095, plus [some discussions](https://hypothes-is.slack.com/archives/C1M8NH76X/p1689935097734789) that have happened afterwards, and implements the logic to handle toast messages for the video app, in a way that we can eventually extract the generic components into frontend-shared and reuse them in other apps, like client.

The main highlights are:

* The concept of a controlled `ToastMessages` component is still there. This is the main piece that we could potentially extract.
  This time, we get toast message transitions back, but wrapped in a reusable `TransitionComponent`.
  > This transition component is named `BaseToastMessageTransition`, because it wraps a couple of animations, so I couldn't find a consistent name that better described it. That's probably fine for now, as it's an internal component, but if we ever extract it, it should be renamed.
* State management has been simplified. There's no use of context yet, just a simple `useToastMessages` hook which wraps `useState`.
  This hook exposes two more helper functions to append and dismiss messages, and these function + the `toastMessages` state is propagated from `VideoPlayerApp` to the only two components that currently need access to it, being the main one of them, the `ToastMessages` component described above.

### Todo

- [x] Add tests
- [x] ~Try to handle autodismiss also via CSS animation, instead of setTimeout~. This has many implications, as it's not easy to fit this concept in current `TransitionComponent` API. Also, telling which messages have been dismissed is a responsibility of the parent component. We can revisit this in the future.

### Testing steps

*You can follow the testing steps described in https://github.com/hypothesis/via/pull/1095*

> This PR closes https://github.com/hypothesis/via/issues/1132


via/static/scripts/video_player/components/CopyButton.tsx
```diff
@@ -1,25 +1,32 @@
...
       : '';
     try {
       await navigator.clipboard.writeText(formattedTranscript);
     } catch (err) {
-      // TODO: Replace this with a toast message in the UI.
-      console.warn('Failed to copy transcript', err);
+      appendToastMessage({
+        type: 'error',
+        message: 'Failed to copy transcript',
+      });
```
acelaya, line 33: I would reconsider displaying a toast message when the copy succeeds. Currently it feels like nothing happened.

robertknight: I agree, especially since the button itself doesn't provide any kind of visual information. A green "Transcript copied" message would be fine.

via/static/scripts/video_player/hooks/use-toast-messages.ts
```diff
@@ -0,0 +1,28 @@
...
+      setToastMessages(messages =>
+        messages.filter(message => message.id !== id)
+      ),
+    []
+  );
+
+  return { toastMessages, appendToastMessage, dismissToastMessage };
+}
+
+export type ToastMessageAppender = ReturnType<
```
robertknight: I think it would aid readability to create an explicit type for the return value of `useToastMessages`.

via/static/scripts/video_player/utils/generate-hex-string.ts
```diff
@@ -0,0 +1,15 @@
+function byteToHex(val: number): string {
```
robertknight: Strictly speaking generating secure hex strings with `window.crypto.getRandomValues` is kinda overkill for this purpose. A sequential counter would also be fine.

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -0,0 +1,187 @@
...
+        return;
+      }
+
+      // Track that this message has been scheduled to be dismissed. After a
+      // period of time, actually dismiss it
+      scheduledMessages.current.add(id);
+      _setTimeout(() => {
+        dismissMessage(id);
+        scheduledMessages.current.delete(id);
+      }, 5000);
```
acelaya, line 133: I have been struggling in the tests to mock calls to `setTimeout`. I have tried using `sinon.useFakeTimers()` (like in `VideoPlayerApp-test`) as well as `sinon.stub(window, 'setTimeout').callsArg(0)`, and for some reason the callback was never invoked.

I have finally decided that having a test seam prop was simple enough.

acelaya:
@robertknight this is now ready for review.

Things that have been added/changed since your first review as draft:

* Added a `success` toast message when copying the transcript to the clipboard succeeds.
* Added tests to cover new code (`useToastMessages`, `ToastMessages`, etc).

Considerations for follow-up PRs:

* The `ToastMessages` test does a decent job on mocking timeouts, but it still needs to wait for CSS transitions, making the test a bit slow. We may need to consider having a different tailwind config for tests, where we can set shorter animations (10ms instead of 300ms, for example).
* As opposed to the work we did on the `Dialog` component, that can work with or without transitions, toast messages rely on the TransitionComponent wrapping them to be properly dismissed. I wouldn't worry too much for now, but the work done [here](https://github.com/hypothesis/frontend-shared/pull/1131) would be useful if we want transitions to be optional at some point.
* Animations in the sidebar can have a fade-in or slide-from-right "in" transition depending on the size of the frame.
  In this case I did it so that it's visually equivalent, but it is the opposite in terms of media queries, because the sidebar frame is usually narrow, while the video app host frame is big. I need to align this with Lyza and see how to address it before we eventually extract it to frontend-shared.

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -0,0 +1,190 @@
...
+const BaseToastMessageTransition: TransitionComponent = ({
+  direction = 'in',
+  onTransitionEnd,
+  children,
+}) => {
+  const isDismissed = direction === 'out';
+  const containerRef = useRef<HTMLDivElement>(null);
+  const handleAnimation = (e: AnimationEvent) => {
+    // Ignore non-relevant animations:
+    // * Happening on children elements
```
robertknight:
```suggestion
    // * Happening on child elements
```

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -0,0 +1,190 @@
...
+    >
+      {children}
+    </div>
+  );
+};
+
+export type ToastMessagesProps = {
+  messages: ToastMessage[];
+  onMessageDismiss: (id: string) => void;
+  _setTimeout?: typeof setTimeout;
```
robertknight:
```suggestion
  setTimeout_?: typeof setTimeout;
```

It is more conventional in our code to put the underscore for this kind of seam at the end of the name rather than the beginning (see `window_` and `document_` in the client).

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -0,0 +1,190 @@
...
+        onMessageDismiss(message.id);
+        setDismissedMessages(ids => ids.filter(id => id !== message.id));
+      }
+    },
+    [scheduleMessageDismiss, onMessageDismiss]
+  );
+
+  // The `ul` containing any toast messages is absolute-positioned and the full
+  // width of the viewport. Each toast message `li` has its position and width
+  // constrained by `container` configuration in tailwind.
```
robertknight: This comment looks out of date to me. The `<ul>` is the full width of its _container_, but is not absolute positioned, and the sizing is relative to the container rather than the viewport.

via/static/scripts/video_player/components/ToastMessages.tsx
robertknight: A test seam is fine with me.

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * A collection of toast messages. These are rendered within an `aria-live`
+ * region for accessibility with screen readers.
+ */
+export default function ToastMessages({
+  messages,
+  onMessageDismiss,
+  /* istanbul ignore next - test seam */
+  _setTimeout = setTimeout,
```
robertknight: See comment about naming above.

via/static/scripts/video_player/components/test/ToastMessages-test.js
```diff
@@ -0,0 +1,109 @@
...
+  toastMessages.forEach((message, index) => {
+    it('dismisses messages when clicked', async () => {
+      const wrapper = createToastMessages(toastMessages);
+
+      wrapper.find('Callout').at(index).props().onClick();
+      // onMessageDismiss is not immediately called. Transition has to finish
+      assert.notCalled(fakeOnMessageDismiss);
+
+      // Once dismiss transition has finished, onMessageDismiss is called
+      await delay(350); // "out" transition lasts 300ms
```
robertknight: We definitely need to find a way to make this run faster in the tests. Also this could cause test flakes if the animation runs more slowly than expected for any reason.

You mentioned using a custom Tailwind config as one option. That would be high fidelity, but could turn out to be brittle. Some other ideas:

- Dispatch fake `animationend` events at the DOM elements
- Add a test seam to allow animation ends to be triggered sooner another way

via/static/scripts/video_player/components/test/ToastMessages-test.js
```diff
@@ -0,0 +1,109 @@
...
+      [
+        ...toastMessages,
+        {
+          id: 'foo',
+          type: 'success',
+          message: 'Not to be dismissed',
+          autoDismiss: false,
+        },
+      ],
+      // Fake internal setTimeout, to immediately call its callback
```
robertknight, line 83: As a rule of thumb, I would recommend that fakes of async events should still be async, so as to better reflect the real flow of execution. Changing a fake timer from 100ms to 0ms makes the test run faster without affecting the overall flow of events.

robertknight (commented):
We definitely need to avoid waiting for a 300ms delay for each animation in the tests. I suggested some possible approaches in addition to the one you suggested. Otherwise this looks good.

via/static/scripts/video_player/components/test/ToastMessages-test.js
acelaya: Thanks for the suggestions. That didn't occur to me, and I think it should be possible to get something workable with that.

via/static/scripts/video_player/components/ToastMessages.tsx
acelaya: Oh, right. I think I got it mixed with the prefix for private props 😅

via/static/scripts/video_player/components/ToastMessages.tsx
acelaya: Oh, right. I copied this from `client` like a million years ago, and completely forgot about it 😅

via/static/scripts/video_player/components/test/ToastMessages-test.js
acelaya: I'm not sure if I understand this 🤔 

Were you just making a general comment, or are you suggesting some change?

via/static/scripts/video_player/utils/generate-hex-string.ts
acelaya: I copied this logic from `client` (this is how we generate toast message IDs there), but I do agree is probably overkill.

Perhaps the `useToastMessages` hook could globally keep track of that sequential counter you suggest.

via/static/scripts/video_player/components/test/ToastMessages-test.js
robertknight: This is just a general comment. I don't think it is critical to act on it here (but if you want to, that's fine too).

robertknight:
The GitHub UI is not reflecting the latest changes on this branch, so I looked at the diff locally and made notes:

1. I realized in re-reviewing this that dismissal is handled differently than I originally imagined. I was thinking that dismissing a message would immediately invoke `onMessageDismiss`, but keep a copy of the dismissed message internally for rendering during the dismissal animation. The approach in the current design of deferring the `onMessageDismiss` callback works here, though we might find we need to revisit this in future. For example if some other UI update needed to happen as soon as the user clicked on a toast message to dismiss it.
2. In `useToastMessages` tests, assert arguments order should always be `assert.equal(actual, expected)` - think "assert that what I got is what I expected". This also applies to every test framework I can think of.
3. In `schedules dismiss only once per message` test, you can change the `ms` argument in `setTimeout(callback, ms)` and `await delay(ms)` to zero without affecting the semantics of the test. All that matters here is that the maximum value passed to `setTimeout` is <= the max value passed to `delay`. This is because a series of timeouts scheduled with the same delay value [are guaranteed to execute in order](https://hyp.is/bUUSzC-VEe6j-RdXJdRVrw/html.spec.whatwg.org/multipage/timers-and-user-prompts.html).
4. In "the invokes `onTransitionEnd for proper animation" test change "proper" to "supported".

acelaya:
Thanks Rob!

There's just one point of feedback from me on regards to this.

> 1. I realized in re-reviewing this that dismissal is handled differently than I originally imagined. I was thinking that dismissing a message would immediately invoke `onMessageDismiss`, but keep a copy of the dismissed message internally for rendering during the dismissal animation. The approach in the current design of deferring the `onMessageDismiss` callback works here, though we might find we need to revisit this in future. For example if some other UI update needed to happen as soon as the user clicked on a toast message to dismiss it.

In my mental model, I think it makes more sense that the message is effectively removed from the state once the animation/transition has finished. The opposite would not be very intuitive for me.

Also, doing otherwise means the messages "displayed" can eventually not map 1 to 1 with the messages in the store, because messages getting dismissed are removed from the store immediately, and there's a short period of time in which they are still "displayed".

Ultimately, it would force to handle some internal state derived from props, making it slightly more complex to handle.

But I do agree there are grey areas and see your point, so I'm open to revisit it later.

As per the rest of your points, I'll address them.

acelaya:
> In useToastMessages tests, assert arguments order should always be assert.equal(actual, expected) - think "assert that what I got is what I expected". This also applies to every test framework I can think of.

I think I always get this the wrong way because of the `expect(expected).toEqual(actual)` form, from jest and jest-like frameworks 😅

robertknight:
> I think I always get this the wrong way because of the expect(expected).toEqual(actual) form, from jest and jest-like frameworks

But that's not the Jest syntax! It is `expect(actual).toEqual(expected)`. See https://jestjs.io/docs/expect#expectvalue:

> The argument to `expect` should be the value that your code produces, and any argument to the matcher should be the correct value. If you mix them up, your tests will still work, but the error messages on failing tests will look strange.

acelaya:
> But that's not the Jest syntax! It is `expect(actual).toEqual(expected)`. See https://jestjs.io/docs/expect#expectvalue:

OMG! All tests I have ever written are wrong 😱

robertknight (approved):
Github is still displaying an old version of this branch in the diff viewer, so I checked it locally manually.

PR #1126, opened 2023-07-27
Add fallback method for highlighting filter matches

### LINKED ISSUE #1075: Support highlighting search matches in browsers that don't support CSS Custom Highlight API
Filtering the transcript in the video player uses the [CSS Custom Highlight API](https://github.com/hypothesis/via/pull/929) to highlight the matching terms in the transcript:
 
Search term highlight

This currently only works in Chrome. In Safari and Firefox, the user just sees the matching transcript segments. It would be good to have a fallback implementation for browsers that don't support this API.

The reason we use CSS Custom Highlight API rather than modifying the DOM (eg. to wrap the matches in `<mark>` elements) as it avoids interfering with any highlights created by the Hypothesis client. Potentially this also allows us to benefit from integration this API has with assistive technology in future.

Some potential solutions:

1. Find a generic solution to allow Hypothesis client highlights to persist even if the DOM structure in/around the content is changed. This would allow us to insert highlights by modifying the DOM.
2. Draw highlights behind or on top of the transcript using other methods. Since these highlights are not adding new information (the search term is available from the filter bar) I think it is acceptable if they are not visible to assistive technology.

### LINKED ISSUE #1122: Filter matches may not be highlighted if transcript segment is highlighted
The highlighting of filter matches in the video player can fail if the segment contains highlights.

In the screenshot below, the text "quarks" should be highlighted in the first segment, but isn't because it occurs after a highlight.

Search highlights

The issue occurs because the highlighting code currently only works when the match can be found in the first text node child of the `<p>` element.

robertknight:
Add a fallback method for highlighting filter matches in browsers that don't support CSS custom highlights. This involves wrapping matches in `<mark>` elements, similar to how the Hypothesis client inserts highlights. We continue to use CSS custom highlights where available, since it has the potential to be faster due to not modifying the DOM tree.

In the process an issue was fixed where filter matches would not be highlighted if they occurred after highlights inserted by the Hypothesis client.

Fixes https://github.com/hypothesis/via/issues/1122
Fixes https://github.com/hypothesis/via/issues/1075

----

**Testing:**

1. Load a video in Via
2. Type in the search bar. You should see matches for the filter underlined in the text
3. Repeat (2) in Chrome, Firefox and Safari. The result should look the same in each browser.
4. Highlight part of a segment and try searching for matches that occur after the Hypothesis highlight. Matches should be underlined whether they occur before, within or after Hypothesis highlights.

### DIFF
--- a/via/static/scripts/types/css-custom-highlights.d.ts
+++ b/via/static/scripts/types/css-custom-highlights.d.ts
@@ -10,6 +10,7 @@ declare class Highlight {
 
 interface HighlightsRegistry {
   delete(name: string): void;
+  get(name: string): Highlight | undefined;
   set(name: string, highlight: Highlight): void;
 }
 
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -10,6 +10,7 @@ import {
 } from 'preact/hooks';
 
 import { useScrollAnchor } from '../hooks/use-scroll-anchor';
+import { TextHighlighter } from '../utils/highlighter';
 import { formatTimestamp } from '../utils/time';
 import type { MatchOffset, Segment, TranscriptData } from '../utils/transcript';
 import { filterTranscript } from '../utils/transcript';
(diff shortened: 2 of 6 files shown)


acelaya (approved):
👍🏼

Issue #1029, opened 2023-06-30
Optimize re-rendering of transcripts for very long videos


robertknight:
Typing in the filter bar in the video player and other parts of the UI can get janky for very long videos (eg. multiple hours of conversation).

There are some basic optimizations we could implement to improve this, such as memoizing `TranscriptSegment` and `Transcript`, taking care not to re-render them when only event handlers have changed.

Example video: https://youtu.be/diJp4zoQPqo (also many others on this channel)

robertknight:
This is now done. In the process I did notice that filtering in very long transcripts is laggy in Safari / Firefox compared to Chrome. This is because Safari and Chrome require DOM manipulation to insert highlights, whereas Chrome can use the more efficient CSS Custom Highlights. That is separate from this issue.

PR #1129, opened 2023-07-28
Add `useStableCallback` hook

robertknight:
We have seen several use cases for creating a callback which always uses the latest props and state, but has a stable identity so that it doesn't cause effects to re-run or memo-d values to be invalidated when the closed-over values change.

Add a `useStableCallback` hook for this purpose, and use it to simplify code in several components. This hook is similar to the unstable `useEffectEvent` [1] React hook, which was at one point a more general `useEvent` [2].

As well as simplifying some existing code, I am intending to use this in https://github.com/hypothesis/via/issues/1029.

[1] https://react.dev/reference/react/experimental_useEffectEvent
[2] https://blog.logrocket.com/what-you-need-know-react-useevent-hook-rfc/

### DIFF
--- a/via/static/scripts/video_player/hooks/use-stable-callback.ts
+++ b/via/static/scripts/video_player/hooks/use-stable-callback.ts
@@ -1,0 +1,23 @@
+import { useRef } from 'preact/hooks';
+
+/**
+ * Return a function which wraps a callback to give it a stable value.
+ *
+ * The wrapper has a stable value across renders, but always forwards to the
+ * callback from the most recent render. This is useful if you want to use a
+ * callback inside a `useEffect` or `useMemo` hook without re-running the effect
+ * or re-computing the memoed value when the callback changes.
+ */
+export function useStableCallback<R, A extends any[], F extends (...a: A) => R>(
+  callback: F
+): F {
+  const wrapper = useRef({
+    callback,
+    call: (...args: A) => wrapper.current.callback(...args),
+  });
+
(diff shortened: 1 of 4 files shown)


via/static/scripts/video_player/hooks/use-stable-callback.ts
```diff
@@ -0,0 +1,23 @@
...
+): F {
+  const wrapper = useRef({
+    callback,
+    call: (...args: A) => wrapper.current.callback(...args),
+  });
+
+  // On each render, save the last callback value.
+  wrapper.current.callback = callback;
+
+  return wrapper.current.call as F;
```
acelaya, line 22: So, if I'm getting this right, with this approach you make sure the reference to `call` never changes (hence, it's stable when it comes to side effects), while it forwards calls to `callback`, which is the one changing on every render to make sure it's up to date.

Is that correct?

acelaya (approved):
I think this is a clever approach.

Perhaps we can consider moving this hook to frontend-shared, if we find use cases for it in other apps.

via/static/scripts/video_player/hooks/use-stable-callback.ts
robertknight: That's correct.

robertknight:
> Perhaps we can consider moving this hook to frontend-shared, if we find use cases for it in other apps.

Yes, indeed.

PR #1130, opened 2023-07-28
Memo the transcript segment list and segment contents

robertknight:
Reduce re-rendering of the segment list and individual segments with `useMemo` and `useStableCallback`.

This improves responsiveness when typing in the filter bar in very long videos in Chrome (see https://github.com/hypothesis/via/issues/1029 for example video). In Safari and Firefox the improvements here still apply, but are overshadowed by the much slower fallback highlighting process in `TextHighlighter`, which needs to be improved separately.

Use the option to [ignore whitespace changes](https://github.com/hypothesis/via/pull/1130/files?diff=unified&w=1) in the diff when reviewing this.

Part of https://github.com/hypothesis/via/issues/1029.


acelaya (approved):
👍🏼

PR #1131, opened 2023-07-31
Center "X" button using top + translate in filter input

### LINKED ISSUE #1124: Clear filter button is misplaced in mobile device resolutions
The "X" button on the right edge of the filter bar is not vertically centered within the input field on touch-input devices:

image

To reproduce, enable the mobile device mode in Chrome DevTools.

robertknight:
`IconButton`s change their size depending on whether mouse vs touch input is being used. Change the positioning of the "X" button so that it remains centered in both cases.

Fixes https://github.com/hypothesis/via/issues/1124


lyzadanger (approved):
Does the trick 👍🏻

PR #1137, opened 2023-07-31
Add migration to add video table

seanh:


via/migrations/versions/d4e3e1bf95eb_add_the_video_table.py
```diff
@@ -0,0 +1,32 @@
...
+def upgrade() -> None:
+    op.create_table(
+        "video",
+        sa.Column(
+            "created", sa.DateTime(), server_default=sa.text("now()"), nullable=False
+        ),
+        sa.Column(
+            "updated", sa.DateTime(), server_default=sa.text("now()"), nullable=False
+        ),
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
```
marcospri: Same as with the other migration, we could move this to the top to reflect the mixins.

seanh: :+1: Done

marcospri (approved):

PR #1153, opened 2023-08-01
Upgrade to Yarn v3

robertknight:
Follow the steps at https://yarnpkg.com/getting-started/migration to upgrade to Yarn v3:

 1. `yarn set version berry`
 2. Edit `.yarnrc.yml` to add `nodeLinker: node-modules`.
 3. Run `yarn install` to migrate lockfile
 4. Update `.gitignore` (see comment)

This follows similar changes in the client repo [1], with the addition of changes to the Docker image build:

 1. Revert Node version to v20.2.x to work around the issue reported at https://github.com/vercel/turbo/issues/5335. This issue will apparently be fixed in v20.4.x.
 2. Run `yarn install` in a subdirectory of `/` to avoid an error when running in the root (see https://github.com/yarnpkg/berry/issues/5380)
 3. Copy the new files required by Yarn v3 for the build into the container

Testing:

I built the docker image using `make docker` and tested it using `make docker-run`. This required some changes described in https://github.com/hypothesis/via/pull/1153#issuecomment-1661708721.

[1] https://github.com/hypothesis/client/commit/044f6b82cbdbebff9cdfa1dc4a27914ca4aab2c7


robertknight:
Notes on testing the Docker image locally, using the video player app. I had to:

- Update `.docker.env` with env var values from `tox.ini`, for `API_JWT_SECRET`, `DATABASE_URL` and `YOUTUBE_TRANSCRIPTS`
- Edit `.devdata.env` to remove quotes around the `YOUTUBE_API_KEY` env var value

acelaya (approved):
LGTM. Let's give it a try.

Dockerfile
```diff
@@ -1,11 +1,14 @@
 # Stage 1: Build frontend assets
-FROM node:20.3.1-alpine as frontend-build
+FROM node:20.2.0-alpine as frontend-build
 
 ENV NODE_ENV production
-COPY .babelrc rollup.config.mjs gulpfile.mjs package.json tailwind.config.mjs yarn.lock ./
-COPY via/static ./via/static
+RUN mkdir -p /tmp/frontend-build
+COPY .babelrc rollup.config.mjs gulpfile.mjs package.json tailwind.config.mjs .yarnrc.yml yarn.lock /tmp/frontend-build
```
acelaya, line 6: I just noticed this after after my first review.

~Does it make sense that this `COPY` has `/tmp/frontend-build`? Isn't that something we are creating inside docker itself?~ Nothing to see here 😛

acelaya: Ugh, that's the destination. Never mind 😅

PR #1159, opened 2023-08-03
Return the first transcript we find, instead of searching only for "en"

jon-betts:
This paves the way for storing non "en" transcripts in the DB, should we find them. This includes automatically generated, other English dialects like "en-gb" and named English transcripts.

## Testing notes

 * `make service dev`
 * Visit: http://localhost:9083/https://www.youtube.com/watch?v=hzjX4Eldh84

### DIFF
--- a/tests/factories/transcript.py
+++ b/tests/factories/transcript.py
@@ -10,20 +10,6 @@ class TranscriptFactory(SQLAlchemyModelFactory):
 
     video_id = Sequence(lambda n: f"video_id_{n}")
     transcript_id = Sequence(lambda n: f"transcript_id_{n}")
-    transcript = [
-        {
-            "text": "[Music]",
-            "start": 0.0,
-            "duration": 7.52,
-        },
-        {
-            "text": "how many of you remember the first time",
-            "start": 5.6,
-            "duration": 4.72,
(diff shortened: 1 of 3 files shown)


seanh (approved):

tests/factories/transcript.py
```diff
@@ -10,20 +10,6 @@ class Meta:
...
-        },
-        {
-            "text": "you saw a playstation 1 game if you were",
-            "start": 7.52,
-            "duration": 4.72,
-        },
-    ]
+    transcript = Sequence(
+        lambda n: [{"text": "[Music]", "start": float(n), "duration": float(n + 1)}]
+    )
```
jon-betts, line 15: As we only get the transcript text value back, not the transcript object, having different ones allows us to compare them and tell the difference.

seanh: :+1:

This always generates lists of only one dict, do we want to make it generate a few or maybe a random number of dicts? Could also replace `"[Music]"` with lorem text from `faker`

seanh (approved):

seanh:
Closing in favour of https://github.com/hypothesis/via/pull/1163/

PR #1160, opened 2023-08-03
Integrate our own YouTube code and lookup algorithm

jon-betts:
Requires:

 * https://github.com/hypothesis/via/pull/1159
 * https://github.com/hypothesis/via/pull/1100

Note: No handling of HTTPService errors at the moment

PR #1162, opened 2023-08-15
Replace the third-party `youtube-transcript-api` library with our own code

### LINKED ISSUE #1110: Fix Via's default YouTube transcript selection
### Problem

The third-party library (<https://github.com/jdepoix/youtube-transcript-api>) that Via uses to select the transcript to use for a given YouTube video (and also to then download the selected transcript) does a poor job of transcript selection when the video has multiple English transcripts. The library's code incorrectly assumes that a YouTube video can only have one transcript for a given language code (e.g. `en`), but in fact videos can have more than one transcript with the same language code. See <https://github.com/jdepoix/youtube-transcript-api/issues/150>.

This results in poor transcript selection behaviour as the library's faulty code does something accidental that ends up picking one of the transcripts. For example this video (https://www.youtube.com/watch?v=rSTqpRYzJbo) has four English transcripts:

1. English
2. English - CC1
3. English - DTVCC1
4. English (United States)

The best transcript to choose by default is clearly either **English** or **English (United States)**. YouTube's own UI chooses **English (United States)** by default. But Via currently chooses **English - DTVCC1** for this video.

We'd like to replace this transcript selection algorithm with one that we can design ourselves, for example so that we can make it prefer the **English (United States)** transcript over **English - CC1** and **English - DTVCC1**.

In addition, <https://github.com/jdepoix/youtube-transcript-api> looks to be a poor-quality dependency so we'd like to replace it with our own code that will be easier for us to maintain. <https://github.com/jdepoix/youtube-transcript-api> also scrapes the HTML of YouTube pages in order to discover a JSON blob that contains the list of transcripts, whereas we think we can do this better by calling a (legacy, undocumented) YouTube/Google API to get this same JSON.

### Solution

We'd like to replace the body of [Via's `get_transcript()` method](https://github.com/hypothesis/via/blob/6c6662a2b1d47ec6e0c4a31fee7c4f80b9d4748f/via/services/youtube.py#L78-L85) with one that uses our own code to get the list of transcripts for a video, select a default transcript to use, and download that transcript. This should allow us to remove <https://github.com/jdepoix/youtube-transcript-api> from Via's Python requirements (dependencies).

### Scope

This issue is just to replace the body of [Via's `get_transcript()` method](https://github.com/hypothesis/via/blob/6c6662a2b1d47ec6e0c4a31fee7c4f80b9d4748f/via/services/youtube.py#L78-L85) with one that uses our own code instead of <https://github.com/jdepoix/youtube-transcript-api> to select and return a default English transcript, without making any other changes.

Later on we're going to need other things such as a method for returning the list of transcripts, and possibly extracting our YouTube API code into a library so that both Via and LMS can use it. LMS will also need a way of persisting users transcript selections in its DB and of communicating transcript selections from LMS to Via. These other changes are **out of scope** for this issue.

seanh:
Fixes https://github.com/hypothesis/via/issues/1110.

This is a rewrite of https://github.com/hypothesis/via/pull/1107, https://github.com/hypothesis/via/pull/1100 and https://github.com/hypothesis/via/pull/1160, making the code more minimal and straightforward and combining them into one PR.

This is a large PR in terms of line count but it does do "one thing": it replaces the third-party `youtube-transcript-api` library with our own `YouTubeTranscriptService` code. See https://github.com/hypothesis/via/issues/1110 for the reasons why we're doing this. There's a follow-up PR to actually remove `youtube-transcript-api` from the requirements files (https://github.com/hypothesis/via/pull/1164).

The main new things this adds are:

1. Code for getting the list of transcripts for a YouTube video: `YouTubeTranscriptService.get_transcript_infos()`. The `youtube-transcript-api` library did this by scraping the video's HTML page. Our own code does it by calling a deprecated V1 API instead.

2. Code for choosing the default transcript (i.e. the transcript that we will use) for a video: `YouTubeTranscriptService.pick_default_transcript()`

3. Code for actually downloading the text of a transcript from the undocumented/private YouTube XML library: `YouTubeTranscriptService.get_transcript()`

4. Code for generating unique IDs for YouTube transcripts that're unique even when the video has more than one transcript with the same language code. These custom unique IDs are used for the `transcript.transcript_id` column in the DB. This code for generating thse is in the new `TranscriptInfo.id()` method.

   (Technically (4) could be split into a separate follow-up PR but it wouldn't reduce the size of this PR by much: the ID-generation code and tests are small.)

# Testing

Just visit some YouTube videos in Via and look at the transcripts that you get in the frontend. For example:

* A video with four different English transcripts: <http://localhost:9083/https://www.youtube.com/watch?v=rSTqpRYzJbo>. This branch will pick the "English" transcript for this video, whereas the code on `main` would pick the `"English - DTVCC1"` transcript (which is not what we want). We might want to tweak our code to prefer `"English (United States)"`?
* A video with a single English auto-generated transcript: <http://localhost:9083/https://www.youtube.com/watch?v=EU6TDnV5osM>
* A video with multiple transcripts including English, English auto-generated, and some other languages: <http://localhost:9083/https://www.youtube.com/watch?v=qQ6a0iOzyHE>
* A video with no transcript: <http://localhost:9083/https://www.youtube.com/watch?v=rw_scW4D3ik> (this crashes and results in an "Unable to load transcript" error in the frontend, we might want to send a follow-up PR to improve the error handling here)

After visiting a bunch of YouTube videos you should see transcripts cached in the DB with our custom generated unique IDs:

```python
$ make shell
>>> db.scalars(select(models.Transcript)).all()
[Transcript(id=2, video_id='EU6TDnV5osM', transcript_id='en.a.RW5nbGlzaCAoYXV0by1nZW5lcmF0ZWQp'),
 Transcript(id=3, video_id='qQ6a0iOzyHE', transcript_id='en..RW5nbGlzaA=='),
 Transcript(id=4, video_id='rSTqpRYzJbo', transcript_id='en..RW5nbGlzaA==')]
```


tests/factories/transcript_info.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class TranscriptInfoFactory(Factory):
+    class Meta:
+        model = TranscriptInfo
+
+    language_code = "en-us"
+    name = "English (United States)"
+    url = Sequence(lambda n: f"https://example.com/api/timedtext?v={n}")
+    autogenerated = False
```
seanh, line 13: This factory doesn't do much (it's mostly hard-coded values) but it's still useful because it lets you create a `TranscriptInfo` while only specifying the values that you care about:

```python
TranscriptInfoFactory(name="Foo")
```

Instead of always having to do:

```python
TranscriptInfo(language_code="en", name="Foo", url="https://...", autogenerated=False)
```

It also gives each `TranscriptInfo` a different `url` so that different `TranscriptInfo`'s won't compare equal.

tests/unit/via/services/youtube_test.py
```diff
@@ -94,38 +95,45 @@ def test_get_video_title_raises_YouTubeDataAPIError(self, svc, http_service):
...
                 {
                     "video_id": "test_video_id",
-                    "transcript": YouTubeTranscriptApi.get_transcript.return_value,
+                    "transcript_id": transcript_info.id,
+                    "transcript": "test_transcript",
                 }
             )
         ]
 
-    @pytest.mark.usefixtures("db_session")
```
seanh, line 119: This isn't necessary here: the `svc` fixture depends on `db_session`.

tests/unit/via/services/youtube_test.py
```diff
@@ -94,38 +95,45 @@ def test_get_video_title_raises_YouTubeDataAPIError(self, svc, http_service):
...
-        self, transcript, svc, YouTubeTranscriptApi
+        self, svc, transcript, youtube_transcript_service
     ):
         returned_transcript = svc.get_transcript(transcript.video_id)
 
-        YouTubeTranscriptApi.get_transcript.assert_not_called()
+        youtube_transcript_service.get_transcript.assert_not_called()
         assert returned_transcript == transcript.transcript
 
-    @pytest.mark.usefixtures("db_session")
```
seanh, line 128: Again no necessary here: the `svc` fixture depends on `db_session`

via/services/youtube.py
```diff
@@ -101,10 +107,18 @@ def get_transcript(self, video_id):
...
+        transcript_info = self._transcript_svc.pick_default_transcript(transcript_infos)
+        transcript = self._transcript_svc.get_transcript(transcript_info)
+
         self._db.add(
-            Transcript(video_id=video_id, transcript_id="en", transcript=transcript)
+            Transcript(
+                video_id=video_id,
+                transcript_id=transcript_info.id,
+                transcript=transcript,
+            )
```
seanh, line 119: Note the `transcript_id=transcript_info.id` instead of `"en"`

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,135 @@
...
+        name = b64encode(self.name.encode("utf-8")).decode("utf-8")
+
+        return ".".join(
+            part or ""
+            for part in [
+                self.language_code,
+                "a" if self.autogenerated else None,
+                name,
+            ]
+        ).rstrip(".")
```
seanh, line 29: If we ever need to we can actually re-compute the values of `TranscriptInfo.language_code`, `TranscriptInfo.name` and `TranscriptInfo.autogenerated` from just the value of `TranscriptInfo.id` (which is the only thing that gets saved to the DB). We wouldn't be able to re-compute the download `url` from the `id`.

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,135 @@
...
+        "language_code": re.compile("en"),
+        "name": re.compile(".*"),
+        "autogenerated": True,
+    },
+    {
+        "language_code": re.compile("en-"),
+        "name": re.compile(".*"),
+        "autogenerated": True,
+    },
+)
```
seanh, line 70: This ordered list of preferences define how we choose the default transcript when a video has more than one transcript. For example we'll first look for a human-generated transcript with `language_code` `"en"` and `name` `"English"`. If we don't find one of those then we'll look for `"en-*` and `"English (*)"`, and so on.

A nod to efficiency here: compiling these regexes at module level so they get compiled just once at application startup.

It's not necessary to anchor these regexes to the starts and ends of the strings with `^` and `$` (or `\A` and `\Z`) because we use `fullmatch()` with them below.

tests/unit/via/services/youtube_transcript_test.py
```diff
@@ -0,0 +1,300 @@
...
+)
+
+
+class TestTranscriptInfo:
+    @pytest.mark.parametrize(
+        "transcript_info,expected_id",
+        [
+            (
+                TranscriptInfoFactory(),
+                "en-us..RW5nbGlzaCAoVW5pdGVkIFN0YXRlcyk=",
```
seanh, line 26: Our custom unique IDs for human-generated transcripts look like this: `"{language_code}..{name}"` (with the name being base64-encoded so we know it won't contain any `.`'s).

tests/unit/via/services/youtube_transcript_test.py
```diff
@@ -0,0 +1,300 @@
...
+            ),
+            (
+                TranscriptInfoFactory(autogenerated=True),
+                "en-us.a.RW5nbGlzaCAoVW5pdGVkIFN0YXRlcyk=",
```
seanh, line 30: And IDs for auto-generated transcripts look like this: `"{language_code}.a.{name}"`.

tests/unit/via/services/youtube_transcript_test.py
```diff
@@ -0,0 +1,300 @@
...
+            # No "simpleText".
+            (
+                b'{"captions": {"playerCaptionsTracklistRenderer": {"captionTracks": [{"languageCode": "en", "name": {}, "baseUrl": "https://example.com"}]}}}',
+                KeyError,
+            ),
+            # No "baseUrl".
+            (
+                b'{"captions": {"playerCaptionsTracklistRenderer": {"captionTracks": [{"languageCode": "en", "name": {"simpleText": "English"}}]}}}',
+                KeyError,
+            ),
```
seanh, line 155: I put in tests for quite a few (all?) of the ways that `get_transcript_infos()` can crash on unexpected data. Maybe a bit excessive? But it might come in handy if we ever decide we want to bullet-proof this function.

This still doesn't cover everything that can go wrong, because it only covers crashes. If the `language_code`, `name` or `baseUrl` is present but isn't a string then it'll return `TranscriptInfo`'s whose `language_code`, `name` or `url`'s aren't strings. This would probably fail later on when code elsewhere tries to save these non-strings in the DB.

Or if `captionTracks` is an empty list then it'll return an empty list, which will probably crash when code elsewhere tries to pick the default transcript.

tests/unit/via/services/youtube_transcript_test.py
```diff
@@ -0,0 +1,300 @@
...
+        with pytest.raises(UnhandledUpstreamException):
+            svc.get_transcript(transcript_info)
+
+    def test_get_transcript_unexpected_response(
+        self, svc, transcript_info, http_service
+    ):
+        http_service.get.return_value.text = "foo"
+
+        with pytest.raises(ElementTree.ParseError):
+            svc.get_transcript(transcript_info)
```
seanh, line 284: The XML parsing in `get_transcript()` is very accepting and will generally return something even if the XML is quite different from what we expect: it just iterates over all the XML elements and returns the text of each. The one way I could make it crash was if the response body isn't valid XML at all.

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class YouTubeTranscriptService:
```
seanh, line 73: Getting transcripts is kind of a whole separate thing, so I put it in a separate `YouTubeTranscriptService` that the main `YouTubeService` uses. So `YouTubeTranscriptService` is basically a helper service for `YouTubeService`, but technically other services or views could use `YouTubeTranscriptService` directly as well.

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,147 @@
...
+    @property
+    def id(self) -> str:  # pylint:disable=invalid-name
+        """Return a unique ID for this transcript."""
+        return ".".join(
+            [
+                self.language_code,
+                "a" if self.autogenerated else "",
+                b64encode(self.name.encode("utf-8")).decode("utf-8"),
+            ]
+        ).rstrip(".")
```
seanh, line 35: This is our algorithm for generating our own unique IDs for transcripts. It's similar to the algorithm in https://github.com/hypothesis/via/pull/1107 but simplified.

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,147 @@
...
+
+        return [
+            TranscriptInfo(
+                language_code=caption_track_dict["languageCode"].lower(),
+                autogenerated=caption_track_dict.get("kind", None) == "asr",
+                name=caption_track_dict["name"]["simpleText"],
+                url=caption_track_dict["baseUrl"],
+            )
+            for caption_track_dict in dicts
+        ]
```
seanh, line 105: The third-party `youtube-transcript-api` library gets a blob of JSON about the video's transcripts by scraping the video's HTML page. We're doing it by calling this deprecated and undocumented legacy V1 API instead. This is no better in terms of reliability or legitimateness, but the code is simpler because we just call a JSON API rather than scraping HTML.

No error handling here: this will just crash if we get unexpected responses from the Google API. It's probably not any worse than the dodgy `youtube-transcript-api` library. We can try to make this more robust in future if we need to, but to be honest we're calling a deprecated, undocumented API.

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,147 @@
...
+                    return False
+
+            return preference["autogenerated"] == transcript_info.autogenerated
+
+        for preference in DEFAULT_TRANSCRIPT_PREFERENCES:
+            for transcript_info in transcript_infos:
+                if matches(transcript_info, preference):
+                    return transcript_info
+
+        return transcript_infos[0]
```
seanh, line 123: This is our algorithm for picking the default transcript for a video based on the `DEFAULT_TRANSCRIPT_PREFERENCES` above.

Note that if it doesn't find a match then if falls back on just returning the first transcript.

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,147 @@
...
+
+        return [
+            {
+                "text": strip_html(xml_element.text),
+                "start": float(xml_element.attrib["start"]),
+                "duration": float(xml_element.attrib.get("dur", "0.0")),
+            }
+            for xml_element in xml_elements
+            if xml_element.text is not None
+        ]
```
seanh, line 144: Again, no error handling here. It's probably not any worse than the dodgy `youtube-transcript-api` library, and we can add more error handling code in future if we need to but to be honest we're calling an undocumented private API.

tests/unit/via/services/youtube_transcript_test.py
```diff
@@ -0,0 +1,300 @@
...
+                    "language_code": caption_tracks[0]["languageCode"],
+                    "autogenerated": False,
+                    "name": caption_tracks[0]["name"]["simpleText"],
+                    "url": caption_tracks[0]["baseUrl"],
+                }
+            ),
+            Any.instance_of(TranscriptInfo).with_attrs(
+                {
+                    "language_code": caption_tracks[1]["languageCode"],
+                    "autogenerated": False,
```
seanh, line 92: I should make one of these test transcripts be an auto-generated one so that we're covering that code path

via/services/youtube_transcript.py
```diff
@@ -0,0 +1,148 @@
...
+                    return False
+
+            return preference["autogenerated"] == transcript_info.autogenerated
+
+        for preference in DEFAULT_TRANSCRIPT_PREFERENCES:
+            for transcript_info in transcript_infos:
+                if matches(transcript_info, preference):
+                    return transcript_info
+
+        return transcript_infos[0]
```
seanh, line 124: This is our algorithm for picking the default transcript for a video based on the `DEFAULT_TRANSCRIPT_PREFERENCES` above.

Note that if it doesn't find a match then if falls back on just returning the first transcript.

marcospri (approved):
Looks good, I don't have much more to add.

I looked at the original PRs which were similar. This seems sorter and a tad easier to follow.

PR #1163, opened 2023-08-15
Return the oldest transcript matching the video ID

seanh:
Just return the oldest saved transcript in the DB that matches the given `video_id`, instead of also searching for the hard-coded `transcript_id` value of `"en"`.

This is a slightly simplified rewrite of https://github.com/hypothesis/via/pull/1159.

`YouTubeService.get_transcript()` currently always saves transcripts in the DB with `transcript.transcript_id="en"` (hard-coded) and it always queries the DB for a transcript with `transcript.video_id=video_id` and `transcript.transcript_id="en"`.

In https://github.com/hypothesis/via/pull/1162 we're going to start saving transcripts in the DB with different `transcript_id`'s (we're going to be generating unique transcript IDs that work even when a video has multiple transcripts with the same language code) so we can no longer always query for a hard-coded `transcript.transcript_id="en"`.

Instead, this PR changes `get_transcript()` to query for `transcript.video_id=video_id` only and then, since there could be more than one saved transcript with the same `video_id`, it just returns the oldest saved transcript.

This should work, and it'll also mean that the default transcripts of previously-annotated videos and assignments don't change when we change the default transcript selection algorithm.

What we should _really_ do here (probably) is persist the choice of default transcript by adding a nullable `video.default_transcript` column to the `video` table (foreign key to the `transcript` table). But for the sake of expedience this will do for now. We can always back-fill a `video.default_transcript` column in future by finding the oldest saved transcript for each `video_id`.

### DIFF
--- a/tests/factories/transcript.py
+++ b/tests/factories/transcript.py
@@ -10,20 +10,17 @@ class TranscriptFactory(SQLAlchemyModelFactory):
 
     video_id = Sequence(lambda n: f"video_id_{n}")
     transcript_id = Sequence(lambda n: f"transcript_id_{n}")
-    transcript = [
-        {
-            "text": "[Music]",
-            "start": 0.0,
-            "duration": 7.52,
-        },
-        {
-            "text": "how many of you remember the first time",
-            "start": 5.6,
-            "duration": 4.72,
-        },
-        {
-            "text": "you saw a playstation 1 game if you were",
-            "start": 7.52,
-            "duration": 4.72,
-        },
-    ]
+    transcript = Sequence(
+        lambda n: [
+            {
+                "text": f"This is the first line of transcript {n}",
+                "start": 0.0,
+                "duration": 7.52,
+            },
+            {
+                "text": f"This is the second line of transcript {n}",
+                "start": 5.6,
+                "duration": 4.72,
+            },
+        ]
+    )
--- a/tests/unit/via/services/youtube_test.py
+++ b/tests/unit/via/services/youtube_test.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from io import BytesIO
 from unittest.mock import sentinel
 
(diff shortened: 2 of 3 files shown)


tests/factories/transcript.py
```diff
@@ -10,20 +10,17 @@ class Meta:
...
+                "start": 0.0,
+                "duration": 7.52,
+            },
+            {
+                "text": f"This is the second line of transcript {n}",
+                "start": 5.6,
+                "duration": 4.72,
+            },
+        ]
+    )
```
seanh, line 26: Fixed `TranscriptFactory` so that each `Transcript` object has a different value for `Transcript.transcript`. This is needed by the tests below, and is probably a good idea in general: `TranscriptFactory` always returning `Transcript`'s with the same `Transcript.transcript` is a bit of a footgun (tests that compare the values of two `Transcript.transcript`'s could have false positives thinking they came from the same `Transcript` when in fact they belonged to different ones).

tests/unit/via/services/youtube_test.py
```diff
@@ -116,7 +117,6 @@ def test_get_transcript(self, db_session, svc, YouTubeTranscriptApi):
         ]
 
     @pytest.mark.usefixtures("db_session")
-    @pytest.mark.parametrize("transcript__transcript_id", ["en"])
```
seanh, line 119: This `parametrize` was fixing the value of `transcript.transcript_id` to `"en"` because `get_transcript()`'s DB query would only find transcripts with `Transcript.transcript_id="en"`. It's no longer necessary for the test to do this.

marcospri (approved):
Looks good :+1: 

Tested in my local DB with an existing transcript from main and then:

```insert into transcript (select now(), now(), 2, video_id, transcript_id || 'new', transcript from transcript limit 1);```

```
update trasnscript set transcript='[]' where id = 2;  
```

all good

```
update trasnscript set transcript='[]' where id = 1;  
```

no transcript lines now

PR #1166, opened 2023-08-16
Remove SQLALCHEMY_SILENCE_UBER_WARNING now that we are in SQLA 2.x

marcospri:
Background on https://docs.sqlalchemy.org/en/20/changelog/changelog_14.html#change-e67bfa1efbe52ae40aa842124bc40c51

`
This deprecation warning intends to notify users who may not have set an appropriate constraint in their requirements files to block against a surprise SQLAlchemy 2.0 upgrade and also alert that the SQLAlchemy 2.0 upgrade process is available, as the first full 2.0 release is expected very soon. The deprecation warning can be silenced by setting the environment variable SQLALCHEMY_SILENCE_UBER_WARNING to "1".
`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,6 @@ setenv =
     dev: NEW_RELIC_ENVIRONMENT = {env:NEW_RELIC_ENVIRONMENT:dev}
     tests,functests: PYTEST_PLUGINS = tests.pytest_plugins.factory_boy
     dev: ALEMBIC_CONFIG = {env:ALEMBIC_CONFIG:conf/alembic.ini}
-    SQLALCHEMY_SILENCE_UBER_WARNING=1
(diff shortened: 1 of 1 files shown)


seanh (commented):
Hmm, I think this comes from the cookiecutter: https://github.com/hypothesis/cookiecutters/blob/da05a2a3d08df7b3b2c162c025258622b331d268/_shared/project/tox.ini#L35

I think we could remove that from the cookiecutter: I don't think we have any cookiecutter'ed projects that're on SQLAlchemy 1.x. If we do have some (now or in future) those projects can use cookiecutter includes to add the envvar back.

marcospri:
Closed in favor of  https://github.com/hypothesis/cookiecutters/pull/149

PR #1182, opened 2023-09-19
Make ToastMessage transitions generic and customizable by consumers

acelaya:
As mentioned in https://github.com/hypothesis/client/pull/5691#issuecomment-1709581938, this PR makes `ToastMessages` use `animate-fade-in` and `animate-fade-out` as its only default transition classes, decoupling it from consumers' needs.

It also adds a new `transitionClasses` optional prop that can be used to override those animation classes with custom values. 

```ts
export type ToastMessageTransitionClasses = {
  /** Classes to apply to a toast message when appended. Defaults to 'animate-fade-in' */
  transitionIn?: string;
  /** Classes to apply to a toast message being dismissed. Defaults to 'animate-fade-out' */
  transitionOut?: string;
};
```

This makes the `ToastMessages` component generic enough to move it to frontend-shared and use it in other apps.

> This PR is part of https://github.com/hypothesis/frontend-shared/issues/1075

### DIFF
--- a/via/static/scripts/video_player/components/ToastMessages.tsx
+++ b/via/static/scripts/video_player/components/ToastMessages.tsx
@@ -1,8 +1,12 @@
 import type { TransitionComponent } from '@hypothesis/frontend-shared';
 import { Callout } from '@hypothesis/frontend-shared';
 import classnames from 'classnames';
-import type { ComponentChildren } from 'preact';
-import { useCallback, useRef, useState } from 'preact/hooks';
+import type {
+  ComponentChildren,
+  ComponentProps,
+  FunctionComponent,
+} from 'preact';
+import { useCallback, useMemo, useRef, useState } from 'preact/hooks';
 
 export type ToastMessage = {
   id: string;
@@ -20,6 +24,13 @@ export type ToastMessage = {
    * Defaults to true.
    */
   autoDismiss?: boolean;
+};
+
+export type ToastMessageTransitionClasses = {
+  /** Classes to apply to a toast message when appended. Defaults to 'animate-fade-in' */
+  transitionIn?: string;
+  /** Classes to apply to a toast message being dismissed. Defaults to 'animate-fade-out' */
+  transitionOut?: string;
 };
 
 type ToastMessageItemProps = {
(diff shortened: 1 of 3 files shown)


via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -466,6 +466,10 @@ export default function VideoPlayerApp({
               <ToastMessages
                 messages={toastMessages}
                 onMessageDismiss={dismissToastMessage}
+                transitionClasses={{
+                  transitionIn:
+                    'lg:motion-safe:animate-slide-in-from-right animate-fade-in motion-reduce:animate-fade-in',
```
acelaya, line 471: These classes are different from the ones we will use in `client`. See https://github.com/hypothesis/client/pull/5691/files#diff-668f1611f0e3aaafda884f6e51f7a98af4efaaa00ff44e78cf7c1d38f107a9bcR34

via/static/scripts/video_player/components/ToastMessages.tsx
```diff
@@ -22,6 +26,13 @@ export type ToastMessage = {
   autoDismiss?: boolean;
 };
 
+export type ToastMessageTransitionClasses = {
+  /** Classes to apply to a toast message when appended. Defaults to 'animate-fade-in' */
+  transitionIn?: string;
+  /** Classes to apply to a toast message being dismissed. Defaults to 'animate-fade-out' */
+  transitionOut?: string;
```
robertknight, line 33: Does the component make any assumptions about what the classes will do or how long transitions will take? Using CSS classes allows for plenty of flexibility, but if there are hidden assumptions we need to capture those.

acelaya: I was going to say no, as we rely on the animationend event, but just remembered we do check the name of the animation in order to "ignore" those not relevant

https://github.com/hypothesis/via/blob/674de3cf2ca6539446748ab0dcc199e3fb3d1fb0/via/static/scripts/video_player/components/ToastMessages.tsx#L69-L74

This check is probably not needed though, and just checking the animation happened on the component itself should be reliable enough.

robertknight: > This check is probably not needed though, and just checking the animation happened on the component itself should be reliable enough.

Seems reasonable. Can you make that update in this PR then.

robertknight: Ah, but wait. There is a comment about ignoring animations on child elements. Should this check be kept but ignore events that have bubbled?

acelaya: I was suggesting keeping the `e.target !== containerRef.current` condition. Wouldn't that precisely ignore events that have bubbled? Or am I missing something?

acelaya: FYI, I have removed the `!['slide-in-from-right', 'fade-in', 'fade-out'].includes(e.animationName)` condition.

via/static/scripts/video_player/components/test/ToastMessages-test.js
```diff
@@ -158,9 +136,23 @@ describe('ToastMessages', () => {
...
       .getDOMNode()
-      .dispatchEvent(
-        new AnimationEvent('animationend', { animationName: 'invalid' })
-      );
+      .dispatchEvent(new AnimationEvent('animationend'));
+
+    assert.called(fakeOnMessageDismiss);
+  });
+
+  it('does not invoke onTransitionEnd for animations on invalid target element', () => {
```
robertknight: The meaning of "invalid" could be clearer here. You mean a child of the toast message container, right?

via/static/scripts/video_player/components/ToastMessages.tsx
acelaya: > I was suggesting keeping the e.target !== containerRef.current condition. Wouldn't that precisely ignore events that have bubbled? Or am I missing something?

Ok, I kind-of answered this myself.

There was apparently a missing test case which would cover the `e.target !== containerRef.current`. By trying to add one, I'm finding this is not working as I would expect.

acelaya: Ignore that ☝🏼 

It does cover events bubbling. I just forgot to add `{ bubbles: true }` in the test 😅

via/static/scripts/video_player/components/test/ToastMessages-test.js
acelaya: Yeah. I just changed it to `does not invoke onTransitionEnd for animation events bubbling from children`

robertknight (approved):

PR #1217, opened 2023-10-19
Update Babel

robertknight:
This is the result of running `yarn up '@babel/*'` to resolve https://github.com/hypothesis/via/security/dependabot/131.

For future reference, `yarn up --recursive @babel/traverse` would also have worked to update just the specific dependency affected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,10 +11,10 @@
     "url": "https://github.com/hypothesis/via/issues"
   },
   "dependencies": {
-    "@babel/core": "^7.23.0",
-    "@babel/preset-env": "^7.22.20",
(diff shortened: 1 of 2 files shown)


robertknight:
Correction: This didn't actually fix https://github.com/hypothesis/via/security/dependabot/131. The issue is that `@babel/traverse` is pinned to an old version by `@trivago/prettier-plugin-sort-imports`.

acelaya:
Maybe related? https://github.com/trivago/prettier-plugin-sort-imports/issues/262

acelaya:
There's also a PR to address it https://github.com/trivago/prettier-plugin-sort-imports/pull/263

PR #1231, opened 2023-11-07
Add rounded corners to UI components

### LINKED ISSUE #1230: Integrate rounded corners in UI components
Integrate rounded corners from [frontend-shared](https://github.com/hypothesis/frontend-shared/pull/1289), like it was done [in client](https://github.com/hypothesis/client/issues/5602).

The only difference is that we probably don't need a feature flag, as there are very little components which are going to actually render rounded corners.

acelaya:
Take advantage of rounded corner capabilities from [frontend-shared](https://github.com/hypothesis/frontend-shared/pull/1289), and add missing rounded corners where needed for consistency.

The visual changes are very small (only the search input has slightly bigger roundiness, and a few other elements only when focused), so I'm adding this with no feature flags.

Closes #1230

### DIFF
--- a/tailwind.config.mjs
+++ b/tailwind.config.mjs
@@ -14,8 +14,10 @@ export default {
         'slide-in-from-right': 'slide-in-from-right 0.3s forwards ease-in-out',
       },
       borderRadius: {
-        // Equivalent to tailwind default `rounded-sm` size
-        DEFAULT: '0.125rem',
+        // Equivalent to tailwind default.
+        // Remove once frontend-shared preset does not define its own config
+        DEFAULT: '0.25rem',
+        lg: '0.5rem',
       },
       fontFamily: {
         sans: [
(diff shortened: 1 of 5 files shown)


tailwind.config.mjs
```diff
@@ -14,8 +14,10 @@ export default {
         'slide-in-from-right': 'slide-in-from-right 0.3s forwards ease-in-out',
       },
       borderRadius: {
-        // Equivalent to tailwind default `rounded-sm` size
-        DEFAULT: '0.125rem',
+        // Equivalent to tailwind default.
+        // Remove once frontend-shared preset does not define its own config
+        DEFAULT: '0.25rem',
+        lg: '0.5rem',
```
acelaya, line 20: I was planning to remove this config and add the CSS variables instead, but since we don't have any CSS file in this project, this approach was easier.

via/static/scripts/video_player/components/VideoPlayerApp.tsx
```diff
@@ -402,7 +402,7 @@ export default function VideoPlayerApp({
                   onClick={() => setPlaying(playing => !playing)}
                   data-testid="play-button"
                   variant="custom"
-                  classes="text-grey-7 hover:text-grey-9 font-semibold"
+                  classes="text-grey-7 hover:text-grey-9 font-semibold rounded"
```
acelaya, line 405: Adding `rounded` only for when the button is focused.

robertknight (approved):

PR #1262, opened 2024-01-04
Reorganise Via's Gunicorn config

seanh:
This reorganise's Via's Gunicorn config files following the same pattern
as was recently done in h:

hypothesis/h#8407

I'm applying this simpler and more explicit Gunicorn config approach to
all our apps consistently. For motivation see:

https://docs.google.com/document/d/13AnUPtu9AO3PfRm-fhH7_3RZzyLGSFd4uPv36dIeynA

Benefits:

* Gunicorn config files are explicitly mentioned in the `gunicorn`
  commands in the `supervisord[-dev].conf` files. This makes things less
  confusing/surprising, less likely that a Gunicorn config file will be
  missed (compared to Gunicorn's default behavior if you run it without
  a `--config` argument: it reads any `gunicorn.conf.py` file in the
  current working directory).

* The Gunicorn config files are in the `conf/` directory with the other
  config files. Again: makes it less likely that a config file will go
  unnoticed.

* Separate production and development Gunicorn config files. This is
  consistent with our separate production and development Pyramid config
  files, and is probably a good idea to prevent development settings
  accidentally getting applied in production. (Gunicorn's default
  behavior of reading a root `gunicorn.conf.py` file would mean that
  file gets read in both dev and production).

### DIFF
--- a/conf/gunicorn-dev.conf.py
+++ b/conf/gunicorn-dev.conf.py
@@ -1,0 +1,4 @@
+bind = "0.0.0.0:9082"
+reload = True
+reload_extra_files = "via/templates"
+timeout = 0
--- a/conf/gunicorn.conf.py
+++ b/conf/gunicorn.conf.py
@@ -1,0 +1,2 @@
+bind = "unix:/tmp/gunicorn-web.sock"
+worker_tmp_dir = "/dev/shm"
--- a/.cookiecutter/cookiecutter.json
+++ b/.cookiecutter/cookiecutter.json
@@ -34,6 +34,7 @@
     "__docker_network": "via_default",
     "__github_url": "https://github.com/hypothesis/via",
     "__hdev_project_type": "application",
-    "__copyright_year": "2016"
+    "__copyright_year": "2016",
+    "__gunicorn_bind": "unix:/tmp/gunicorn-web.sock"
   }
 }
--- a/.cookiecutter/includes/conf/supervisord-dev.conf.json
+++ b/.cookiecutter/includes/conf/supervisord-dev.conf.json
@@ -5,7 +5,7 @@
       "startsecs": "0"
     },
     "web": {
-      "command": "gunicorn -c conf/gunicorn/dev.conf.py --paste conf/development.ini"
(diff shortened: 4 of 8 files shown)


conf/gunicorn-dev.conf.py
```diff
@@ -0,0 +1,5 @@
+bind = "0.0.0.0:9082"
+reload = True
+reload_extra_files = "via/templates"
+timeout = 0
+workers = 2
```
seanh: Similar to the change made in Checkmate: `conf/gunicorn/dev.conf.py` has been moved to `conf/gunicorn-dev.conf.py`. `reload_extra_files` has been added and `workers` has been changed from `4` to `2`, but otherwise the config should be the same.

conf/gunicorn.conf.py
```diff
@@ -0,0 +1,2 @@
+bind = "unix:/tmp/gunicorn-web.sock"
```
seanh, line 1: Via differs from our other apps in that Gunicorn is bound to a Unix socket rather than to an HTTP port. This is necessary because Via has an internal NGINX instance that [connects to Gunicorn on this socket](https://github.com/hypothesis/via/blob/7df9e122f094b1014af8df64386313322a630c68/conf/nginx/includes/app_upstream.conf#L12).

I don't think it's possible for our other apps that don't have an internal NGINX to use a Unix socket: I think Elastic Beanstalk's external NGINX instance needs to connect to Gunicorn over HTTP not a socket?

Also not sure whether it'd be desirable for internal NGINX's to connect to Gunicorn via HTTP: I think Unix sockets are more efficient.

So it's probably best to leave this as it is and let it differ between apps.

conf/gunicorn.conf.py
```diff
@@ -0,0 +1,2 @@
...
+worker_tmp_dir = "/dev/shm"
```
seanh, line 2: Adding the `worker_tmp_dir` setting from LMS to Via as well, @marcospri

marcospri (approved):

PR #1267, opened 2024-01-09
Upgrade python dependencies

marcospri:
This is the result of running:

make requirements --always-make args='--upgrade'

Non top level dependencies can become outdated as dependabot doesn't send PR for those.

On top of that upgrading single packages with all the existing constraints might make impassible to satisfy all constraints.

This approach makes sure we are up to date with all our dependencies as specified on the .in files but without considering any constraints in the .txt files.


requirements/build.txt
```diff
@@ -2,40 +2,40 @@
 # This file is autogenerated by pip-compile with Python 3.8
 # by the following command:
 #
-#    pip-compile --allow-unsafe --config=pyproject.toml requirements/build.in
```
marcospri, line 5: This seems to undo the format changes made by https://github.com/hypothesis/via/commit/20c3783fc530fcaf53381b823aad7c5582b71618

but it's mostly generated the same way :thinking:

seanh (approved):

PR #1268, opened 2024-01-10
Replace usage of deprecated CGI module

marcospri:
Only used to parse MIMETYPES, replaced by email.Messsage

https://peps.python.org/pep-0594/#cgi

### DIFF
--- a/tests/unit/via/services/url_details_test.py
+++ b/tests/unit/via/services/url_details_test.py
@@ -94,7 +94,7 @@ class TestGetURLDetails:
         result = svc.get_url_details(sentinel.youtube_url)
 
         youtube_service.get_video_id.assert_not_called()
-        assert result == ("dummy", 200)
+        assert result == ("application/x-testing", 200)
 
     def test_it_when_url_blocked_by_checkmate(self, checkmate_service, svc):
         checkmate_service.raise_if_blocked.side_effect = BadURL("Bad URL")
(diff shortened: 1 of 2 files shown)


tests/unit/via/services/url_details_test.py
```diff
@@ -94,7 +94,7 @@ def test_it_when_youtube_disabled(self, youtube_service, GoogleDriveAPI, svc):
         result = svc.get_url_details(sentinel.youtube_url)
 
         youtube_service.get_video_id.assert_not_called()
-        assert result == ("dummy", 200)
```
marcospri, line 97: dummy gets parsed as the default according to the spec, text/plain.

Using a more explicit testing mimetype.

marcospri: As far as I can tell in the code base we expect explicit values to dispatch to PDFs for example.

seanh: Would `sentinel.content_type` work here? Sentinels are often a bit more explicit than strings that a test sentinel value is being used

seanh (approved):

tests/unit/via/services/url_details_test.py
marcospri: Not without mocking the content type parsing which I reckon will make the test less usueful.

PR #1276, opened 2024-01-31
Add Sentry support to status endpoint

seanh:


tests/unit/via/views/status_test.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 from checkmatelib import CheckmateException
 
-from via.views.status import get_status
+from via.views.status import status
```
seanh, line 4: Fixing the `get_status` name to just `status` again

via/views/status.py
```diff
@@ -1,15 +1,12 @@
-"""Monitoring views."""
-
```
seanh, line 2: Removing pointless docstrings again.

marcospri (approved):

PR #1291, opened 2024-02-07
Notify screen readers when transcript is filtered

### LINKED ISSUE #1274: Announce result count when searching transcripts
From our [2024 VPAT review](https://docs.google.com/document/d/1YdYr5jIsKEm4CkOQF9un3LUc-A8xLtwFP5kWbNpqYOM/edit), searching the transcript of an annotation should result in an announcement of the number of results, for the benefit of assistive technology users.

See the section that begins:

> Searching within a YouTube transcript does not communicate anything about the results of the search except by visually removing lines in the transcript that don’t match the search string

acelaya:
Closes #1274 

This PR ensures a hidden toast message is appended when filtering the transcript via search term, making screen readers announce the amount of results.

The way this has been implemented is not 100% correct regarding to https://react.dev/learn/you-might-not-need-an-effect, since it adds a side effect to `Transcript` based on some internal state.

The more correct solution would probably imply appending the toast message as part of the `onChange` event handler for the `FilterInput` (I think it would fall under this case https://react.dev/learn/you-might-not-need-an-effect#notifying-parent-components-about-state-changes), however, at that point we don't know how many segments matched the search term, hence the proposed solution here.

I evaluated other approaches as well, like wrapping `FilterInput` into a component which calculates `filterMatches` and appends the toast message as part of `FilterInput`'s `setFilter` callback, and make that a piece of state that is handled by `VideoPlayerApp` and passed down to `Transcript`. This would mean the raw `filter` is no longer exposed beyond that wrapper component, but it doesn't seem to be needed anyway.

This would require more changes, though.

EDIT: Another argument in favor of going in a different direction is that, if `onFilterMatch` is not properly memoized, you get infinite toast messages appended, which is very unintuitive.

### Testing steps

1. Run a screen reader
2. Try filtering the transcript.
3. The screen reader should announce a message in the lines of `"foo" returned 5 results` every time the search term changes to more than 2 characters.

### DIFF
--- a/via/static/scripts/video_player/components/Transcript.tsx
+++ b/via/static/scripts/video_player/components/Transcript.tsx
@@ -67,6 +67,12 @@ export type TranscriptProps = {
   filter?: string;
 
   /**
+   * Invoked when the filter is applied to the transcript segments, with the
+   * amount that matched and the filter itself
+   */
+  onFilterMatch?: (filter: string, results: number) => void;
+
+  /**
    * Callback invoked when the user selects a segment from the transcript.
    */
   onSelectSegment?: (segment: Segment) => void;
@@ -270,6 +276,7 @@ export default function Transcript({
   controlsRef,
   currentTime,
   filter = '',
+  onFilterMatch,
   onSelectSegment,
   transcript,
 }: TranscriptProps) {
@@ -360,6 +367,12 @@ export default function Transcript({
     }
     return filterTranscript(transcript.segments, filter);
   }, [filter, transcript]);
+
+  useEffect(() => {
+    if (filterMatches) {
+      onFilterMatch?.(filter, filterMatches.size);
+    }
+  }, [filter, filterMatches, onFilterMatch]);
 
   // Adjust the scroll position when the transcript container is resized, so the
   // user doesn't lose their place. See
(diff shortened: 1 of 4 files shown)


acelaya:
We have decided to go with this approach for now.

This [slack thread](https://hypothes-is.slack.com/archives/C1M8NH76X/p1707300258118989) gives more details on what to do in future if filtered segments need to be used in other components.

robertknight (changes requested):
When testing this manually I noticed that the current approach is prone to reading duplicate notifications because the search input uses incremental search. This means if I type "some query", then a new search is executed for each character typed, and multiple hidden toast messages are rendered. Screen readers will try to read each of these. Additionally the "success" prefix that is added is redundant in this context.

I think what we'll need to do is avoid using toast messages and instead add a dedicated visually hidden container with an `aria-live` status that contains the current number of matches. That way the screen reader can use its existing logic for coalescing updates to a live element.

acelaya:
> When testing this manually I noticed that the current approach is prone to reading duplicate notifications because the search input uses incremental search. This means if I type "some query", then a new search is executed for each character typed, and multiple hidden toast messages are rendered. Screen readers will try to read each of these. Additionally the "success" prefix that is added is redundant in this context.
> 
> I think what we'll need to do is avoid using toast messages and instead add a dedicated visually hidden container with an `aria-live` status that contains the current number of matches. That way the screen reader can use its existing logic for coalescing updates to a live element.

Yeah, that makes sense. Also, it would avoid the problematic behavior explained in the PR description, where a side effect is triggered when a prop changes.

I'll close this PR and prepare a new one with the other approach.

PR #1335, opened 2024-04-18
Add a more verbose accessible label for timestamps

robertknight:
Set the accessible label of timestamps to a more verbose format that requires less contextual awareness to understand. This also matches how timestamps are rendered in YouTube's transcript viewer.

See test cases in `time-test.ts` for examples.

Verbose timestamp

### DIFF
--- a/via/static/scripts/video_player/utils/time.ts
+++ b/via/static/scripts/video_player/utils/time.ts
@@ -2,6 +2,18 @@ function pad(value: number): string {
   return value.toString().padStart(2, '0');
 }
 
+function formatTimePiece(
+  value: number,
+  singular: string,
+  plural: string
+): string {
+  if (value === 1) {
+    return `${value} ${singular}`;
+  } else {
+    return `${value} ${plural}`;
+  }
+}
+
 /**
  * Format a video timestamp.
  *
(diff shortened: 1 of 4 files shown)


via/static/scripts/video_player/utils/time.ts
```diff
@@ -10,18 +22,46 @@ function pad(value: number): string {
...
+        formatted.push(formatTimePiece(mins, 'minute', 'minutes'));
+      }
+      if (secs > 0 || (hours === 0 && mins === 0)) {
+        formatted.push(formatTimePiece(secs, 'second', 'seconds'));
+      }
+      return formatted.join(', ');
+    }
+    /* istanbul ignore next */
+    default:
+      return '';
```
acelaya, line 65: Maybe it's worth extracting the logic for both cases to local/non-exposed functions.

```ts
const times = { hours, mins, secs };
return format === 'digits' ? timeAsDigits(times) : timeAsDescription(times);
```

```ts
function timeAsDigits({ hours, mins, secs }): string {
  if (hours > 0) {
    return `${hours}:${pad(mins)}:${pad(secs)}`;
  }

  return `${mins}:${pad(secs)}`;
}

function timeAsDescription({ hours, mins, secs }): string {
  const formatted = [];
  if (hours > 0) {
    formatted.push(formatTimePiece(hours, 'hour', 'hours'));
  }
  if (mins > 0) {
    formatted.push(formatTimePiece(mins, 'minute', 'minutes'));
  }
  if (secs > 0 || (hours === 0 && mins === 0)) {
    formatted.push(formatTimePiece(secs, 'second', 'seconds'));
  }
  return formatted.join(', ');
}
```

acelaya (approved):
Looks good. Just added a small suggestion on attempting to simplify `formatTimestamp`.

via/static/scripts/video_player/utils/time.ts
robertknight: Could do, but as we're talking about a ~30 line function I'm not sure that splitting it up has that much value at this point. I'm going to leave as is and we get split it up when there is a clearer need.

PR #1371, opened 2024-06-25
Correct Postgres version in dev and CI

seanh:
Via's staging and production environments use Postgres 15.5 not 15.3.

### DIFF
--- a/.cookiecutter/cookiecutter.json
+++ b/.cookiecutter/cookiecutter.json
@@ -30,7 +30,7 @@
     "postgres": "yes",
     "docker": "yes",
     "__postgres_port": "4433",
-    "__postgres_version": "15.3-alpine",
+    "__postgres_version": "15.5-alpine",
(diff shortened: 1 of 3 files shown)


marcospri (changes requested):
I don't think we should merge these if we are in fact auto-updating the RDS minor versions. We'll need to recreate the local containers as we try to keep up with those updates.

seanh:
> I don't think we should merge these if we are in fact auto-updating the RDS minor versions. We'll need to recreate the local containers as we try to keep up with those updates.

Agree. We _could_ automate this. For example we could have something that sends automated PRs to upgrade dev and CI to the latest Postgres minor version (even if that would sometimes mean that dev and CI are temporarily on a _newer_ minor version than production). But I think it's probably not worth the effort to implement that automation or the hassle of regular development environment Postgres version upgrades.

PR #1399, opened 2024-08-09
Support configuring a proxy for fetching YouTube transcripts

robertknight:
Add a `YOUTUBE_PROXY` environment variable which specifies a proxy service to use when fetching caption tracks/transcripts from YouTube.

The format of this variable should be an argument that can be used with the `proxies` argument to `requests.post`, eg.
"http://proxy_user:proxy_pass@proxy_host:port".

Part of https://github.com/hypothesis/support/issues/143.
Slack thread: https://hypothes-is.slack.com/archives/C4K6M7P5E/p1723188976861449

Testing: Set the `YOUTUBE_PROXY` environment variable before starting the service. See Slack thread for details.

### DIFF
--- a/via/services/youtube_transcript.py
+++ b/via/services/youtube_transcript.py
@@ -4,6 +4,8 @@ from dataclasses import dataclass
 from typing import Dict, List
 from xml.etree import ElementTree
 
+import requests
+
 from via.services.http import HTTPService
 
 
@@ -145,4 +147,8 @@ class YouTubeTranscriptService:
 
 
 def factory(_context, request):
-    return YouTubeTranscriptService(http_service=request.find_service(HTTPService))
+    session = requests.Session()
+    if youtube_proxy := request.registry.settings.get("youtube_proxy"):
+        session.proxies["https"] = youtube_proxy
+    http_svc = HTTPService(session=session)
+    return YouTubeTranscriptService(http_service=http_svc)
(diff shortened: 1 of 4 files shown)


marcospri (approved):
Looks good, let's see how it behaves :+1:

via/services/youtube_transcript.py
```diff
@@ -145,4 +147,8 @@ def strip_html(xml_string):
 
 
 def factory(_context, request):
-    return YouTubeTranscriptService(http_service=request.find_service(HTTPService))
+    session = requests.Session()
+    if youtube_proxy := request.registry.settings.get("youtube_proxy"):
+        session.proxies["https"] = youtube_proxy
+    http_svc = HTTPService(session=session)
+    return YouTubeTranscriptService(http_service=http_svc)
```
marcospri, line 154: We'll use the proxy for all requests of the service, we might have to  change that in the future but this does the tricky for now :+1:

robertknight: This is a good point. There are two requests this service makes:

- The https://youtubei.googleapis.com/youtubei/v1/player endpoint which returns JSON containing signed URLs for caption tracks. This is the one where the proxy is needed.
- Fetching the caption track URLs which are XML content containing the captions. I don't know if the proxy is essential for this.

PR #1506, opened 2025-01-06
Upgrade python to 3.12.7

mtomilov:


pyproject.toml
```diff
@@ -8,6 +8,7 @@ filterwarnings = [
     "ignore:^pkg_resources is deprecated as an API:DeprecationWarning:pyramid",
     "ignore:^Deprecated call to .pkg_resources\\.declare_namespace\\('.*'\\).\\.:DeprecationWarning:pkg_resources",
     "ignore:^'cgi' is deprecated and slated for removal in Python 3\\.13$:DeprecationWarning:webob",
+    "ignore:^datetime\\.datetime\\.utcnow\\(\\) is deprecated and scheduled for removal in a future version\\.:DeprecationWarning",
```
mtomilov, line 11: This one due to `webob` which sadly hasn't been updated 2021 where even the current main [has the issue](https://github.com/Pylons/webob/blob/39d5af3c797e7b867f152c2e8c979de42d029403/src/webob/response.py#L1270).

via/services/google_drive.py
```diff
@@ -157,7 +157,7 @@ def parse_file_url(cls, public_url):
 
         return data
 
-    def iter_file(self, file_id, resource_key=None) -> Iterator[ByteString]:
+    def iter_file(self, file_id, resource_key=None) -> Iterator[bytes]:
```
mtomilov, line 160: `ByteString` is now deprecated and we should be using plain `bytes` instead.

robertknight:
Will revisit this when we're not in the middle of a project.

PR #1509, opened 2025-01-21
Move to Ruff

seanh:
This is the result of applying https://github.com/hypothesis/cookiecutters/pull/196 to this project.


tests/unit/via/views/conftest.py
```diff
@@ -1,2 +1 @@
 # Import this so that our webargs customization is registered in the view tests.
-import via.views.webargs  # pylint: disable=unused-import
```
seanh, line 2: FIXME. Apparently this line wasn't needed because `ruff check --fix` removed it and the tests are still passing. If removing this line need to also remove the comment above it. Or don't remove this line and add a `# noqa` instead. I'm not sure which is correct.

via/views/exceptions.py
```diff
@@ -124,7 +124,7 @@ def _serialise_requests_info(request, response):
         try:
             data["text"] = response.text[:5000]
 
-        except Exception:  # pylint: disable=broad-except
+        except Exception:  # pylint: disable=broad-except  # noqa: BLE001
```
seanh, line 127: FIXME: need to grep for `pylint` comments and remove them, there's still a bunch in Via.

tests/unit/via/missing_init_files_test.py
```diff
@@ -1,22 +0,0 @@
...
-
-    message = "You need to add these missing __init__.py file(s):\n\n"
-    message += "\n".join(missing)
-    message += (
-        "\n\n"
-        "All directories containing test files need to have __init__.py\n"
-        "files, otherwise PyLint doesn't lint them. See:\n\n"
-        "https://github.com/hypothesis/lms/issues/1606"
-    )
-    assert not missing, message
```
seanh, line 22: We don't need this test: Ruff [has a lint rule for it](https://docs.astral.sh/ruff/rules/implicit-namespace-package/). That lint rule is [currently disabled in our Ruff config](https://github.com/hypothesis/cookiecutters/blob/0436f59e6984641ad1ff0077849a9530da511b21/_shared/project/pyproject.toml#L94) but we can aim to enable it soon. Deleting the test now anyway, otherwise we'll inevitably forget to delete it. I think some other projects may have versions of this test too

marcospri (approved):

PR #1587, opened 2025-05-06
Define coverage threshold in vitest config

acelaya:
During the migration to Vitest, we lost the code [coverage threshold](https://github.com/hypothesis/via/pull/1562/files#diff-fd24fb7620ddf07dcf82a4d412502e4f953ce9f23a0f8262b0388e36d0ae2fb2L31-L35) requirement. This PR recovers it

### TODO
- [x] Fix missing coverage


via/static/scripts/video_player/test/config-test.js
```diff
@@ -0,0 +1,56 @@
...
+
+    try {
+      readConfig();
+    } catch (e) {
+      error = e;
+    }
+
+    assert.deepEqual(
+      new Error('No config object found for selector ".js-config"'),
+      error,
```
robertknight: I would prefer `assert.equal(error.message, expectedMessage)`, since not all properties of a fresh error are guaranteed to match. The stack trace for example happens not to be considered here, but the `error.stack` properties will differ.

via/static/scripts/video_player/test/config-test.js
```diff
@@ -0,0 +1,56 @@
...
+
+    try {
+      readConfig();
+    } catch (e) {
+      error = e;
+    }
+
+    assert.deepEqual(
+      new Error('Failed to parse frontend configuration'),
+      error,
```
robertknight: See above about error comparisons.

via/static/scripts/video_player/test/config-test.js
```diff
@@ -0,0 +1,56 @@
...
+    );
+  });
+
+  it('parses config body as JSON', () => {
+    createConfigElement('{ "foo": ["bar"] }');
```
robertknight: Could also use `JSON.stringify` to generate the JSON.

robertknight (approved):

PR #1646, opened 2025-09-17
Switch to CSS-based Tailwind config

robertknight:
~~**Depends on https://github.com/hypothesis/frontend-shared/pull/2090**~~

See https://github.com/hypothesis/h/pull/9892


robertknight:
Test failure:

```
FAIL   chromium  build/scripts/tests.bundle.js > useScrollAnchor > adjusts scroll position when container is resized
AssertionError: expected 50 to not equal 50

 ❯ notEqual via/static/scripts/video_player/hooks/test/use-scroll-anchor-test.js:124:11
```

Doesn't happen for me locally. Let's see if this is a flake in CI...

acelaya:
> Doesn't happen for me locally.

Does this project define the window dimensions in vitest config? Otherwise, it could cause different results in tests between local and CI envs.

robertknight:
> Does this project define the window dimensions in vitest config? Otherwise, it could cause different results in tests between local and CI envs.

It does. The test fixed the size of a container but did not fix the font size, so I guessed that fixing the font size might help. I pushed a commit to do that and it seems to have resolved the issue.

acelaya (approved):

PR #1733, opened 2026-01-29
Update YouTube with Supadata for transcript service

santicomp2014:
Update transcript service to use supadata instead.


via/services/youtube_transcript.py
```diff
@@ -39,108 +125,103 @@ def id(self) -> str:
...
-                    autogenerated=caption_track_dict.get("kind", None) == "asr",
-                    name=caption_track_dict["name"]["simpleText"],
-                    url=caption_track_dict["baseUrl"],
+                    language_code=lang.lower(),
+                    # Supadata doesn't tell us if it's autogenerated, assume False
+                    autogenerated=False,
+                    name=_get_language_name(lang),
+                    # URL format for Supadata (stores video_id and original lang code)
+                    url=f"supadata://{video_id}/{lang}",
+                )
```
karenrasmussen, line 163: Could we add a video_id attribute? That way, `get_transcript` can use the video_id directly instead of having to parse it from the URL

PR #1741, opened 2026-02-24
Remove via access

santicomp2014:
This pull request updates the `index.html.jinja2` template to reflect recent changes in access to the Via proxy service. The main change is the removal of the URL submission form and the addition of new messaging that directs users to alternative ways to annotate web pages and PDFs.

#1708 

Access and user guidance updates:

* Removed the URL submission form and replaced it with a notice about restricted access to Via, including a link to a blog post explaining the change.
* Updated instructions to direct users to use the Chrome extension, Bookmarklet, or sites that embed Hypothesis for annotation, instead of Via.


santicomp2014:
this was superseded by #1739

